PR #6, opened 2019-03-16
Dockerize triagebot and add CI

pietroalbini:
This PR does multiple things to allow us to deploy triagebot with Docker:

* Removed the requirement for the `GITHUB_USERNAME` environment variable: the username can be easily fetched through the API, and it's less error prone to do so.
* Created a Dockerfile to package the bot in a small docker container.
* Configured Travis CI both to test/fmt/clippy and to build and deploy the docker image to our registry.
* Improved the error messages when an environment variable is missing.

Before merging this PR you need to create the image on ECR and add the AWS credentials to Travis to push the image.


.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: rust
+rust: nightly-2019-03-15
+cache:
+  - cargo
+
+script: ci/run.sh
+
+# Publish the Docker image to the infra AWS registry
+before_deploy:
+  - pip installi --user awscli
```
Mark-Simulacrum: Presumably a typo? `installi`?

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+    /tmp/rustup-init -y --no-modify-path --default-toolchain $(cat /tmp/rust-toolchain)
+ENV PATH=/root/.cargo/bin:$PATH
+
+# Build the dependencies in a separate step to avoid rebuilding all of them
+# every time the source code changes. This takes advantage of Docker's layer
+# caching, and it works by copying the Cargo.{toml,lock} with dummy source code
+# and doing a full build with it.
+WORKDIR /tmp/source
+COPY Cargo.lock Cargo.toml /tmp/source/
+COPY parser/Cargo.toml /tmp/source/parser/Cargo.toml
```
Mark-Simulacrum: Technically only the root Cargo.lock should matter I think.

pietroalbini: Unfortunately not. If I remove that line this is the output:

```
Step 9/18 : RUN cargo fetch
 ---> Running in 5ffccb66d836
error: failed to read `/tmp/source/parser/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

Mark-Simulacrum: Oh, right, I was just thinking about caching -- not the actual requirements from Cargo.

Mark-Simulacrum:
Okay, everything should be setup with AWS so merging.

PR #7, opened 2019-03-16
More changes needed for Docker

pietroalbini:
* Switched port from 8000 to 80 inside the container (configurable through `TRIAGEBOT_PORT`)
* ~~Added the `ctrlc` crate to properly exit inside Docker (at the moment it just hangs until it gets killed inside Docker)~~
* Changed how the bot is started inside docker to properly exit on SIGINT/SIGTERM.

r? @Mark-Simulacrum

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,4 +53,4 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
 
 COPY --from=build /tmp/source/target/release/triagebot /usr/local/bin/
 ENV TRIAGEBOT_PORT=80
-CMD ["triagebot"]
+CMD triagebot
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,4 +52,5 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     ca-certificates
 
 COPY --from=build /tmp/source/target/release/triagebot /usr/local/bin/
+ENV TRIAGEBOT_PORT=80
 CMD ["triagebot"]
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
The `ctrlc` change seems quite unfortunate -- what is triagebot doing that requires that? I'm wondering if maybe there's some other way of accomplishing it; introducing a new crate dependency just to be compatible with Docker feels wrong.

Looks like you will also need to run rustfmt.

pietroalbini:
> The ctrlc change seems quite unfortunate -- what is triagebot doing that requires that? I'm wondering if maybe there's some other way of accomplishing it; introducing a new crate dependency just to be compatible with Docker feels wrong.

Yeah, found a (better?) solution to this. The `CMD` instruction of Docker has two syntaxes: one with an array of arguments (like `CMD ["triagebot"]`) and one with a shell command (like `CMD triagebot`). The first syntax spawns the bot as PID 1, while the latter spawns a shell as PID 1 which then runs the bot. Linux manages the PID 1 differently though, and if the process doesn't register a signal handler for SIGINT and SIGTERM it just ignores those signals.

Since triagebot doesn't handle those signals the two solutions to make it exit properly are handling them (with `ctrlc`) or just wrap the binary in a shell (with `CMD triagebot`). My previous commit implemented the former solution, while my new one the latter.

Personally I'd prefer to handle this explicitly, but I understand why you don't want to add another dependency.

Mark-Simulacrum:
Okay, sounds good. Yeah, it seems like making triagebot (or indeed any Rust binary) would need more thought than just ctrl-c handling, or at least so I presume. Thanks for fixing it!

PR #9, opened 2019-04-13
Add per-repo config and prepare for org-wide webhook

pietroalbini:
This PR contains multiple refactors to allow other teams to easily add triagebot on their own repositories, and to reduce the maintenance overhead on our end.

Each repository now has its own configuration file, `triagebot.toml`, which lives at the root of the master branch that repository, and it's fetched by triagebot at runtime whenever needed (and cached for 2 minutes). If a repository doesn't have the configuration file and an user tries to execute a valid command an error will be shown in the comments instead.

Each of the handlers has its own section in the configuration file, and triagebot will call the handler only if the configuration section is present, allowing repository owners to toggle the handlers they want on their end. As with disabled repos, if an user tries to use a disabled handler an error will be shown.

The way handlers are registered in the code changed: instead of using dynamic dispatch they have to be registered in a macro in `handlers.rs` (which replaces the `mod foo;`). This way it's possible to automatically disable not configured handlers, and associated types can be used in the trait.

The trait also changed, with `handle_event` being split into two methods (`parse_input` and `handle_input`). This way the input is also parsed when the handler is disabled, allowing the code to show the disabled handler error message only when a valid command was actually present in the comment.

Finally I did some bits of cleanup in the code, and renamed the `label` handler in `relabel` since that's what it actually does.

r? @Mark-Simulacrum 

---

Before deploying this PR, the following `triagebot.toml` must be landed in the rustc repo:

```toml
[relabel]
allow-unauthenticated = [
    "C-*", "A-*", "E-*", "NLL-*", "O-*", "S-*", "T-*", "WG-*",
    # I-* without I-nominated
    "I-compilemem", "I-compiletime", "I-crash", "I-hang", "I-ICE", "I-slow",
]
```


Mark-Simulacrum:
This looks good to me (let's merge the config to rust-lang rust as soon as possible and then merge this).

PR #73, opened 2019-07-28
relabel: refactor to support negative patterns and add tests

pietroalbini:
This allows negative patterns that override whitelisted labels, for example `["I-*", "!I-nominated"]` (useful in rustc), adding tests in the meantime.

r? @Mark-Simulacrum

### DIFF
--- a/src/handlers/relabel.rs
+++ b/src/handlers/relabel.rs
@@ -84,7 +84,7 @@ async fn handle_input(
                  we were unable to check if you are a team member.",
                  name
             )),
-            Err(err) => Some(err.to_string()),
+            Err(err) => Some(err),
         };
         if let Some(msg) = err {
             let cmnt = ErrorComment::new(&event.issue().unwrap(), msg);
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
CI failure looks potentially legitimate?

pietroalbini:
...of course I forgot to save before committing. Fixed.

src/handlers/relabel.rs
```diff
@@ -76,8 +76,18 @@ async fn handle_input(
...
-            let cmnt = ErrorComment::new(&event.issue().unwrap(), msg.to_string());
+        let err = match check_filter(name, config, is_member(&event.user(), &ctx.github).await) {
+            Ok(CheckFilterResult::Allow) => None,
+            Ok(CheckFilterResult::Deny) => Some(format!("Label {} can only be set by Rust team members", name)),
+            Ok(CheckFilterResult::DenyUnknown) => Some(format!(
+                "Label {} can only be set by Rust team members;\
+                 we were unable to check if you are a team member.",
+                 name
+            )),
+            Err(err) => Some(err.to_string()),
```
Mark-Simulacrum: We probably want to have check_filter return `Result<CheckFilterResult, String>` or something like that instead here to avoid the potentially not great conversion to string here (since we send this to the user in a GH comment)

pietroalbini: Is 818e6b4 what you had in mind?

PR #234, opened 2019-12-19
Add glacier command for adding ICE snippets easily

hellow554:
This is a first implementation of the glacier command for triagebot.

The "tracking issue" is https://github.com/rust-lang/glacier/issues/231

I want a small feedback on the parser itself and its syntax. 
Is there a command am I missing? How is the usability?

One thing I don't like is that I have to quote the URL, but I don't see a *easy* way around. Any thoughts on it?

### DIFF
--- a/parser/src/command.rs
+++ b/parser/src/command.rs
@@ -3,6 +3,7 @@ use crate::error::Error;
 use crate::token::{Token, Tokenizer};
 
 pub mod assign;
+pub mod glacier;
 pub mod nominate;
 pub mod ping;
 pub mod relabel;
@@ -17,6 +18,7 @@ pub enum Command<'a> {
     Assign(Result<assign::AssignCommand, Error<'a>>),
     Ping(Result<ping::PingCommand, Error<'a>>),
     Nominate(Result<nominate::NominateCommand, Error<'a>>),
+    Glacier(Result<glacier::GlacierCommand, Error<'a>>),
     None,
 }
 
@@ -95,6 +97,11 @@ impl<'a> Input<'a> {
             Command::Nominate,
             &original_tokenizer,
         ));
+        success.extend(parse_single_command(
+            glacier::GlacierCommand::parse,
+            Command::Glacier,
+            &original_tokenizer,
+        ));
 
         if success.len() > 1 {
             panic!(
@@ -133,6 +140,7 @@ impl<'a> Command<'a> {
             Command::Assign(r) => r.is_ok(),
             Command::Ping(r) => r.is_ok(),
             Command::Nominate(r) => r.is_ok(),
+            Command::Glacier(r) => r.is_ok(),
             Command::None => true,
         }
     }
--- a/parser/src/command/glacier.rs
+++ b/parser/src/command/glacier.rs
@@ -1,0 +1,99 @@
+//! The glacier command parser.
+//!
+//! The grammar is as follows:
+//!
+//! ```text
+//! Command:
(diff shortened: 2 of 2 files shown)


JohnTitor:
What do you expect from `glacier remove`?

Mark-Simulacrum:
It is not immediately apparent to me what the `CodeSource::Post` is intended to be -- I think we would not want to try to parse a codeblock from the input comment. I would prefer that we stick to just parsing the URL as you've suggested. One other possibility is accepting a GitHub gist URL. I'm not sure what our requirements on glacier's side are.

hellow554:
(I should explain my thought a little bit more next time :sweat_smile:)

The intention for `glacier remove` was that an ICE can be removed from the repository (e.g. somebody added one, but a maintainer decided, that it already exists as a dup or something else (malicious code).

`CodeSource::Post` was thought as a quicker way to add an ICE than using the playground. If ony one codeblock (\```rust code... ```) is available that one will be taken. 
I thought about using `gist` as well, but multiple files can be added to one gist, which I don't know yet how to handle. If we use the playground, a gist will be created in the background as well, but only one file, so that would be okay.

Mark-Simulacrum:
Let's stick to just the playpen for now.

I would expect this command to be limited to authorized users (given that it takes external action, rather than e.g. adding a label to the given issue). This is similar to most other commands we have today, so we don't need to worry too much about removal.

The remove command probably needs to take some form of token as well (perhaps the same playpen URL, and we can key by that in whatever database/filesystem/etc we're using to store these pens).

JohnTitor:
Currently, glacier checks if there are fixed ICEs every day with GHA so I think `glacier remove` command is unhelpful maybe.
And it'd be nice that `glacier add` submits a PR to glacier [like autofix](https://github.com/rust-lang/glacier/pull/229), then we(members) can review it on glacier, I think the behavior is useful enough to add ICEs to glacier.

Mark-Simulacrum:
I don't see what relevance there is that glacier checks every day on GHA to how useful the remove command is?

Regardless, it seems like the remove/add commands filing PRs (versus directly editing the repository) is reasonable. Then they can be permissionless, I imagine.

JohnTitor:
Hm, I cannot imagine the needs of `remove` but I won't argue anymore since at least you want that.

Mark-Simulacrum:
Does autofix remove fixed ICEs via PR automatically? That was my envisioned use case for the remove command but if that already happens then there's no point most likely.

JohnTitor:
> Does autofix remove fixed ICEs via PR automatically?

Yeah, autofix does it.

hellow554:
So I will remove the `remove` command, because triagebot will open a new PR instead of commiting directy to the glacier repository?

Mark-Simulacrum:
That sounds correct, yes. So we'll just have the add command which for now takes a URL in quotes (i.e., `@bot glacier add "URL"`).

Elinvynia:
Hello, what is the status of this PR? I'd be interested in helping out.

Mark-Simulacrum:
I believe someone could take the current PR and attempt to finish it, but beyond reviewing PRs I don't currently have the bandwidth to figure out the exact next steps here. Feel free to ping me on Zulip though if you have questions (#t-release/triagebot stream) and I can try to answer them

hellow554:
hey @Elinvynia I haven't developed this any further, but please feel free to take this! I hope this helps to start with the issue. I don't have time for it, sorry :(

hellow554:
Closing in favor of #526

PR #370, opened 2020-02-27
Initiating logging is not an error

Mark-Simulacrum:
cc @dtolnay (please ignore, just generating a ping to test new functionality and can't self-ping as we ignore those)

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -42,7 +42,7 @@ pub fn init() {
     ))
     .unwrap();
     log::set_max_level(log::LevelFilter::Trace);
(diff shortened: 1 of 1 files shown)


src/logger.rs
```diff
@@ -42,7 +42,7 @@ pub fn init() {
     ))
     .unwrap();
     log::set_max_level(log::LevelFilter::Trace);
-    log::error!("initialized logging infra");
+    log::trace!("initialized logging infra");
```
Mark-Simulacrum: (one more time @dtolnay, sorry again :)

PR #433, opened 2020-03-30
Handle r? as assignment

Mark-Simulacrum:
This adds parsing support for handling r? to direct assignment; anyone can do so
on a PR to any org member.

A future commit will add support for r? with a team.

This does not currently auto-assign reviewers like highfive to new PRs, though.

### DIFF
--- a/parser/src/command.rs
+++ b/parser/src/command.rs
@@ -58,43 +58,98 @@ impl<'a> Input<'a> {
         }
     }
 
-    pub fn parse_command(&mut self) -> Command<'a> {
-        let start = match find_commmand_start(&self.all[self.parsed..], self.bot) {
+    fn maybe_parse_review(&mut self) -> Option<(Tokenizer<'a>, assign::AssignCommand)> {
+        // Both `r?` and `R?`.
+        let start = match self.all[self.parsed..]
+            .find("r?")
+            .or_else(|| self.all[self.parsed..].find("R?"))
+        {
             Some(pos) => pos,
-            None => return Command::None,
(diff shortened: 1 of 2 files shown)


parser/src/command.rs
```diff
@@ -58,43 +58,98 @@ impl<'a> Input<'a> {
...
+                let mentions = crate::mentions::get_mentions(w);
+                if let [a] = &mentions[..] {
+                    return Some((
+                        tok,
+                        assign::AssignCommand::User {
+                            username: (*a).to_owned(),
+                        },
+                    ));
+                } else {
+                    log::trace!("{:?} had non-one mention: {:?}", w, mentions);
```
pietroalbini: I think it might be useful to allow more than a person to be assigned: occasionally I write `r? @foo @bar` and manually assign `@bar` once highfive assigns `@foo`. Highfive was written before GitHub allowed multiple assignees, so I guess that's the reason why highfive didn't support that.

Mark-Simulacrum: I largely agree, though I think it is pretty useful to have just one person assigned at a time -- then there's less "well they'll do it" :)

But I don't want to try to make things better before we make things work; migrating things to triagebot is the first priority, and then once that's done we can enhance.

pietroalbini:
Other than that comment this PR looks good!

kellda:
Rebased on master at https://github.com/kellda/triagebot/tree/highfive-review

fmease:
This can be closed since the functionality is now implemented, right?

PR #477, opened 2020-04-21
Add the `@rustbot promote` command

LeSeulArtichaut:
As discussed [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/227806-t-compiler.2Fwg-prioritization/topic/flipping.20regressions).

cc @Mark-Simulacrum

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -501,6 +501,25 @@ pub struct Repository {
     pub full_name: String,
 }
 
+impl Repository {
+    fn url(&self) -> String {
+        format!("https://api.github.com/repos/{}", self.full_name)
+    }
+
+    pub async fn get_issues(&self, client: &GithubClient, label_filter: &[&Label]) -> anyhow::Result<Vec<Issue>> {
+        let label_filter = label_filter
+            .iter()
+            .map(|label| label.name.as_str())
+            .collect::<Vec<&str>>()
+            .join(",");
+        let url = format!("{}/issues?labels={}", self.url(), label_filter);
+        client
+            .json(client.get(&url))
+            .await
+            .with_context(|| format!("failed to list issues from repo {}", self.url()))
+    }
+}
+
 #[derive(Debug)]
 pub enum Event {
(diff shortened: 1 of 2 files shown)


src/github.rs
```diff
@@ -501,6 +501,25 @@ pub struct Repository {
...
+        format!("https://api.github.com/repos/{}", self.full_name)
+    }
+
+    pub async fn get_issues(&self, client: &GithubClient, label_filter: &[&Label]) -> anyhow::Result<Vec<Issue>> {
+        let label_filter = label_filter
+            .iter()
+            .map(|label| label.name.as_str())
+            .collect::<Vec<&str>>()
+            .join(",");
+        let url = format!("{}/issues?labels={}", self.url(), label_filter);
```
Mark-Simulacrum, line 515: We will unfortunately need to support pagination here. I'm not sure if we have something in triagebot yet which does that; please look. If not, we'll need to add it (see github docs).

src/zulip.rs
```diff
@@ -495,3 +504,45 @@ async fn move_notification(
...
+        .iter()
+        .map(|issue| {
+            let mut labels = issue.labels().to_owned();
+            if let Some(pos) = labels.iter().position(|l| l == &past) {
+                labels.remove(pos);
+            }
+            labels.push(current.clone());
+            issue.set_labels(&gh_client, labels)
+        });
+    futures::future::join_all(fut).await;
```
Mark-Simulacrum, line 546: GitHub recommends that we issue requests one at a time -- otherwise they'll aggressively rate limit us. Let's use a combinator of some sort here that does that and spawn it to run asynchronously in the background.

We might look at the APIs, too, I know in the UI you can set a label on a bunch of issues at once. Maybe there's something in the API that can do that as well.

We should make sure we don't block the return from this on the queries completing (we should queue a "post message zulip API call" when it's all done, though).

kellda:
Rebased on master at https://github.com/kellda/triagebot/tree/promote-command

PR #514, opened 2020-05-09
Allow running without database

bjorn3:
This will result in a reduced feature set, but can be useful for local testing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -238,6 +238,11 @@ version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "fallible-streaming-iterator"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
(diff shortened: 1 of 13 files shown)


Mark-Simulacrum:
I think we should be able to avoid most of the cfg's if we keep the dependencies but don't actually connect to the database? Maybe there's not a nice way to create a "fake" database client though which just fails on all actions.

Ideally I would rather not have a bunch of cfg's throughout the codebase and require people editing the code to keep testing it in both --features db and --no-default-features modes to make sure they've properly peppered cfg's everywhere.

But I definitely appreciate that postgres is a pretty heavy dependency to ask for locally. If the previous suggestions don't work, maybe we can instead explore switching to sqlite as the backing database locally? That should be trivial as a dependency (particularly if we, for example, use rusqlite and specify the bundled feature) and would hopefully allow us to mostly "just work" locally almost entirely like in production.

bjorn3:
I won't be working on this anytime soon.

PR #515, opened 2020-05-09
Make error messages easier to read

jyn514:
Erroneous input (from https://github.com/rust-lang/rust/issues/72050):
```
@rustbot modify labels A-diagnostics D-papercut
```

Before:
```
Error: Parsing label command in comment failed: ...ify labels|error: must have : or to as label starter at >| A-diagnos...
```

After:
```
Error: Parsing label command in comment failed: ...'ify labels' | error: must have `:` or `to` as label starter at >| ' A-diagnos'...
```

### DIFF
--- a/parser/src/command/relabel.rs
+++ b/parser/src/command/relabel.rs
@@ -56,8 +56,8 @@ impl fmt::Display for ParseError {
         match self {
             ParseError::EmptyLabel => write!(f, "empty label"),
             ParseError::ExpectedLabelDelta => write!(f, "a label delta"),
-            ParseError::MisleadingTo => write!(f, "forbidden to, use +to"),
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
I am happy to merge this (doing so) but I think it would also be worthwhile to explore doing something like what rustc does via error-snippets crate - probably in a code block for the monospace?

jyn514:
If you're interested in error reporting, you could look into https://docs.rs/codespan-reporting/0.9.3/codespan_reporting/. That's more focused on multi-line errors, not sure if that's what you want.

Mark-Simulacrum:
Hm, I think we don't need that and I'd rather be consistent with what rustc uses as that'll be immediately familiar to everyone using triagebot most likely.

PR #526, opened 2020-05-11
Add `glacier` command to track ICEs

Elinvynia:
This is building on the work done in https://github.com/rust-lang/triagebot/pull/234, huge props to @hellow554 for the initial effort!

I have no idea how to run the bot, but the tests I added pass.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ tokio-postgres = { version = "0.5", features = ["with-chrono-0_4"] }
 postgres-native-tls = "0.3"
 native-tls = "0.2"
 serde_path_to_error = "0.1.2"
-octocrab = { git = "https://github.com/XAMPPRocky/octocrab", branch = "master" }
+octocrab = "0.3"
 
 [dependencies.serde]
 version = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ tokio-postgres = { version = "0.5", features = ["with-chrono-0_4"] }
 postgres-native-tls = "0.3"
 native-tls = "0.2"
 serde_path_to_error = "0.1.2"
+octocrab = "0.2"
 
 [dependencies.serde]
 version = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ tokio-postgres = { version = "0.5", features = ["with-chrono-0_4"] }
 postgres-native-tls = "0.3"
 native-tls = "0.2"
 serde_path_to_error = "0.1.2"
-octocrab = "0.2"
+octocrab = { git = "https://github.com/XAMPPRocky/octocrab", branch = "master" }
 
 [dependencies.serde]
 version = "1"
--- a/parser/src/command/glacier.rs
+++ b/parser/src/command/glacier.rs
@@ -1,6 +1,7 @@
 //! The glacier command parser.
 //!
 //! This adds the option to track ICEs. Do note that the gist must be from a playground link.
+//! The link must also be in quotes.
 //!
 //! The grammar is as follows:
 //!
@@ -31,8 +32,8 @@ impl std::error::Error for ParseError {}
 impl fmt::Display for ParseError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            Self::NoLink => write!(f, "no link provided"),
-            Self::InvalidLink => write!(f, "invalid link"),
+            Self::NoLink => write!(f, "no link provided - did you forget the quotes around it?"),
+            Self::InvalidLink => write!(f, "invalid link - must be from a playground gist"),
         }
     }
 }
--- a/parser/src/command/glacier.rs
+++ b/parser/src/command/glacier.rs
@@ -1,0 +1,116 @@
+//! The glacier command parser.
+//!
+//! This adds the option to track ICEs.
+//!
+//! The grammar is as follows:
+//!
+//! ```text
+//! Command: `@bot glacier add <code-source>?`
+//!
+//! <code-source>:
+//!   - "https://play.rust-lang.org/.*"
+//! ```
+
+use std::fmt;
+use crate::error::Error;
+use crate::token::{Token, Tokenizer};
+use regex::Regex;
+
+#[derive(PartialEq, Eq, Debug)]
+pub struct GlacierCommand {
+    source: String
+}
+
+#[derive(PartialEq, Eq, Debug)]
+pub enum ParseError {
+    NoLink,
+    InvalidLink,
+}
+
+impl std::error::Error for ParseError {}
+
+impl fmt::Display for ParseError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::NoLink => write!(f, "no link provided"),
+            Self::InvalidLink => write!(f, "invalid link"),
+        }
+    }
+}
+
+impl GlacierCommand {
+    pub fn parse<'a>(input: &mut Tokenizer<'a>) -> Result<Option<GlacierCommand>, Error<'a>> {
+        let mut toks = input.clone();
+        if let Some(Token::Word("glacier")) = toks.peek_token()? {
+            toks.next_token()?;
+            match toks.next_token()? {
+                Some(Token::Quote(s)) => {
+                    let source = s.to_owned();
+                    let playground_url = Regex::new("https://play.rust-lang.org/.*").unwrap();
+                    if playground_url.is_match(&source) {
+                        return Ok(Some(GlacierCommand{source}));
+                    } else {
+                        return Err(toks.error(ParseError::InvalidLink));
+                    }
+                }
+                Some(Token::Word(_)) => {
+                    return Err(toks.error(ParseError::InvalidLink));
+                }
+                _ => {
+                    return Err(toks.error(ParseError::NoLink));
+                }
+            }
+
+        } else {
+            Ok(None)
+        }
+    }
+}
+
+
+#[cfg(test)]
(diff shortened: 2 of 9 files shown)


parser/src/command/glacier.rs
```diff
@@ -0,0 +1,116 @@
...
+
+impl GlacierCommand {
+    pub fn parse<'a>(input: &mut Tokenizer<'a>) -> Result<Option<GlacierCommand>, Error<'a>> {
+        let mut toks = input.clone();
+        if let Some(Token::Word("glacier")) = toks.peek_token()? {
+            toks.next_token()?;
+            match toks.next_token()? {
+                Some(Token::Quote(s)) => {
+                    let source = s.to_owned();
+                    let playground_url = Regex::new("https://play.rust-lang.org/.*").unwrap();
```
hellow554, line 49: Why not using [`starts_with`](https://doc.rust-lang.org/std/primitive.str.html#method.starts_with)? I don't like using a regex for this case, because you pull up the crate just this piece and you could circumvent it.

Elinvynia: That's a very good suggestion! It didn't cross my mind, heh.

Elinvynia:
Status update: Currently waiting on https://github.com/XAMPPRocky/octocrab/pull/12 to be merged to allow me to do this cleanly.

LeSeulArtichaut:
@Elinvynia XAMPPRocky/octocrab#12 has been merged

src/handlers/glacier.rs
```diff
@@ -59,6 +61,47 @@ impl Handler for GlacierHandler {
...
+    };
+
+    let url = cmd.source;
+    let response = reqwest::get(&format!("{}{}", url.replace("github", "githubusercontent"), "/playground.rs")).await?;
+    let body = response.text().await?;
+
+    let number = event.issue().unwrap().number;
+    let user = event.user();
+
+    let octocrab = octocrab::instance();
```
LeSeulArtichaut, line 83: As asked by @Mark-Simulacrum, we should not use the static instance here. Let's store it in `Context` alongside the `GithubClient`

src/handlers/glacier.rs
```diff
@@ -59,6 +61,47 @@ impl Handler for GlacierHandler {
...
+
+    let fork = octocrab.repos("rust-lang", "glacier");
```
LeSeulArtichaut, line 85: I assume ~~we~~ @Mark-Simulacrum will create a fork for the bot, so to not pollute the main `rust-lang/glacier` repository? In that case you just need to set the owner to `"rustbot"` instead of `"rust-lang"`

src/handlers/glacier.rs
```diff
@@ -59,6 +61,47 @@ impl Handler for GlacierHandler {
...
+
+    let master = base.get_ref(&Reference::Branch("master".to_string())).await?.object;
+    let master = if let Object::Commit { sha, ..} = master {
+        sha
+    } else {
+        panic!()
+    };
+
+    fork.create_ref(&Reference::Branch(format!("triagebot-ice-{}", number)), master).await?;
+    fork.create_file(format!("ices/{}.rs", number), format!("This PR was created by {} on issue {}.", user.login, number), body)
```
LeSeulArtichaut, line 96: This message is actually the commit message. You probably want to change that, to something like "Add ICE reproduction for issue #{}"?

src/handlers/glacier.rs
```diff
@@ -59,6 +61,47 @@ impl Handler for GlacierHandler {
...
+        .branch("triagebot-ice")
```
LeSeulArtichaut, line 97: Use the branch you created earlier.
```suggestion
        .branch(format!("triagebot-ice-{}", number))
```

src/handlers/glacier.rs
```diff
@@ -59,6 +61,47 @@ impl Handler for GlacierHandler {
...
+        .send()
+        .await?;
+
+    octocrab.pulls("rust-lang", "glacier")
+        .create(format!("ICE - {}", number), format!("triagebot-ice-{}", number), "master")
+        .body("This is a fake new catastrophic avalanche of ICE!")
```
LeSeulArtichaut, line 103: This is the body (description) of the PR. Let's change it to something like "Automatically created by @{} in issue #{}"

src/handlers/glacier.rs
```diff
@@ -59,6 +61,47 @@ impl Handler for GlacierHandler {
...
+    };
+
+    fork.create_ref(&Reference::Branch(format!("triagebot-ice-{}", number)), master).await?;
+    fork.create_file(format!("ices/{}.rs", number), format!("This PR was created by {} on issue {}.", user.login, number), body)
+        .branch("triagebot-ice")
+        .send()
+        .await?;
+
+    octocrab.pulls("rust-lang", "glacier")
+        .create(format!("ICE - {}", number), format!("triagebot-ice-{}", number), "master")
```
LeSeulArtichaut, line 102: If we actually create a fork for rustbot, you'll need to set this to `"rustbot:triagebot-ice-{}"`

src/handlers/glacier.rs
```diff
@@ -59,6 +61,47 @@ impl Handler for GlacierHandler {
...
+    } else {
+        true
+    };
+
+    if !is_team_member {
+        return Ok(())
+    };
+
+    let url = cmd.source;
+    let response = reqwest::get(&format!("{}{}", url.replace("github", "githubusercontent"), "/playground.rs")).await?;
```
LeSeulArtichaut, line 77: Use the `reqwest::Client` inside of the `GithubClient` using `GithubClient::raw`

parser/src/command/glacier.rs
```diff
@@ -0,0 +1,113 @@
...
+    NoLink,
+    InvalidLink,
+}
+
+impl std::error::Error for ParseError {}
+
+impl fmt::Display for ParseError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::NoLink => write!(f, "no link provided"),
```
Mark-Simulacrum, line 34: We should note the requirement for quotes around the link.

src/handlers/glacier.rs
```diff
@@ -0,0 +1,107 @@
...
+        let body = if let Some(b) = event.comment_body() {
+            b
+        } else {
+            // not interested in other events
+            return Ok(None);
+        };
+
+        match event {
+            Event::IssueComment(_) => (),
+            _ => {return Ok(None);}
```
Mark-Simulacrum, line 36: We probably want to run on newly opened issues too, not just comments?

src/handlers/glacier.rs
```diff
@@ -0,0 +1,107 @@
...
+        };
+
+        let mut input = Input::new(&body, &ctx.username);
+        match input.parse_command() {
+            Command::Glacier(Ok(command)) => Ok(Some(command)),
+            Command::Glacier(Err(err)) => {
+                return Err(format!(
+                    "Parsing assign command in [comment]({}) failed: {}",
```
Mark-Simulacrum, line 44: nit: glacier, not assign command.

src/handlers/glacier.rs
```diff
@@ -0,0 +1,107 @@
...
+    let octocrab = &ctx.octocrab;
+
+    let fork = octocrab.repos("rustbot", "glacier");
+    let base = octocrab.repos("rust-lang", "glacier");
+
+    let master = base.get_ref(&Reference::Branch("master".to_string())).await?.object;
+    let master = if let Object::Commit { sha, ..} = master {
+        sha
+    } else {
+        panic!()
```
Mark-Simulacrum, line 92: Please add the actual value of master here so that we can debug if it does occur.

src/handlers/glacier.rs
```diff
@@ -0,0 +1,107 @@
...
+    };
+
+    fork.create_ref(&Reference::Branch(format!("triagebot-ice-{}", number)), master).await?;
```
Mark-Simulacrum, line 95: Will this fail if the branch already exists? Sometimes we have multiple distinct examples for a single ICE.

src/handlers/glacier.rs
LeSeulArtichaut: Well I think this should be `unreachable!()` instead.

Mark-Simulacrum: Regardless, we should note what we actually encountered.

src/handlers/glacier.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+async fn handle_input(ctx: &Context, event: &Event, cmd: GlacierCommand) -> anyhow::Result<()> {
+    let is_team_member = if let Err(_) | Ok(false) = event.user().is_team_member(&ctx.github).await
+    {
+        false
+    } else {
+        true
+    };
```
XAMPPRocky, line 65:
```suggestion
    let is_team_member = event.user().is_team_member(&ctx.github).await.unwrap_or(false);
```

src/handlers/glacier.rs
Elinvynia: Honesly I am not sure what happens here, how should we proceed? Append a random string and hope for no conflict?

Mark-Simulacrum: Seems fine to leave as is for now.

Cargo.toml
```diff
@@ -32,6 +32,7 @@ tokio-postgres = { version = "0.5", features = ["with-chrono-0_4"] }
 postgres-native-tls = "0.3"
 native-tls = "0.2"
 serde_path_to_error = "0.1.2"
+octocrab = { git = "https://github.com/XAMPPRocky/octocrab", branch = "master" }
```
XAMPPRocky, line 35:
```suggestion
octocrab = "0.3.1"
```

PR #580, opened 2020-05-30
Fixes glacier command

Elinvynia:
This updates octocrab to fix the issue mentioned in: https://github.com/rust-lang/glacier/issues/231
Also, it fixes my mistake in parsing the URL. This also allows ANY gist to be used instead of just a playground link.

### DIFF
--- a/src/handlers/glacier.rs
+++ b/src/handlers/glacier.rs
@@ -63,8 +63,7 @@ async fn handle_input(ctx: &Context, event: &Event, cmd: GlacierCommand) -> anyh
         return Ok(())
     };
 
-    let url = cmd.source;
-    let response = ctx.github.raw().get(&format!("{}{}", url, "/raw")).send().await?;
+    let response = ctx.github.raw().get(&cmd.source).send().await?;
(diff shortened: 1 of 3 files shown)


src/handlers/glacier.rs
```diff
@@ -64,7 +64,7 @@ async fn handle_input(ctx: &Context, event: &Event, cmd: GlacierCommand) -> anyh
     };
 
     let url = cmd.source;
-    let response = ctx.github.raw().get(&format!("{}{}", url.replace("github", "githubusercontent"), "/playground.rs")).send().await?;
+    let response = ctx.github.raw().get(&format!("{}{}", url, "/raw")).send().await?;
```
Mark-Simulacrum, line 67: Maybe it makes sense to just directly send a request to the relevant URL? That would give more flexibility and be easier to understand I think-- just saying "please provide a URL resolving to plain-text rust code" makes a lot of sense I think.

Elinvynia: Good idea, I'll change it!

PR #648, opened 2020-06-24
More detailed Zulip announcement message for major changes

Mark-Simulacrum:
r? @nikomatsakis from https://rust-lang.zulipchat.com/#narrow/stream/185694-t-compiler.2Fwg-meta/topic/text.20of.20the.20zulip.20stream.20comment

### DIFF
--- a/src/handlers/major_change.rs
+++ b/src/handlers/major_change.rs
@@ -127,11 +127,18 @@ async fn handle_input(
                 cmnt.post(&ctx.github).await?;
                 return Ok(());
             }
-            (format!(
-                "A new proposal has been announced: [#{}]({}). It will be brought up at the next meeting.",
-                issue.number,
-                event.html_url().unwrap()
-            ), config.meeting_label.clone())
+            (
+                format!(
+                    "A new proposal has been announced: [#{}]({}). It will be
+                announced at the next meeting to try and draw attention to it,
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
Not sure if you have rights to merge, feel free to if so. I'm personally fine with it as is, though a link to proposing meetings might be nice, but requires extracting this message to configuration since it would differ between compiler and lang teams.

src/handlers/major_change.rs
```diff
@@ -127,11 +127,18 @@ async fn handle_input(
...
-                issue.number,
-                event.html_url().unwrap()
-            ), config.meeting_label.clone())
+            (
+                format!(
+                    "A new proposal has been announced: [#{}]({}). It will be
+                announced at the next meeting to try and draw attention to it,
+                but usually MCPs are not discussed during triage meetings. If
+                you think this would benefit from discussion amongst the
+                compiler team, consider proposing a design meeting.",
```
nikomatsakis, line 136: this mentions compiler team, but is this meant to be generic for both teams?

Mark-Simulacrum: Fixed, good catch. Also spotted another case of this (the team to ping on seconding) and added configuration for that (and pushed updates to lang and compiler team repos).

PR #653, opened 2020-06-28
Enhancements around Zulip APIs

LeSeulArtichaut:
This PR proposes some enhancements for sending and receiving messages, and adds a new `link` commands which sends back a link to the current channel (topic or private message).

- The Zulip `MessageAPIRequest` is now more statically typed and uses an enum to determine the type of recipient, i.e. `"private"` or `"stream"`.
- The `Message` data structure now parses more information, such as the `stream_id` and `topic` of the message. I didn't test this but only tried to construct a valid data structure from the [Zulip API documentation examples](https://www.zulipchat.com/api/get-events#example-response). I think a second look is necessary.

The `link` command is meant as a helper for e.g. mobile clients which have no easy way to determine the link to a topic. Its implementation could also be reused in the context of other commands.

cc @Mark-Simulacrum

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -21,9 +21,14 @@ pub struct Request {
 
 #[derive(Debug, serde::Deserialize)]
 struct Message {
-    sender_id: usize,
+    sender_id: u64,
+    recipient_id: u64,
     sender_short_name: String,
     sender_full_name: String,
+    stream_id: Option<u64>,
+    topic: Option<String>,
+    #[serde(rename = "type")]
+    type_: String,
 }
 
 #[derive(serde::Serialize, serde::Deserialize)]
@@ -63,7 +68,7 @@ pub async fn respond(ctx: &Context, req: Request) -> String {
     }
 
     log::trace!("zulip hook: {:?}", req);
-    let gh_id = match to_github_id(&ctx.github, req.message.sender_id).await {
+    let gh_id = match to_github_id(&ctx.github, req.message.sender_id as usize).await {
         Ok(Some(gh_id)) => Ok(gh_id),
         Ok(None) => Err(serde_json::to_string(&Response {
             content: &format!(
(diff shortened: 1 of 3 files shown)


src/zulip.rs
```diff
@@ -303,48 +305,94 @@ struct MembersApiResponse {
...
-        let always_safe = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_.-~";
-        let mut encoded_topic = Vec::new();
-        for ch in self.topic.expect("topic").bytes() {
-            if !(always_safe.contains(ch as char) || ch == b'.') {
-                write!(encoded_topic, "%{:02X}", ch).unwrap();
-            } else {
-                encoded_topic.push(ch);
-            }
+impl<'a> MessageApiRequest<'a> {
+    pub fn private(user_id: u64, user_email: &'a str, content: &'a str) -> Self {
```
Mark-Simulacrum, line 368: I am not a fan of this as it has two &str fields without much protection for getting it right, and the u64 is also not obviously just one thing. I would prefer that we construct the ApiRequest directly at the call sites for both of these constructor functions.

src/zulip.rs
```diff
@@ -520,3 +568,21 @@ async fn move_notification(
         .unwrap()),
     }
 }
+
+fn message_channel_link(msg: &Message) -> String {
```
Mark-Simulacrum, line 572: Can we at least split this addition to a separate commit? I'm also not clear on the use case--Zulip makes it pretty easy to get links, no?

LeSeulArtichaut: Here were my thoughts on this (from this PR's comment):

> The `link` command is meant as a helper for e.g. mobile clients which have no easy way to determine the link to a topic. Its implementation could also be reused in the context of other commands.

LeSeulArtichaut: Removed it and will add it back later

PR #657, opened 2020-06-29
Print repository name along the issue number

apiraino:
Closes https://github.com/rust-lang/compiler-team-prioritization/issues/4

This is the least invasive patch I could think of to show the issue as described [here](https://github.com/rust-lang/compiler-team-prioritization/issues/4#issuecomment-650560817), prepending the repository name before the issue number.

```
[rust#13](https://github.com/rust-lang/rust/issues/13)
[compiler-team#143](https://github.com/rust-lang/compiler-team/issues/143)
```

@spastorino opinions?

Thanks for the review!

### DIFF
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -31,6 +31,7 @@ pub struct IssueDecorator {
     pub number: u64,
     pub title: String,
     pub html_url: String,
+    pub repo_name: String,
     pub labels: String,
     pub assignees: String,
 }
@@ -75,6 +76,13 @@ impl<'a> Action for Step<'a> {
                                         title: issue.title.clone(),
                                         number: issue.number,
                                         html_url: issue.html_url.clone(),
+                                        repo_name: repository
+                                            .full_name
+                                            .clone()
+                                            .split("/")
+                                            .last()
+                                            .expect("Failed to split repository name")
+                                            .to_string(),
                                         labels: issue
                                             .labels
                                             .iter()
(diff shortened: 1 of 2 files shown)


src/actions.rs
```diff
@@ -75,6 +76,13 @@ impl<'a> Action for Step<'a> {
                                         title: issue.title.clone(),
                                         number: issue.number,
                                         html_url: issue.html_url.clone(),
+                                        repo_name: repository
+                                            .full_name
+                                            .clone()
```
Mark-Simulacrum, line 81: You shouldn't need to clone this -- split operates on &str, not String.

spastorino: I think it depends on how do we want to see this issue. If we want to think of `IssueDecorator` as a self contained object I'd say it's fine to clone as it's already done for the other fields. On the other hand, it's true that `IssueDecorator` decorates `Issue` so, in some sense the true source of data is issue, so we can make everything a reference. title, html_url for sure and the repository full_name comes from the Repository object but still I think it can be a reference without any trouble.

spastorino: After commenting this gave a look to the code and what I've said won't work. Repository is created here https://github.com/rust-lang/triagebot/blob/0b8db52370391a521047f12b0d2a4a89cc563a75/src/actions.rs#L62-L64 and the decorator is later used here https://github.com/rust-lang/triagebot/blob/0b8db52370391a521047f12b0d2a4a89cc563a75/src/actions.rs#L133-L135

spastorino: Of course the code can be structured differently to make this work but unsure how much is worth.

Mark-Simulacrum: My point is that the clone isn't needed; the to_string is.

spastorino: Ahhh hehehe, I didn't realize there was a `to_string` call at the end. :+1:

PR #709, opened 2020-07-18
Revert "Cleanup and refactor handlers"

Mark-Simulacrum:
Reverts rust-lang/triagebot#686

Seems to be causing problems for us (unintentional spam) so let's revert for now.

### DIFF
--- a/parser/src/error.rs
+++ b/parser/src/error.rs
@@ -5,7 +5,7 @@ use std::fmt;
 pub struct Error<'a> {
     pub input: &'a str,
     pub position: usize,
-    pub source: Box<dyn error::Error + Send>,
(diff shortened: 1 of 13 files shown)


kellda:
First of all sorry for this bug. For reference it can be seen on docs.rs#158

It is due to the fact that with my refactor, the configuration is checked first, event when the bot doesn't recieve any command. Futhermore, each error message generate an event,  which generate the next error message.

kellda:
@Mark-Simulacrum is there a way to unlock https://github.com/rust-lang/docs.rs/issues/158 now that the rustbot's spam issue is fixed ?

jyn514:
@kelida I unlocked the issue.

PR #728, opened 2020-07-26
Prevent Authorization header from being logged

Mark-Simulacrum:
This logs the GitHub request token into our AWS logs, which is not great. That
said this is not really all that important because AWS access is basically
"keys to the kingdom" already.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -20,11 +20,16 @@ pub struct User {
 impl GithubClient {
     async fn _send_req(&self, req: RequestBuilder) -> anyhow::Result<(Response, String)> {
         const MAX_ATTEMPTS: usize = 2;
-        log::debug!("_send_req with {:?}", req);
         let req_dbg = format!("{:?}", req);
-        let req = req
+        let mut req = req
             .build()
             .with_context(|| format!("building reqwest {}", req_dbg))?;
+
+        let auth = req.headers_mut().remove(AUTHORIZATION);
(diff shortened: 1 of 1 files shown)


src/github.rs
```diff
@@ -20,12 +20,17 @@ pub struct User {
...
-        let req = req
+        let mut req = req
             .build()
             .with_context(|| format!("building reqwest {}", req_dbg))?;
 
+        let auth = req.headers_mut().remove(AUTHORIZATION);
+        log::debug!("_send_req with {:?}", auth);
+        if let Some(auth) = auth {
+            req.headers_mut().insert(AUTHORIZATION, auth);
+        }
```
pietroalbini, line 32: Instead of removing and then adding back the header, you can mark its value as sensitive to prevent it from ever being included in the debug representation:

```rust
if let Some(auth) = req.headers_mut().get_mut(AUTHORIZATION) {
    auth.set_sensitive(true);
}
```

[Example of the output](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b0478b92767f1adcd11efac99500a13a)

Mark-Simulacrum: Hm, okay, we should try to do that when we add it to the request builder I guess. I didn't know about the sensitive flag!

I'm somewhat surprised it's not set by default for AUTHORIZATION, but I guess it's on the value not the key.

I'll look into doing this soon.

PR #742, opened 2020-08-02
Track PR number of bors merges

Mark-Simulacrum:
This enables tooling like perf to post back comments on additional results.


Mark-Simulacrum:
cc https://github.com/rust-lang/rustc-perf/issues/711 -- we'll need this PR to land so that we can easily get the PRs to post back to in all cases

PR #777, opened 2020-08-24
Edit template to avoid some repetitions

apiraino:
Should fix https://github.com/rust-lang/compiler-team-prioritization/issues/10

not the most elegant solution, but I tried not to touch the templates too much.

Old text
```
  - No pending FCP requests on compiler-team repo this time.
  - No pending FCP requests on rust repo this time.
  - No pending FCP requests on forge repo this time.
```

new rendering
```
  - No pending FCP requests this time.
```

and equivalent for other FCP and MCP (cannot test everything until github returns nothing)

@spastorino how does it look?

### DIFF
--- a/templates/_issues.tt
+++ b/templates/_issues.tt
@@ -2,6 +2,5 @@
 
 {% macro render(issues, indent="", empty="No issues this time.") %}
 {%- for issue in issues %}
-{{indent}}- {{issue::render(issue=issue)}}{% else %}
-{{indent}}- {{empty}}{% endfor -%}
+{{indent}}- {{issue::render(issue=issue)}}{% else %}{{empty}}{% endfor -%}
 {% endmacro %}
--- a/templates/_issues.tt
+++ b/templates/_issues.tt
@@ -1,6 +1,7 @@
 {% import "_issue.tt" as issue %}
 
-{% macro render(issues, indent="", empty="No issues this time.") %}
+{% macro render(issues, indent="", newline_and_indent=true, empty="No issues this time.") %}
 {%- for issue in issues %}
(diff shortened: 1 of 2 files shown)


apiraino:
self note: when merged [this line](https://github.com/rust-lang/rust-forge/blame/master/src/compiler/prioritization/procedure.md#L117) can go away

Mark-Simulacrum:
This is waiting on approval from @spastorino.

spastorino:
This is not exactly how I'd have imagine fixing this problem. What I would have done is just have one call that returns the whole collection of issues. We would need to make a call to our logic that contacts all those three queries into just one list.

On the other hand, we can also merge this that fixes the issue and try to find a more elegant solution later.
So for now I'm :+1: with this :).

spastorino (approved):

spastorino:
Having a second thought now ... @apiraino does this break new lines separations and such in some way?.

apiraino:
> I would have done is just have one call that returns the whole collection of issues

@spastorino That's the approach I had discarded but I'm ok going that way and work on something more invasive but more structured, not a big deal.

> Having a second thought now ... @apiraino does this break new lines separations and such in some way?.

I've tested with this weeks agenda (2020-08-27) and didnt notice anything weird happening when rendering. Pleae let me know if something slipped my checks.

spastorino:
Let's merge this as it solves the existing issue and we can work on a better solution afterwards. cc @Mark-Simulacrum

spastorino:
We would need to revert this as it doesn't really work. @apiraino is already aware of it. We should go directly to the real solution :).

PR #807, opened 2020-09-06
Avoid empty lines repetitions in FCP templates (take 2)

apiraino:
Second attempt to close rust-lang/compiler-team-prioritization#10

I tried applying a conservative approach, probably some details can be improved.

Happy to receive suggestions!

r? @spastorino

### DIFF
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -48,6 +48,10 @@ lazy_static! {
 }
 
 // Group all the pending FCP for these teams
+pub const IN_PRE_FCP: &str = "in_pre_fcp";
+pub const IN_FCP: &str = "in_fcp";
+pub const FCP_FINISHED: &str = "fcp_finished";
+
 pub const IN_PRE_FCP_COMPILER_TEAM: &str = "in_pre_fcp_compiler_team";
 pub const IN_PRE_FCP_RUST: &str = "in_pre_fcp_rust";
 pub const IN_PRE_FCP_FORGE: &str = "in_pre_fcp_forge";
@@ -121,11 +125,11 @@ impl<'a> Action for Step<'a> {
 
                                 // group query results for multiline FCP and add them later
                                 if PENDINGFCP.contains(name) {
-                                    all_pending_fcp.push(("in_pre_fcp", issues_decorator));
+                                    all_pending_fcp.push((IN_PRE_FCP, issues_decorator));
                                 } else if THINGSINFCP.contains(name) {
-                                    all_things_in_fcp.push(("in_fcp", issues_decorator));
+                                    all_things_in_fcp.push((IN_FCP, issues_decorator));
                                 } else if FINALIZEDFCP.contains(name) {
-                                    all_finalized_fcp.push(("fcp_finished", issues_decorator));
+                                    all_finalized_fcp.push((FCP_FINISHED, issues_decorator));
                                 } else {
                                     context.insert(*name, &issues_decorator);
                                 }
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -47,12 +47,38 @@ lazy_static! {
     };
 }
 
+// Group all the pending FCP for these teams
+pub const IN_PRE_FCP_COMPILER_TEAM: &str = "in_pre_fcp_compiler_team";
+pub const IN_PRE_FCP_RUST: &str = "in_pre_fcp_rust";
+pub const IN_PRE_FCP_FORGE: &str = "in_pre_fcp_forge";
+const PENDINGFCP: [&str; 3] = [IN_PRE_FCP_COMPILER_TEAM, IN_PRE_FCP_RUST, IN_PRE_FCP_FORGE];
+
+// Group all the FCP for these teams
+pub const IN_FCP_COMPILER_TEAM: &str = "in_fcp_compiler_team";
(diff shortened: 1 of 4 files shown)


src/actions.rs
```diff
@@ -47,13 +47,39 @@ lazy_static! {
...
+
+// Group all the Finalized FCP for these teams
+pub const FCP_FINISHED_COMPILER_TEAM: &str = "fcp_finished_compiler_team";
+pub const FCP_FINISHED_RUST: &str = "fcp_finished_rust";
+pub const FCP_FINISHED_FORGE: &str = "fcp_finished_forge";
+const FINALIZEDFCP: [&str; 3] = [
+    FCP_FINISHED_COMPILER_TEAM,
+    FCP_FINISHED_RUST,
+    FCP_FINISHED_FORGE,
+];
```
apiraino, line 70: All the github queries names for FCPs are now listed here, using constants to avoid typos .

templates/prioritization_agenda.tt
```diff
@@ -15,19 +15,13 @@ tags: weekly, rustc
 - Old MCPs (not seconded, take a look)
 {{-issues::render(issues=mcp_old_not_seconded, indent="  ", empty="No old proposals this time.")}}
 - Pending FCP requests (check your boxes!)
-{{-issues::render(issues=in_pre_fcp_compiler_team, indent="", newline_and_indent=false, empty="")}}
-{{-issues::render(issues=in_pre_fcp_rust, indent="", newline_and_indent=false, empty="")}}
-{{-issues::render(issues=in_pre_fcp_forge, indent="  ", empty="No pending FCP requests this time.")}}
+{{-issues::render(issues=in_pre_fcp, indent="  ", empty="No pending FCP requests this time.")}}
```
apiraino, line 18: The FCP template now is one element (each FCP status) with all the issues together, so I can effortlessy remove empty lines. Only issue here, we might lose the sorting (compiler_team, rust, forge). But I think we can live without it.

src/actions.rs
```diff
@@ -93,7 +119,16 @@ impl<'a> Action for Step<'a> {
...
                                     .collect();
 
-                                context.insert(*name, &issues_decorator);
+                                // group query results for multiline FCP and add them later
+                                if PENDINGFCP.contains(name) {
+                                    all_pending_fcp.push(("in_pre_fcp", issues_decorator));
+                                } else if THINGSINFCP.contains(name) {
+                                    all_things_in_fcp.push(("in_fcp", issues_decorator));
+                                } else if FINALIZEDFCP.contains(name) {
+                                    all_finalized_fcp.push(("fcp_finished", issues_decorator));
```
spastorino, line 128: Wouldn't we also want constants for these ones?, "in_pre_fcp", "in_fcp" and "fcp_finished"

templates/prioritization_agenda.tt
spastorino: I don't think we will lose the sorting.

spastorino (commented):
@apiraino great Antonio!!! at some point in the future we may also want to change how Queries work on multiple repos and by doing that not only will be cleaner but also we could remove this https://github.com/rust-lang/triagebot/pull/807/files#diff-d2f2b5e57a90bd802ffeef1fdd498c52R122-R128, https://github.com/rust-lang/triagebot/pull/807/files#diff-d2f2b5e57a90bd802ffeef1fdd498c52R163-R172 and just fire one of this three queries https://github.com/rust-lang/triagebot/pull/807/files#diff-d1190d2bb14248e42ee716873ff00136R31.

Anyway, I'd just merge this one after you add the constants.

src/actions.rs
apiraino: why not :+1: although they're only used in the template

apiraino:
> at some point in the future we may also want to change how Queries work on multiple repos 

completely agree. I've explicitly avoided getting deep into the guts of this tool, but a refactor of the Query/QueryMap would allo getting rid of more code.

apiraino:
@spastorino would you evaluate another PR after this one the fix all the clippy lints? I'm used to have clippy on by default and in this code there are ~40 warning, they kind of get in the way when reading the compiler output.

Mark-Simulacrum:
I personally am usually not *entirely* happy with fixing all clippy lints, since I disagree with some of them, but would be happy to review a PR that fixes them -- and would accept adding some allows to the root of the crate.

apiraino:
> I personally am usually not /entirely/ happy with fixing all clippy
> lints, since I disagree with some of them, but would be happy to review
> a PR that fixes them -- and would accept adding some allows to the root
> of the crate.

I would be *really* happy to receive feedback and mentoring! My usual
approach is to accept clippy and rustfmt defaults because set by people
that know better than me, therefore a discussion on some lints would be
absolutely helpful.

spastorino (approved):

spastorino:
> > at some point in the future we may also want to change how Queries work on multiple repos
> 
> completely agree. I've explicitly avoided getting deep into the guts of this tool, but a refactor of the Query/QueryMap would allo getting rid of more code (and my dirty hacks).

Right, we can leave that for a future PR.

spastorino:
@Mark-Simulacrum this can be merged now.

PR #811, opened 2020-09-07
Synchronize GitHub Releases with the in-repo changelog

pietroalbini:
This PR adds a new handler to triagebot that synchronizes the GitHub Releases with the changelog present inside a repository itself. It's meant to be used for rustc, but it's designed to be easily extensible. Namely, I want to implement support for [keep a changelog](https://keepachangelog.com/en/1.0.0/) to synchronize [rustwide's changelog](https://github.com/rust-lang/rustwide/blob/master/CHANGELOG.md) in a future PR.

The handler can be enabled by adding this snippet to a `triagebot.toml`:

```toml
[github-releases]
format = "rustc"
project-name = "Rust"
changelog-path = "RELEASES.md"
changelog-branch = "master"
```

The handler will run when either a new tag is pushed, or when a new commit on the `changelog-branch` branch is pushed. It will create releases named `"{project-name} {tag}"` with the relevant section of the changelog as their body. It will also automatically update old releases if the changelog was updated after the fact.

The only supported format right now is `rustc`, which maps to rustc's RELEASES.md file. It's possible to add new formats by changing the `src/changelogs/` module, so this handle can be used for other Rust projects too.

You can see the result of running this in [my fork of rustc](https://github.com/pietroalbini/rust/releases).

This PR is best reviewed commit-by-commit.
r? @Mark-Simulacrum

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ postgres-native-tls = "0.3"
 native-tls = "0.2"
 serde_path_to_error = "0.1.2"
 octocrab = "0.5"
+comrak = "0.8.2"
 
 [dependencies.serde]
 version = "1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,6 +25,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ee49baf6cb617b853aa8d93bf420db2383fab46d314482ca2803b40d5fde979b"
+dependencies = [
+ "winapi 0.3.9",
+]
+
+[[package]]
 name = "anyhow"
 version = "1.0.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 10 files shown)


pietroalbini:
A future improvement would be to check the diff and only run this on push if the `changelog-path` file is changed, instead of every push.

Cargo.toml
```diff
@@ -34,6 +34,7 @@ postgres-native-tls = "0.3"
 native-tls = "0.2"
 serde_path_to_error = "0.1.2"
 octocrab = "0.5"
+comrak = "0.8.2"
```
Mark-Simulacrum, line 37: Can we avoid using a different parser here? parser uses pulldown-cmark currently.

Mark-Simulacrum (commented):
Looks good overall.

Cargo.toml
pietroalbini: From what I can see `pulldown-cmark` doesn't have an AST, and also doesn't have a way to generate markdown as output instead of HTML :(

Mark-Simulacrum: Sounds good. Seems fine then.

Mark-Simulacrum:
@pietroalbini to confirm -- this should be good to merge, right? No outstanding problems?

pietroalbini:
There are no outstanding problems I can see, no.

pietroalbini:
[Wrote documentation on how to use this.](https://github.com/rust-lang/triagebot/wiki/GitHub-Releases)

PR #846, opened 2020-09-22
Add missing space to message

camelid:


src/handlers/ping.rs
```diff
@@ -40,8 +40,8 @@ pub(super) async fn handle_command(
             let cmnt = ErrorComment::new(
                 &event.issue().unwrap(),
                 format!(
-                    "This team (`{}`) cannot be pinged via this command;\
-                 it may need to be added to `triagebot.toml` on the master branch.",
+                    "This team (`{}`) cannot be pinged via this command; \
+                    it may need to be added to `triagebot.toml` on the master branch.",
```
camelid, line 44: What does "on the master branch" mean exactly? Should we just remove it?

```suggestion
                    it may need to be added to `triagebot.toml`.",
```

Mark-Simulacrum: There is no such thing in git as a canonical triagebot.toml; triagebot currently always reads from the branch called master when running some operation.

Removing it may make sense; in the long run I imagine that we will likely use the default branch for this (or even a dedicated separate branch, but that seems unlikely). I don't really care -- but I think we should not do so in this PR.

camelid: Okay, I'll mark this as resolved.

PR #871, opened 2020-10-03
Allow acknowledging all notifications

### LINKED ISSUE #849: Allow acknowledging all notifications
It would probably be helpful for people who get a ton of notifications and want to acknowledge all of them so they can start fresh.

@rustbot modify labels: enhancement

camelid:
Closes #849.

Now you can send `ack all` or `ack *` to triagebot and all your
notifications will *poof* disappear!

### DIFF
--- a/src/db/notifications.rs
+++ b/src/db/notifications.rs
@@ -59,9 +59,6 @@ pub async fn delete_ping(
                 )
                 .await
                 .context("delete notification query")?;
-            if rows.is_empty() {
-                anyhow::bail!("Did not delete any notifications");
-            }
             Ok(rows
                 .into_iter()
                 .map(|row| {
@@ -150,9 +147,6 @@ pub async fn delete_ping(
                 )
                 .await
                 .context("delete all notifications query")?;
-            if rows.is_empty() {
-                anyhow::bail!("Did not delete any notifications");
-            }
             Ok(rows
                 .into_iter()
                 .map(|row| {
--- a/src/db/notifications.rs
+++ b/src/db/notifications.rs
@@ -40,6 +40,8 @@ pub async fn record_ping(db: &DbClient, notification: &Notification) -> anyhow::
 pub enum Identifier<'a> {
     Url(&'a str),
     Index(std::num::NonZeroUsize),
+    /// Glob identifier (`all` or `*`).
+    All,
 }
 
 pub async fn delete_ping(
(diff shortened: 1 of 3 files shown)


camelid:
Reminder to self: update forge once merged

src/db/notifications.rs
```diff
@@ -139,6 +141,36 @@ pub async fn delete_ping(
...
+            let rows = db
+                .query(
+                    "DELETE FROM notifications WHERE user_id = $1
+                        RETURNING origin_url, origin_html, time, short_description, metadata",
+                    &[&user_id],
+                )
+                .await
+                .context("delete all notifications query")?;
+            if rows.is_empty() {
+                anyhow::bail!("Did not delete any notifications");
```
Mark-Simulacrum, line 154: I don't think this should be an error return -- does our caller not handle an empty vector appropriately?

camelid: Hmm, I don't know; I just copied the code for `ack`ing a single notification and modified it. I'll look into it.

camelid: Okay, here's the caller logic:

https://github.com/rust-lang/triagebot/blob/e60ffaddae21a8ad5763e4c8af3750c7151ae422/src/zulip.rs#L490-L511

Basically, if there were no notifications found, it will print:

> Acknowledged:

without any messages, which looks a bit confusing. So I think it might make sense to keep the error. What do you think?

camelid: Also it seems like maybe it would make sense to set a flag as to whether the notification was acknowledged, rather than deleting it? Although it doesn't matter too much.

Mark-Simulacrum: I think we should change the caller then.

I don't want to keep them around indefinitely.

src/zulip.rs
```diff
@@ -491,20 +491,26 @@ async fn acknowledge(gh_id: i64, mut words: impl Iterator<Item = &str>) -> anyho
...
-                        .as_deref()
-                        .unwrap_or(&deleted.origin_url),
-                    deleted.origin_url,
-                    deleted
-                        .metadata
-                        .map_or(String::new(), |m| format!(" ({})", m)),
-                ));
-            }
+            let resp = if deleted.is_empty() {
+                String::from("You have no pending notifications, so none were deleted.")
```
Mark-Simulacrum, line 495: This is misleading. You could have pending notifications, just none that matched the passed filter.

camelid: Oops, thanks for catching that. I forgot that this runs for everything, not just glob `ack`s.

camelid:
Thanks!

Issue #892, opened 2020-10-08
Make label a one-word command


tmandry:
I tend to find `@rustbot modify labels to` unwieldy, both to remember and type. At this point I've typed it enough times that I probably won't forget, but I remember having to look up docs several times to remember the incantation I'm supposed to use.

I think `@rustbot label +foo` is just as self-explanatory and easier to remember.

Could we add this as a synonym?

Mark-Simulacrum:
I would be happy to take a PR, should be a matter of modifying https://github.com/rust-lang/triagebot/blob/master/parser/src/command/relabel.rs#L130.

PR #909, opened 2020-10-15
Add alias from `@bot modify labels` to `@bot label`

jechasteen:
I've been participating in the discussions at wg-prioritization, and we use `@rustbot modify labels` a lot.
#892 made a great suggestion for an alias which this PR attempts to support as, for example

`@rustbot label +bug`

which is the equivalent of

`@rustbot modify labels +bug`

### DIFF
--- a/parser/src/command.rs
+++ b/parser/src/command.rs
@@ -197,6 +197,13 @@ fn code_2() {
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


camelid:
Hmm, why was this closed?

PR #911, opened 2020-10-16
Fix misspelling of "command"

kraai:


apiraino:
thanks @kraai 

cc: @spastorino 

(sometimes PRs pass under the radar)

PR #914, opened 2020-10-17
Allow MCP "consumers" to add extra text to MCP comment

LeSeulArtichaut:
This would allow e.g. T-compiler to ping compiler contributors on GitHub whenever an MCP is created, as discussed in rust-lang/compiler-team#369.

@Mark-Simulacrum @varkor does this seem like a viable approach?

### DIFF
--- a/src/handlers/major_change.rs
+++ b/src/handlers/major_change.rs
@@ -189,7 +189,9 @@ async fn handle(
         procedural comments, such as volunteering to review, indicating that you \
         second the proposal (or third, etc), or raising a concern that you would \
         like to be addressed. \
+        \n\n{} \
         \n\n[stream]: {}",
+            config.open_extra_text.as_deref().unwrap_or_default(),
             topic_url
         );
         issue
--- a/src/config.rs
+++ b/src/config.rs
@@ -132,6 +132,7 @@ pub(crate) struct MajorChangeConfig {
     pub(crate) second_label: String,
     pub(crate) meeting_label: String,
     pub(crate) zulip_stream: u64,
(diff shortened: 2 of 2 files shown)


src/handlers/major_change.rs
```diff
@@ -189,7 +189,9 @@ async fn handle(
         procedural comments, such as volunteering to review, indicating that you \
         second the proposal (or third, etc), or raising a concern that you would \
         like to be addressed. \
+        \n\n{} \
         \n\n[stream]: {}",
+            config.open_extra_text.as_ref().map(|s| &**s).unwrap_or_else(|| ""),
```
Mark-Simulacrum, line 194: I think this can be `config.open_extra_text.as_deref().unwrap_or_default()`.

Mark-Simulacrum (commented):
Looks very reasonable to me. I'm not sure if serde/toml will handle the Option in the config right, we might need a serde(default) on it.

LeSeulArtichaut:
`Option<String>` is already used directly, e.g. in https://github.com/rust-lang/triagebot/blob/e56c12fa8e31f3b0197755a0d91663bde186bb80/src/config.rs#L67

I remember problems with `Option` and `serde` but IIRC it was only for serialization, not deserialization.

LeSeulArtichaut:
@Mark-Simulacrum Do you want me to add `#[serde(default)]` on the `Option`?

Mark-Simulacrum:
No, sorry, just didn't see this PR. I need to figure out a query that'll give me PRs here too... I guess maybe I can add to the triagebot query stuff :)

PR #924, opened 2020-10-21
Add github token check

apiraino:
I use a script to retrieve the token from a keychain, I have fumbled with the command and assigned a wrong value to the github API token. The GitHub queries then failed but the error returned by the cli tool was confusing, took me a bit to understand where the problem was.

This check ensures that the Github token retrieved is syntactically correct (a-z0-9).

cc: @spastorino

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -957,20 +957,28 @@ impl RequestSend for RequestBuilder {
     }
 }
 
+fn validate_token(t: &str) -> bool {
+    t.chars().all(|char| char.is_digit(16))
+}
+
 /// Finds the token in the user's environment, panicking if no suitable token
 /// can be found.
 pub fn default_token_from_env() -> String {
-    match std::env::var("GITHUB_API_TOKEN") {
-        Ok(v) => return v,
-        Err(_) => (),
+    let token = match std::env::var("GITHUB_API_TOKEN") {
+        Ok(v) => v,
+        Err(_) => match get_token_from_git_config() {
+            Ok(v) => v,
+            Err(_) => {
+                panic!("Could not find token in GITHUB_API_TOKEN or .gitconfig/github.oath-token")
+            }
+        },
+    };
+
+    if !validate_token(&token) {
+        panic!("Invalid Github token found")
     }
 
-    match get_token_from_git_config() {
-        Ok(v) => return v,
-        Err(_) => (),
-    }
-
-    panic!("could not find token in GITHUB_API_TOKEN or .gitconfig/github.oath-token")
+    token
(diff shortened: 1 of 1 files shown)


spastorino:
I'm not sure if I'd "validate" the token in that way but :+1:, I meant, whatever :)

cc @Mark-Simulacrum

src/github.rs
```diff
@@ -957,20 +957,28 @@ impl RequestSend for RequestBuilder {
...
-    match std::env::var("GITHUB_API_TOKEN") {
-        Ok(v) => return v,
-        Err(_) => (),
-    }
+    let token = match std::env::var("GITHUB_API_TOKEN") {
+        Ok(v) => v,
+        Err(_) => match get_token_from_git_config() {
+            Ok(v) => v,
+            Err(_) => {
+                panic!("Could not find token in GITHUB_API_TOKEN or .gitconfig/github.oath-token")
```
bjorn3, line 972:
```suggestion
                panic!("Could not find token in GITHUB_API_TOKEN or .gitconfig/github.oauth-token")
```

apiraino:
> I'm not sure if I'd "validate" the token in that way but +1, I meant, whatever :)

right, only "syntactically correct".

What would you otherwise suggest to validate the token (without performing an actual API call)?

Anyway, it's just a papercut so it's ok to simply close it :+1:  (I just thought to point it out)

src/github.rs
```diff
@@ -957,20 +957,28 @@ impl RequestSend for RequestBuilder {
     }
 }
 
+fn validate_token(t: &str) -> bool {
+    t.chars().all(|char| char.is_digit(16))
```
Mark-Simulacrum, line 961: Do we have docs for this?

Mark-Simulacrum (approved):

src/github.rs
apiraino: mmm ... I'm afraid I'm missing what you mean :-)

Do you mean we should link the documentation on [how to create this token](https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow) or that this assumption about the token format can fail at any time (and thus this patch should just be closed)?

Mark-Simulacrum: I don't really see a reason we need to validate the token to make use of it -- ultimately, the only real validation is GitHub :)

But I would prefer not to validate it ourselves if we don't have evidence the tokens are always hexadecimal.

apiraino: Fair enough. The main argument for this patch in invalid, so I'll close it as such.

Thanks for the review

PR #949, opened 2020-10-30
Fix glacier command

camelid:
It was fetching the HTML rather than the raw source;
see rust-lang/glacier#530 for an example.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -1039,6 +1039,16 @@ impl GithubClient {
         }
     }
 
+    pub async fn raw_gist_from_url(
+        &self,
+        html_url: &str,
+        filename: &str,
+    ) -> anyhow::Result<String> {
+        let url = html_url.replace("github.com", "githubusercontent.com") + "/raw/" + filename;
+        let response = self.raw().get(&url).send().await?;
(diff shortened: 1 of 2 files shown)


kellda:
cc rust-lang/glacier#231

camelid:
@Mark-Simulacrum Could you review this? Thanks!

camelid:
Thanks! Looking forward to using the command now :)

camelid:
Hmm, it doesn't seem to be working: https://github.com/rust-lang/rust/issues/78632#issuecomment-731843345

kellda:
Why did you sent the command twice ?

camelid:
Because the first time it didn't work – you need to include `/rust-play/` in the Gist URL, which I had not done. That should probably be fixed too.

@Mark-Simulacrum did you see anything in the logs that would explain why rustbot silently failed?

kellda:
It probably failed because it wasn't able to recreate a branch with the same name. (The branch having been created by your first command)

Mark-Simulacrum:
I don't have time right now to track down logs unfortunately, but @kellda's theory looks like the right one.

PR #1004, opened 2020-11-22
Support multiple different bot names

### LINKED ISSUE #985: Reply with an error message if the wrong username is pinged
Triagebot uses @rustbot on GitHub and `@triagebot` on Zulip, and it could happen that someone pings `@triagebot` on GitHub by mistake. If triagebot detects a `@triagebot` ping on GitHub with a valid command, an error message should be posted to alert the user they pinged the wrong bot.

Mark-Simulacrum:
Currently the triagebot name on GitHub isn't owned by the Rust org, but seems to
be largely unused too. However, we're calling the tool triagebot, and users can
easily get it wrong by calling the tool triagebot on GitHub rather than rustbot
(the "real" name).

For now, just accept both names interchangeably. We do not issue a warning for
the time being.

Fixes #985.

### DIFF
--- a/parser/src/command.rs
+++ b/parser/src/command.rs
@@ -12,7 +12,7 @@ pub mod relabel;
 pub mod second;
 
 pub fn find_command_start(input: &str, bot: &str) -> Option<usize> {
-    input.find(&format!("@{}", bot))
+    input.to_ascii_lowercase().find(&format!("@{}", bot))
 }
 
 #[derive(Debug, PartialEq)]
@@ -32,7 +32,9 @@ pub struct Input<'a> {
     all: &'a str,
     parsed: usize,
     code: ColorCodeBlocks,
-    bot: &'a str,
+
+    // A list of possible bot names.
+    bot: Vec<&'a str>,
 }
 
 fn parse_single_command<'a, T, F, M>(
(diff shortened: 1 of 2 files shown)


parser/src/command.rs
```diff
@@ -140,7 +147,7 @@ impl<'a> Input<'a> {
         self.parsed += if c.is_ok() {
             tok.position()
         } else {
-            self.bot.len() + 1
+            name_length
```
kellda, line 150: @Mark-Simulacrum 
When the command is successfully parsed this correctly advances the input, but when it isn't, the input is now advanced by the number of names this bot has. See line 170:
https://github.com/rust-lang/triagebot/blob/f346181e014954f66c5c495a753522e0adc343fa/parser/src/command.rs#L166-L170

Mark-Simulacrum: Happy to receive a PR, I wasn't sure on the math myself tbh.

kellda: Is there any reason to skip the whole bot name, not just the initial `@` ? Doing so would be easier and should still prevent the bot from matching again the same command.

Mark-Simulacrum: hm I don't think so, no

PR #1041, opened 2020-12-10
glacier: Append random sequence to branch name to make it unique

camelid:
Sometimes the user (me) will make a mistake (e.g. leave out
`/rust-play/` from the URL) when invoking `@rustbot glacier`
that means the automatically-created PR has to be closed.

However, there was no way to try again, because the branch name
would conflict and triagebot would crash. Now, we append an
8-character random alphanumeric string to the branch name so that
the branch names are unique and you can try again.

This adds a dependency on `rand`, which is probably the most downloaded
crate all-time, but we already depended on it through other crates.

r? @Mark-Simulacrum

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1919,6 +1919,7 @@ dependencies = [
  "openssl",
  "parser",
  "postgres-native-tls",
+ "rand",
  "regex",
  "reqwest",
  "rust_team_data",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ native-tls = "0.2"
 serde_path_to_error = "0.1.2"
 octocrab = "0.5"
 comrak = "0.8.2"
+rand = "0.7"
 
 [dependencies.serde]
 version = "1"
--- a/src/handlers/glacier.rs
+++ b/src/handlers/glacier.rs
@@ -5,6 +5,7 @@ use crate::{config::GlacierConfig, github::Event, handlers::Context};
 use octocrab::models::Object;
 use octocrab::params::repos::Reference;
 use parser::command::glacier::GlacierCommand;
+use rand::Rng as _;
 
 pub(super) async fn handle_command(
     ctx: &Context,
@@ -46,17 +47,22 @@ pub(super) async fn handle_command(
         unreachable!()
     };
 
-    fork.create_ref(
-        &Reference::Branch(format!("triagebot-ice-{}", number)),
-        master,
-    )
-    .await?;
+    let pr_branch_name = format!(
+        "triagebot-ice-{}-{}",
+        number,
+        rand::thread_rng()
+            .sample_iter(&rand::distributions::Alphanumeric)
+            .take(8)
+            .collect::<String>()
+    );
+    fork.create_ref(&Reference::Branch(pr_branch_name.clone()), master)
+        .await?;
     fork.create_file(
         format!("ices/{}.rs", number),
         format!("Add ICE reproduction for issue rust-lang/rust#{}.", number),
         body,
     )
-    .branch(format!("triagebot-ice-{}", number))
+    .branch(pr_branch_name.clone())
     .send()
     .await?;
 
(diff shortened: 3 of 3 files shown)


Mark-Simulacrum:
I'd prefer we use the API to delete the old branch or something like that instead.

camelid:
Ideally that's what we'd do, but it's not necessarily easy to figure out when the branch should be deleted. We'd need to listen to the PR events and only delete it when the PR is closed or merged. I think it's better to just go with something simple for now so that it's not a pain if you make a typo when invoking the command.

Mark-Simulacrum:
I mean that we'd always delete it or "force push" or whatever to the branch? Like, I'd want repeated invocations to update the open PR ideally

camelid:
Hmm... that seems like it might be confusing. I also think it would require significant changes to the current code, and I don't know octocrab etc. well enough to make those changes.

kellda:
What about creating the file and the pull request as currently done if the branch was created successfully (i.e. it didn't exist), and just updating the file if the branch already exists? See https://docs.rs/octocrab/0.12.0/octocrab/repos/struct.RepoHandler.html#method.update_file

camelid:
> What about creating the file and the pull request as currently done if the branch was created successfully (i.e. it didn't exist), and just updating the file if the branch already exists? See https://docs.rs/octocrab/0.12.0/octocrab/repos/struct.RepoHandler.html#method.update_file

That could work, but probably triagebot should only do that if there's no PR open for that branch (i.e., the PR was closed). Otherwise there could be confusion if two people do `@rustbot glacier` with different MCVEs around the same time.

kellda:
> Otherwise there could be confusion if two people do @rustbot glacier with different MCVEs around the same time.

Didn't think of that, but what is the probability that it will happend ? Especially since only a limited number of people are allowed to issue the command ?

Which is better ? Having to close the PR to fix it or avoiding that potential confusion ?

What about duplicates PR for the same ICE ?

kellda:
P.S. ok, definitely need to ignore commands in quoted text. I'll try to implement that

camelid:
> > Otherwise there could be confusion if two people do @rustbot glacier with different MCVEs around the same time.
> 
> Didn't think of that, but what is the probability that it will happend ? Especially since only a limited number of people are allowed to issue the command ?
> 
> Which is better ? Having to close the PR to fix it or avoiding that potential confusion ?
> 
> What about duplicates PR for the same ICE ?

Personally, I think it would be best to start out by only updating the branch if there's no open PR, and then we can always lift that restriction in the future if it causes problems.

kellda:
As you see fit. The question then is what to do when the a PR is open ? Emit a warning / error ? Or open a concurrent PR ?

I still think it's unlikely that two persons issue the command with different MCVEs but for same ICE / on the same issue. Anyway…

camelid:
> As you see fit. The question then is what to do when the a PR is open ? Emit a warning / error ? Or open a concurrent PR ?
> 
> I still think it's unlikely that two persons issue the command with different MCVEs but for same ICE / on the same issue. Anyway…

It's probably not a big deal. I guess I'm just thinking of if someone tried to spam the open PR, but it's probably unlikely.

So maybe updating the branch even if a PR is open is fine. Maybe @Mark-Simulacrum has thoughts on what makes sense here?

Mark-Simulacrum:
Spam is not something we're trying to solve too much.

I think either updating the branch in place or generating a new branch and updating that will both be fine. I think generally leaning towards one PR per invocation is probably better; PRs are relatively cheap and if need be could also be edited in place by reviewers.

Mark-Simulacrum:
It might be a little cleaner to associate the branch with a single (user, issue) pair and update that in place to provide an easy way to amend PRs, perhaps. But either way seems ok.

camelid:
> It might be a little cleaner to associate the branch with a single (user, issue) pair and update that in place to provide an easy way to amend PRs, perhaps. But either way seems ok.

That sounds like a good idea. E.g., if I used it on issue 123, the branch would be named `triagebot-ice-camelid-123` or something like that. Then we wouldn't need to check if it's the same user because the branch name would do it for us.

Mark-Simulacrum:
Right. We'd still need the create if not exists & update logic, but hopefully that's not too hard to pull off.

Urgau:
Thanks for the contribution! Since then [rust-lang/glacier](https://github.com/rust-lang/glacier) has been archived.

Closing.

PR #1078, opened 2020-12-31
Notify Zulip when issue is closed or reopened

camelid:
This adds optional `message_on_close` and `message_on_reopen` fields to
`triagebot.toml` to notify the Zulip topic when the issue is closed or
reopened.

Inspired by [this comment] by Ryan Levick.

[this comment]: https://rust-lang.zulipchat.com/#narrow/stream/245100-t-compiler.2Fwg-prioritization.2Falerts/topic/.2380375.20ICE.3A.20compiler.2Frustc_middle.2Fsrc.2Fty.2Fsubst.2Ers.3A568.3A17.3A.20c.E2.80.A6/near/221205328

### DIFF
--- a/src/handlers/notify_zulip.rs
+++ b/src/handlers/notify_zulip.rs
@@ -11,6 +11,8 @@ pub(super) struct NotifyZulipInput {
 pub(super) enum NotificationType {
     Labeled,
     Unlabeled,
+    Closed,
+    Reopened,
 }
 
 pub(super) fn parse_input(
@@ -18,40 +20,54 @@ pub(super) fn parse_input(
     event: &IssuesEvent,
     config: Option<&NotifyZulipConfig>,
 ) -> Result<Option<NotifyZulipInput>, String> {
-    if let IssuesAction::Labeled | IssuesAction::Unlabeled = event.action {
-        let applied_label = &event.label.as_ref().expect("label").name;
-        if let Some(config) = config.and_then(|c| c.labels.get(applied_label)) {
-            for label in &config.required_labels {
-                let pattern = match glob::Pattern::new(label) {
-                    Ok(pattern) => pattern,
-                    Err(err) => {
-                        log::error!("Invalid glob pattern: {}", err);
-                        continue;
-                    }
-                };
-                if !event
-                    .issue
-                    .labels()
-                    .iter()
-                    .any(|l| pattern.matches(&l.name))
-                {
-                    // Issue misses a required label, ignore this event
-                    return Ok(None);
-                }
-            }
+    let applied_label = &event.label.as_ref().expect("label").name;
 
-            if event.action == IssuesAction::Labeled && config.message_on_add.is_some() {
-                return Ok(Some(NotifyZulipInput {
-                    notification_type: NotificationType::Labeled,
-                }));
-            } else if config.message_on_remove.is_some() {
-                return Ok(Some(NotifyZulipInput {
-                    notification_type: NotificationType::Unlabeled,
-                }));
+    if let Some(config) = config.and_then(|c| c.labels.get(applied_label)) {
+        for label in &config.required_labels {
+            let pattern = match glob::Pattern::new(label) {
+                Ok(pattern) => pattern,
+                Err(err) => {
+                    log::error!("Invalid glob pattern: {}", err);
+                    continue;
+                }
+            };
+            if !event
+                .issue
+                .labels()
+                .iter()
+                .any(|l| pattern.matches(&l.name))
+            {
+                // Issue misses a required label, ignore this event
+                return Ok(None);
             }
         }
+
+        match event.action {
+            IssuesAction::Labeled if config.message_on_add.is_some() => {
+                Ok(Some(NotifyZulipInput {
+                    notification_type: NotificationType::Labeled,
+                }))
+            }
+            IssuesAction::Unlabeled if config.message_on_remove.is_some() => {
+                Ok(Some(NotifyZulipInput {
+                    notification_type: NotificationType::Unlabeled,
+                }))
+            }
+            IssuesAction::Closed if config.message_on_close.is_some() => {
+                Ok(Some(NotifyZulipInput {
+                    notification_type: NotificationType::Closed,
+                }))
+            }
+            IssuesAction::Reopened if config.message_on_reopen.is_some() => {
+                Ok(Some(NotifyZulipInput {
+                    notification_type: NotificationType::Reopened,
+                }))
+            }
+            _ => Ok(None),
+        }
+    } else {
+        Ok(None)
     }
-    Ok(None)
 }
 
 pub(super) async fn handle_input<'a>(
@@ -78,6 +94,8 @@ pub(super) async fn handle_input<'a>(
     let mut msg = match input.notification_type {
         NotificationType::Labeled => config.message_on_add.as_ref().unwrap().clone(),
         NotificationType::Unlabeled => config.message_on_remove.as_ref().unwrap().clone(),
+        NotificationType::Closed => config.message_on_close.as_ref().unwrap().clone(),
+        NotificationType::Reopened => config.message_on_reopen.as_ref().unwrap().clone(),
     };
 
     msg = msg.replace("{number}", &event.issue.number.to_string());
--- a/src/config.rs
+++ b/src/config.rs
@@ -123,6 +123,8 @@ pub(crate) struct NotifyZulipLabelConfig {
     pub(crate) topic: String,
     pub(crate) message_on_add: Option<String>,
     pub(crate) message_on_remove: Option<String>,
+    pub(crate) message_on_close: Option<String>,
+    pub(crate) message_on_reopen: Option<String>,
     #[serde(default)]
     pub(crate) required_labels: Vec<String>,
 }


src/handlers/notify_zulip.rs
```diff
@@ -11,47 +11,63 @@ pub(super) struct NotifyZulipInput {
...
-                };
-                if !event
-                    .issue
-                    .labels()
-                    .iter()
-                    .any(|l| pattern.matches(&l.name))
-                {
-                    // Issue misses a required label, ignore this event
-                    return Ok(None);
+    let applied_label = &event.label.as_ref().expect("label").name;
```
LeSeulArtichaut, line 23: I don't think this will work for close and reopen events?

camelid: Hmm, I think you're right :/

I think I'll need to make this function take a reference to the old state of the issue.

LeSeulArtichaut:
@camelid what's the status of this?

camelid:
I need to fix the bug you found in https://github.com/rust-lang/triagebot/pull/1078#discussion_r550455721 :)

I'll try to get to this sometime soon.

camelid:
I've returned to try working on this again, and I realized what I got stuck on:
If there are multiple labels with a `message_on_close`, we need to return
multiple `NotifyZulipInput`s, one for each of those labels. So I tried changing
the return type of `parse_input` from `Result<Option<NotifyZulipInput>, String>`
to `Result<Vec<NotifyZulipInput>, String>`. But that gives these two errors:

```
error[E0308]: mismatched types
   --> src/handlers.rs:105:20
    |
103 |               match $name::parse_input(ctx, event, config.$name.as_ref()) {
    |                     ----------------------------------------------------- this expression has type `std::result::Result<Vec<NotifyZulipInput>, std::string::String>`
104 |                   Err(err) => errors.push(HandlerError::Message(err)),
105 |                   Ok(Some(input)) => {
    |                      ^^^^^^^^^^^ expected struct `Vec`, found enum `std::option::Option`
...
127 | / issue_handlers! {
128 | |     autolabel,
129 | |     major_change,
130 | |     notify_zulip,
131 | | }
    | |_- in this macro invocation
    |
    = note: expected struct `Vec<NotifyZulipInput>`
                 found enum `std::option::Option<_>`
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0308]: mismatched types
   --> src/handlers.rs:117:20
    |
103 |               match $name::parse_input(ctx, event, config.$name.as_ref()) {
    |                     ----------------------------------------------------- this expression has type `std::result::Result<Vec<NotifyZulipInput>, std::string::String>`
...
117 |                   Ok(None) => {}
    |                      ^^^^ expected struct `Vec`, found enum `std::option::Option`
...
127 | / issue_handlers! {
128 | |     autolabel,
129 | |     major_change,
130 | |     notify_zulip,
131 | | }
    | |_- in this macro invocation
    |
    = note: expected struct `Vec<NotifyZulipInput>`
                 found enum `std::option::Option<_>`
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to 2 previous errors
```

I'm not familiar with this code, so I'm not sure what the right solution is. It
seems like solving these errors might require larger architectural changes.

This is my attempted change in case you want to play around with it to try to
find a solution for this problem:

```diff
diff --git a/src/handlers/notify_zulip.rs b/src/handlers/notify_zulip.rs
index 92f7b1f..12aa7ca 100644
--- a/src/handlers/notify_zulip.rs
+++ b/src/handlers/notify_zulip.rs
@@ -19,55 +19,63 @@ pub(super) fn parse_input(
     _ctx: &Context,
     event: &IssuesEvent,
     config: Option<&NotifyZulipConfig>,
-) -> Result<Option<NotifyZulipInput>, String> {
-    let applied_label = &event.label.as_ref().expect("label").name;
+) -> Result<Vec<NotifyZulipInput>, String> {
+    let config = match config {
+        Some(config) => config,
+        None => return Ok(vec![]),
+    };
+    //let applied_label = &event.label.as_ref().expect("label").name;
 
-    if let Some(config) = config.and_then(|c| c.labels.get(applied_label)) {
-        for label in &config.required_labels {
-            let pattern = match glob::Pattern::new(label) {
-                Ok(pattern) => pattern,
-                Err(err) => {
-                    log::error!("Invalid glob pattern: {}", err);
-                    continue;
+    Ok(event
+        .issue
+        .labels
+        .iter()
+        .filter_map(|label| config.labels.get(&label.name))
+        .flat_map(|config| {
+            for label in &config.required_labels {
+                let pattern = match glob::Pattern::new(label) {
+                    Ok(pattern) => pattern,
+                    Err(err) => {
+                        log::error!("Invalid glob pattern: {}", err);
+                        continue;
+                    }
+                };
+                if !event
+                    .issue
+                    .labels()
+                    .iter()
+                    .any(|l| pattern.matches(&l.name))
+                {
+                    // Issue misses a required label, ignore this event
+                    return None;
                 }
-            };
-            if !event
-                .issue
-                .labels()
-                .iter()
-                .any(|l| pattern.matches(&l.name))
-            {
-                // Issue misses a required label, ignore this event
-                return Ok(None);
             }
-        }
 
-        match event.action {
-            IssuesAction::Labeled if config.message_on_add.is_some() => {
-                Ok(Some(NotifyZulipInput {
-                    notification_type: NotificationType::Labeled,
-                }))
-            }
-            IssuesAction::Unlabeled if config.message_on_remove.is_some() => {
-                Ok(Some(NotifyZulipInput {
-                    notification_type: NotificationType::Unlabeled,
-                }))
-            }
-            IssuesAction::Closed if config.message_on_close.is_some() => {
-                Ok(Some(NotifyZulipInput {
-                    notification_type: NotificationType::Closed,
-                }))
-            }
-            IssuesAction::Reopened if config.message_on_reopen.is_some() => {
-                Ok(Some(NotifyZulipInput {
-                    notification_type: NotificationType::Reopened,
-                }))
+            match event.action {
+                IssuesAction::Labeled if config.message_on_add.is_some() => {
+                    Some(NotifyZulipInput {
+                        notification_type: NotificationType::Labeled,
+                    })
+                }
+                IssuesAction::Unlabeled if config.message_on_remove.is_some() => {
+                    Some(NotifyZulipInput {
+                        notification_type: NotificationType::Unlabeled,
+                    })
+                }
+                IssuesAction::Closed if config.message_on_close.is_some() => {
+                    Some(NotifyZulipInput {
+                        notification_type: NotificationType::Closed,
+                    })
+                }
+                IssuesAction::Reopened if config.message_on_reopen.is_some() => {
+                    Some(NotifyZulipInput {
+                        notification_type: NotificationType::Reopened,
+                    })
+                }
+                _ => None,
             }
-            _ => Ok(None),
-        }
-    } else {
-        Ok(None)
-    }
+        })
+        .collect())
 }
 
 pub(super) async fn handle_input<'a>(
```

camelid:
If you have ideas, let me know! I'm kind of stuck.

LeSeulArtichaut:
Does changing `Result<Vec<NotifyZulipInput>>` to `Result<Vec<Option<NotifyZulipInput>>>` (wrapping the `Vec` in an `Option`) work? It may look suboptimal but I think waiting a little bit of memory is not so bad, we're not in `rustc` here 😅

camelid:
> Does changing `Result<Vec<NotifyZulipInput>>` to `Result<Vec<Option<NotifyZulipInput>>>` (wrapping the `Vec` in an `Option`) work? It may look suboptimal but I think waiting a little bit of memory is not so bad, we're not in `rustc` here 😅

Based on your text, I assume you mean `Result<Option<Vec<NotifyZulipInput>>>` (not `Result<Vec<Option<NotifyZulipInput>>>`), and yes I did try that:

```
error[E0308]: mismatched types
   --> src/handlers.rs:107:65
    |
107 |                           $name::handle_input(ctx, config, event, input).await.unwrap_or_else(|err| errors.push(HandlerError::Other(err)));
    |                                                                   ^^^^^ expected struct `NotifyZulipInput`, found struct `Vec`
...
127 | / issue_handlers! {
128 | |     autolabel,
129 | |     major_change,
130 | |     notify_zulip,
131 | | }
    | |_- in this macro invocation
    |
    = note: expected struct `NotifyZulipInput`
               found struct `Vec<NotifyZulipInput>`
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

To solve this I might have to make ever handler return a `Vec`, which might be feasible, but I was hoping for a solution that wouldn't require changing other things :)

LeSeulArtichaut:
Did you change the signature of `handle_input` in `handlers/notify_zulip.rs`? It should take a `Vec<NotifyZulipInput>`

camelid:
Why didn't I think of that? 🤦 

Thanks :)

camelid:
Unfortunately I had to do a bunch of refactoring and there don't seem to be any tests. Though as an extra benefit, the code is hopefully clearer :)

src/handlers/notify_zulip.rs
```diff
@@ -1,96 +1,171 @@
...
-                    return Ok(None);
+) -> Result<Option<Vec<NotifyZulipInput>>, String> {
+    let config = match config {
+        Some(config) => config,
+        None => return Ok(None),
+    };
+
+    match event.action {
+        IssuesAction::Labeled | IssuesAction::Unlabeled => {
+            let applied_label = event.label.as_ref().expect("label").clone();
```
camelid, line 35: This is probably better, right?

```suggestion
            let applied_label = event.label.cloned().expect("label");
```

src/handlers/notify_zulip.rs
```diff
@@ -1,96 +1,171 @@
...
-                };
-                if !event
-                    .issue
-                    .labels()
-                    .iter()
-                    .any(|l| pattern.matches(&l.name))
-                {
-                    // Issue misses a required label, ignore this event
-                    return Ok(None);
+) -> Result<Option<Vec<NotifyZulipInput>>, String> {
```
camelid, line 27: Unfortunately there's now two ways to represent a lack of notifications: `None` and `Some(vec![])`. However, I don't think there's a way to get rid of that without changing a lot of other code.

src/handlers/notify_zulip.rs
```diff
@@ -1,96 +1,171 @@
...
+
+    match event.action {
+        IssuesAction::Labeled if config.message_on_add.is_some() => Some(NotifyZulipInput {
+            notification_type: NotificationType::Labeled,
+            label,
+        }),
+        IssuesAction::Unlabeled if config.message_on_remove.is_some() => Some(NotifyZulipInput {
+            notification_type: NotificationType::Unlabeled,
+            label,
+        }),
```
camelid, line 69: This previously looked like:

```rust
            if event.action == IssuesAction::Labeled && config.message_on_add.is_some() {
                return Ok(Some(NotifyZulipInput {
                    notification_type: NotificationType::Labeled,
                }));
            } else if config.message_on_remove.is_some() {
                return Ok(Some(NotifyZulipInput {
                    notification_type: NotificationType::Unlabeled,
                }));
            }
```

(notice the catch-all `else if`)

The new version should hopefully be clearer.

src/handlers/notify_zulip.rs
```diff
@@ -1,96 +1,171 @@
...
+    true
 }
 
 pub(super) async fn handle_input<'a>(
     ctx: &Context,
     config: &NotifyZulipConfig,
     event: &IssuesEvent,
-    input: NotifyZulipInput,
+    inputs: Vec<NotifyZulipInput>,
 ) -> anyhow::Result<()> {
```
camelid, line 136: I recommend hiding whitespace changes when reviewing this function's diff.

camelid (commented):
I had to make this refactoring because it turns out there's a fundamental difference between label/unlabel events and close/reopen events:

In the case of label changes, you either notify based on the label's triagebot.toml config, or you don't notify because there's no action specified in the config.

On the other hand, for close/reopen events, you have to notify for every label on the issue that has a `message_on_close`/`message_on_reopen` action in its config. E.g., if we decided to add an `I-nominated` Zulip notification, then an issue with `I-nominated` and `I-prioritize` would need to notify both the `I-nominated` stream *and* the `I-prioritize` stream upon close or reopen.

So, I split out two functions: one for label changes, and one for close/reopen. This seemed like the best way to separate their logic since they have unique logic. I also split out a helper function to check if the issue has all of the `required_labels` so that it isn't duplicated between the label change and close/reopen functions.

src/handlers/notify_zulip.rs
LeSeulArtichaut: I personally find it cleaner

src/handlers/notify_zulip.rs
LeSeulArtichaut: The only difference between `None` and `Some(vec![])` in this case is that `Some(vec![])` will spawn an additional `Future` that does nothing. I think we can always improve/cleanup later.

src/handlers/notify_zulip.rs
```diff
@@ -1,96 +1,171 @@
...
-    topic = topic.replace("{number}", &event.issue.number.to_string());
-    topic = topic.replace("{title}", &event.issue.title);
-    // Truncate to 60 chars (a Zulip limitation)
-    let mut chars = topic.char_indices().skip(59);
-    if let (Some((len, _)), Some(_)) = (chars.next(), chars.next()) {
-        topic.truncate(len);
-        topic.push('…');
-    }
+    for input in inputs {
+        let config = config.labels.get(&input.label.name).unwrap();
```
LeSeulArtichaut, line 138: Nit: I think this is equivalent to
```suggestion
        let config = config.labels[&input.label.name];
```

LeSeulArtichaut (approved):
After a relatively quick review this looks good to me, left two nits.

@Mark-Simulacrum will you have time to review this?

LeSeulArtichaut:
> Unfortunately I had to do a bunch of refactoring and there don't seem to be any tests. Though as an extra benefit, the code is hopefully clearer :)

[Deploying is the test](https://rust-lang.zulipchat.com/#narrow/stream/224082-t-release.2Ftriagebot/topic/Harmless.20tests.20with.20the.20triagebot/near/211780377) :D

Mark-Simulacrum:
I am happy to review, but it'll likely be this weekend - not sure if I'll have time before then. It's in my todo list :)

camelid:
@Mark-Simulacrum could you take a look at this sometime soon? Thanks!

By the way, this might help with reviewing: https://github.com/rust-lang/triagebot/pull/1078#discussion_r584208561

PR #1242, opened 2021-03-11
Fix not finding template files

giraffate:
The triage dashboard was released but it returns 5xx status code: <https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Triage.20PR.20dashboard/near/229336937>.

I guess it's because templates are not copied when building docker image, and triagebot can't find templates and panics. ~So I fixed to load templates when building, not running.~ So I fixed to edit the Docker file to copy the templates.

Also, I forgot to check that it work properly on docker at the time of https://github.com/rust-lang/triagebot/pull/1223. I checked it this time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,10 @@ FROM ubuntu:bionic AS binary
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
     ca-certificates
 
+RUN mkdir -p /opt/triagebot
+
 COPY --from=build /target/release/triagebot /usr/local/bin/
+COPY templates /opt/triagebot/templates
+WORKDIR /opt/triagebot
 ENV PORT=80
 CMD triagebot


Mark-Simulacrum:
It's probably much nicer for development to have them loaded at runtime, right? It shouldn't be a problem to edit the Dockerfile to copy the directory appropriately, so if that would also work seems like a win to me.

giraffate:
Yes, I think so. loading at runtime is helpful for development. But I didn't know an appropriate directory name (`/usr/local/templates`?) which template files are copied to, so I fixed like this. Anyway, I will prepare a fix to edit the Dockerfile. Thanks for your review!

Mark-Simulacrum:
If it works, fine - ideally we should have the same code work locally and in docker, so maybe even something like /templates - I forget if we change the working directory in docker.

giraffate:
Thanks for your review. I updated the fix.

PR #1247, opened 2021-03-14
Fix (intentional) typos on the triage dashboard

JohnTitor:
Found on https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Triage.20PR.20dashboard/near/230171965.

### DIFF
--- a/templates/triage/index.html
+++ b/templates/triage/index.html
@@ -2,7 +2,7 @@
 <html>
     <head>
         <meta charset="utf-8">
-        <title>Toiage</title>
+        <title>Triage dashboard</title>
         <style>
             * { font-family: sans-serif; }
             code { font-family: monospace; }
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
Hm, I'm feeling like we should probably just use Triage here - it doesn't seem like the typo/intentional misspelling is worth it to me, or at least, I'm not sure what purpose it serves.

JohnTitor:
I think the `<title>` should be the same as this line:
https://github.com/rust-lang/triagebot/blob/34c01cb520c0867232e5eb5da2af927ebabbe8a7/templates/triage/pulls.html#L44

So picking up _toriage_ but if you prefer just _triage_, I think that should be also changed. Probably `Triage dashboard` is the best naming here?

Mark-Simulacrum:
triage dashboard seems reasonable to me

JohnTitor:
Alright, done.

PR #1248, opened 2021-03-14
Exclude draft PRs from triage dashboard

camelid:


src/triage.rs
```diff
@@ -52,11 +52,11 @@ pub async fn pulls(
     }
 
     let mut pulls: Vec<Value> = Vec::new();
-    for base_pull in base_pulls.into_iter() {
+    for base_pull in base_pulls.into_iter().filter(|pull| !pull.draft) {
```
camelid, line 55: Instead of excluding draft PRs, it might be a better idea to just mark them in gray. WG-triage doesn't triage draft waiting-on-review and waiting-on-author PRs, though it seems we do triage draft blocked PRs. T-clippy might triage draft PRs as well.

camelid:
I'll split this into two PRs so the date format change can land.

src/triage.rs
Mark-Simulacrum: Hm, I think it would make sense to triage them after some time to close them out (basically draft is similar to waiting-on-author I think, in fact I might say that we should just treat it equivalently)...

camelid: > Hm, I think it would make sense to triage them after some time to close them out

Yeah, they probably should continue to be included.

> basically draft is similar to waiting-on-author I think, in fact I might say that we should just treat it equivalently

I wouldn't say they're equivalent, in fact I would say most of the time they are different. For many PRs, waiting-on-author means "the author needs to address reviewer comments or fix conflicts". On the other hand, draft means "this is only partially implemented".

Dylan-DPC:
i think it's good if we can see the draft PRs. They may be draft, but we still need to check them once in a while to ensure there aren't any that are open for a really long time and rotting a lot.

camelid:
Is there some kind of indicator we could add to the dashboard that would signal that a PR is draft?

src/triage.rs
JohnTitor: +1 for hiding or indicating as a draft, I'd like to have an option to hide draft (and even `S-blocked`) since we don't have to check them frequently and rust-lang/rust's triage queue is quite long. I think something like this would be a good way here (this just hides PRs but we can tweak attributes, e.g., change the color): https://github.com/JohnTitor/triagebot/tree/hide-drafts

Urgau:
@camelid are you still using triagebot dashboard? are you willing to continue this PR?

camelid:
@Urgau I completely forgot both this PR and the dashboard existed haha. I can try continuing the PR -- are you asking because you want this feature, or you just want to check the PR's status? If the latter, I can also just close it.

Urgau:
I'm just asking about the status of the PR. I'm not sure anyone is currently using the dashboard.

So unless you're planning to use it again, I think I may remove it soon-ish.

Regarding the PR, feel free to continue working on it or just close it.

camelid:
I think it would be helpful to have some kind of dashboard to help with PR triage, but it might be easier to start fresh since this one is apparently unused anyway.

I'll close this PR.

PR #1249, opened 2021-03-14
Use shorter date format for updated at

camelid:
Before:

    Wed, 27 May 2020 17:57:14 +0000

After:

    2020-05-27

(Note that I haven't tested it, but that's what it should look like.)

This PR was split out from #1248.


Mark-Simulacrum:
Looks like a to_string() might be needed or something like that?

Issue #1066, opened 2020-12-23
Triagebot checks if setting a label is allowed before checking if it exists


jyn514:
Labels are public, so there's no point to this, and it makes the errors harder to read.

*Originally posted by @rustbot in rust-lang/rust#76447 (comment), then reported at https://github.com/rust-lang/homu/issues/130 because there are too many bots and they are taking over the rust organization.*

Mark-Simulacrum:
Happy to accept a PR, otherwise it'll probably go into backlog :)

PR #1287, opened 2021-04-02
Give a better error when setting an unknown label

jyn514:
This also gives an error when other commands set labels that don't
exist; previously they were just ignored.

I have no idea how to test this.

Closes https://github.com/rust-lang/triagebot/issues/1066.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -223,7 +223,7 @@ pub struct Label {
 }
 
 impl Label {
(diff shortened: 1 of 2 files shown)


jyn514:
Oh I just realized this allows DOS attacks by posting thousands of labels to try and use up the rate limit. Before they would get caught early because you could only post as many labels as existed in the repo. Maybe I should emit the permission denied error after all, but only if you post more labels than exist?

PR #1321, opened 2021-04-23
Add support for adding labels based on files changed

jyn514:
Add support for adding labels based on files changed:

```toml
[autolabel."A-rustdoc"]
# preexisting:
trigger_labels = []
exclude_labels = []
# new:
trigger_files = [
    "src/librustdoc",
    "src/tools/rustdoc",
    "src/rustdoc-json-types",
]
```

### DIFF
--- a/src/handlers/autolabel.rs
+++ b/src/handlers/autolabel.rs
@@ -7,11 +7,55 @@ pub(super) struct AutolabelInput {
     labels: Vec<Label>,
 }
 
-pub(super) fn parse_input(
-    _ctx: &Context,
+pub(super) async fn parse_input(
+    ctx: &Context,
     event: &IssuesEvent,
     config: Option<&AutolabelConfig>,
 ) -> Result<Option<AutolabelInput>, String> {
+    if let Some(diff) = event
+        .diff_between(&ctx.github)
+        .await
+        .map_err(|e| {
+            log::error!("failed to fetch diff: {:?}", e);
+        })
+        .unwrap_or_default()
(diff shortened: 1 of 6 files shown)


jyn514:
@Mark-Simulacrum is this waiting on me? I don't know how to write tests for this.

src/handlers/autolabel.rs
```diff
@@ -12,6 +12,27 @@ pub(super) fn parse_input(
     event: &IssuesEvent,
     config: Option<&AutolabelConfig>,
 ) -> Result<Option<AutolabelInput>, String> {
+    if event.action == IssuesAction::Opened {
```
camelid, line 15: This is fine as an MVP, but it'd be really great if this ran after the PR was opened too. Often people change more files after a PR is opened, e.g., if CI fails.

jyn514: This matches what high-five does; I don't really have time right now to make it work for other events.

camelid: Can't you just remove the conditional?

camelid: > Can't you just remove the conditional?

I guess then if you tried to remove the label, rustbot would automatically read it.

Mark-Simulacrum, line 15: We probably also want to do this on synchronize(?) so that we finally fix the bugs with files pushed after first opening not being detected.

I suspect it's potentially as simple as piping through that issue action, so maybe worth doing?

PR #1345, opened 2021-05-11
Add multi-repo query support

sladyn98:


src/actions.rs
```diff
@@ -57,77 +57,88 @@ impl<'a> Action for Step<'a> {
...
-                    }
+        /*
+         let mut actions = Vec::new();
+        actions.push(Query {
+            repos,
+            queries
+        });
+        */
+
+        for Query { repos, queries} in &self.actions {
```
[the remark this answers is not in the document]
nikomatsakis: Nope, I think this change is roughly what I had in mind, @sladyn98! I expected us to iterate over each repo and then, for each repo, over each of the queries within.

src/actions.rs
```diff
@@ -57,77 +57,88 @@ impl<'a> Action for Step<'a> {
...
-                            }
-                        }
-                    }
+        /*
+         let mut actions = Vec::new();
+        actions.push(Query {
+            repos,
+            queries
+        });
+        */
```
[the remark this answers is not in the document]
nikomatsakis: I'm not sure what this commented code is meant to do :)

nikomatsakis (changes requested):
This looks good! I think I would expect to just fix the code where we write `repo: "rust-lang/rust"` or whatever to say `repos: vec!["rust-lang/rust"]`

src/actions.rs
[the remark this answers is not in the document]
spastorino: What you did in this PR is correct, forget about the actions thing I've said because it's a bit confusing :).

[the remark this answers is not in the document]
spastorino: Just remove this commented out code and squash the commits and I guess this can be merged.

spastorino:
The problem with the taken approach is that the name of the query is what's used in the template as the var that contains the output of the query. See https://github.com/rust-lang/triagebot/pull/1345/files?w=1#diff-ad9375f96c9eec1389952c2f1a9c2191b647d25a31079a38ea3b410ee1f6dfd4R146, now we are executing N queries and overwriting the result.

Maybe one way to fix it is to use `format!("{}.{}", repo, name)` as the variable name. If we follow that approach we would need to adjust all the templates to use that format. Note that at the time we insert in the result repo is not around anymore ... you would need to construct that in the result returned from the iteration.

src/actions.rs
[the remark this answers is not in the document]
sladyn98: Perfect :+1:

sladyn98:
Thanks @spastorino for the feedback, it makes sense and in order to format the template we would need to modify the results hasmap. Just a question would we iterate over the `&self.actions` again at the end and then insert it inside after we have the results already populated ? something along these lines 

```

  for Query { repos, queries} in &self.actions {
            for repo in repos.iter() {
                /// insert inside the hasmap 
                };
}

 for (name, issues) in &results {
            context.insert(*name, issues);
        }
```

spastorino:
I think all you need to do is change here https://github.com/rust-lang/triagebot/pull/1345/files?w=1#diff-ad9375f96c9eec1389952c2f1a9c2191b647d25a31079a38ea3b410ee1f6dfd4L100, fix the compile errors and adjust the templates.

spastorino:
If the link doesn't go to the right place, I'm talking about `.entry(*name)`, change that and use `format!("{}.{}", repo, name)`

src/actions.rs
```diff
@@ -57,77 +57,88 @@ impl<'a> Action for Step<'a> {
...
-                                    .for_each(|cause| eprintln!("because: {}", cause));
-                                std::process::exit(1);
+                        }
+
+                        github::QueryKind::Count => {
+                            let count = repository.get_issues_count(&gh, &query).await;
+
+                            match count {
+                                Ok(count) => {
+                                    context.insert(*name, &count);
```
nikomatsakis, line 129:
```suggestion
                                    *context.entry(*name).or_insert(0) += count;
```

I think this is wrong, right?

nikomatsakis (changes requested):

src/actions.rs
spastorino: yep :). The name there is the name of the query you've fired, which also matches the name of the variable used on the template.
Let's use an example, let's say that you have "p_critical" as name there and get back the number of P-critical issues. When you use 2 repos, let's say "rust-lang/rust" and "rust-lang/compiler_team" if you sum those two you would end with just one variable name that you're supposed to use in 2 templates and the number you would be getting back is the sum of the P-critical issues on both repos. What you want is to contact the name of the repo and the name of the query. In this case "rust-lang/rust.p_critical" and "rust-lang/compiler_team.p_critical" and adjust the templates so you would have the number of p_critical per repo on the right templates. We may also want to shorten those names to just "rust.p_critical" and "compiler_team.critical" for convenience.

sladyn98: Got it :)

nikomatsakis: Wait. I'm not sure. I would have *expected* the sum across all repos, I think. If I didn't want that, I would use distinct queries. I certainly expect all the issues to be grouped together into one big list (I'm not sure it's useful to separate per repo).

src/actions.rs
```diff
@@ -57,82 +57,85 @@ impl<'a> Action for Step<'a> {
...
+                                                .assignees
+                                                .iter()
+                                                .map(|u| u.login.as_ref())
+                                                .collect::<Vec<_>>()
+                                                .join(", "),
+                                        })
+                                        .collect();
+
+                                    results
+                                        .entry(format!("{}.{}", repo, name))
```
nikomatsakis, line 102: same here -- I expect this to just be `name`

src/actions.rs
spastorino: If that's what we want, yeah, agreed, Niko would be right here. So, if we want to aggregate things, let's just use name, do what Niko suggested with count queries and with List queries we want to concat the resulting lists.

nikomatsakis: Great. :)

src/actions.rs
```diff
@@ -57,82 +57,85 @@ impl<'a> Action for Step<'a> {
...
+                                                .assignees
+                                                .iter()
+                                                .map(|u| u.login.as_ref())
+                                                .collect::<Vec<_>>()
+                                                .join(", "),
+                                        })
+                                        .collect();
+
+                                    results
+                                        .entry(name)
```
spastorino, line 102: Why was `*` removed?

src/actions.rs
```diff
@@ -57,82 +57,85 @@ impl<'a> Action for Step<'a> {
...
-                    }
-                };
+                    };
+                }
             }
         }
 
         for (name, issues) in &results {
-            context.insert(*name, issues);
+            context.insert(format!("{}.{}", repo, name), issues);
```
spastorino, line 138: Sorry if we caused you a lot of confusion here but we don't want the `{}.{}` trick :). Just leave the code here as it was before.

spastorino:
I think after the requested changes this should be ready to merge.

nikomatsakis:
@sladyn98 you should convert this from a draft PR I think

nikomatsakis:
Reminder:

> I think I would expect to just fix the code where we write `repo: "rust-lang/rust"` or whatever to say `repos: vec!["rust-lang/rust"]`

src/actions.rs
```diff
@@ -57,77 +57,80 @@ impl<'a> Action for Step<'a> {
...
-                                    .for_each(|cause| eprintln!("because: {}", cause));
-                                std::process::exit(1);
-                            }
-                        }
-                    }
+        for Query { repos, queries} in &self.actions {
 
-                    github::QueryKind::Count => {
-                        let count = repository.get_issues_count(&gh, &query).await;
+            for repo in repos.iter() {
```
spastorino, line 62:
```suggestion
            for repo in &repos {
```

spastorino:
Left a last nit but also, I don't think the code compiles. Have you tried compiling it?. The idea of what you need to implement is what you've done. Then it needs to compile and pass tests.

src/actions.rs
```diff
@@ -57,77 +57,80 @@ impl<'a> Action for Step<'a> {
...
-                                    .for_each(|cause| eprintln!("because: {}", cause));
-                                std::process::exit(1);
+                        }
+
+                        github::QueryKind::Count => {
+                            let count = repository.get_issues_count(&gh, &query).await;
+
+                            match count {
+                                Ok(count) => {
+                                    *context.entry(*name).or_insert(0) += count;
```
spastorino, line 121: In particular, note that the idea behind this line that we've shared with you is right but `entry` is not a method of `Context`. We would need to manually implement this behavior. Check https://docs.rs/tera/1.9.0/tera/struct.Context.html for more info.

sladyn98:
> Left a last nit but also, I don't think the code compiles. Have you tried compiling it?. The idea of what you need to implement is what you've done. Then it needs to compile and pass tests.

Yeah makes complete sense would I necessarily need these ? or can I just `cargo build` and try it out ? 

> The GITHUB_WEBHOOK_SECRET, GITHUB_API_TOKEN and DATABASE_URL environment variables need to be set.
If GITHUB_API_TOKEN is not set, the token can also be stored in ~/.gitconfig in the github.oauth-token setting.

src/actions.rs
sladyn98: Yeah, I got that error while compiling but did not get the time to dig into it, now that I see `Context` does not have an entry method so esentially I would need to see if *name is present in context if not then enter a zero else increment the count. 
I am still getting used to rust, so I am assuming I write a  different function that takes in the context and then manipulates the context object and increment the count accordingly right : )

apiraino:
> Yeah makes complete sense would I necessarily need these ? or can I just `cargo build` and try it out ?

@sladyn98 to test the prioritization agenda build you only need `GITHUB_API_TOKEN`, example:
`GITHUB_API_TOKEN=xxx cargo run --bin prioritization-agenda`

and then check the output, if it's consistent with the version generated from the master branch :-)

(if you run the tool too often you might incur in Github rate limiting your requests)

src/actions.rs
spastorino: Right, unsure if a new function will be needed though. I guess you could write that in a couple of lines of code there.

src/actions.rs
```diff
@@ -57,77 +57,80 @@ impl<'a> Action for Step<'a> {
...
-                                    .for_each(|cause| eprintln!("because: {}", cause));
-                                std::process::exit(1);
-                            }
-                        }
-                    }
+        for Query { repos, queries} in &self.actions {
 
-                    github::QueryKind::Count => {
-                        let count = repository.get_issues_count(&gh, &query).await;
+            for repo in &repos {
```
spastorino, line 62:
```suggestion
            for repo in repos {
```
`repos` is already a reference

src/actions.rs
spastorino: You would need to do something like this ...

```rust
                                    let value = if let Some(value) = context.get(*name) {
                                        value.as_u64().unwrap() + count as u64
                                    } else {
                                        count as u64
                                    };

                                    context.insert(*name, &value);
```

Not sure if that will work, please test it out :).

src/actions.rs
apiraino: I think this API (`.entry`) does not exist, probably a typo

sladyn98:
Yeah I ran it succesfullly and it compiled 
```
### "Check for union field accesses in THIR unsafeck" rust#85263

**Link:** https://github.com/rust-lang/rust/pull/85263



Process finished with exit code 0
```

apiraino:
As a test I have compared the output of the master branch and this patch for these commands:
`cargo run --bin prioritization-agenda`
`cargo run --bin lang-agenda`

and they are identical, so I guess it's LTGM :-)

apiraino (approved):

spastorino:
Looks good to me to :+1:, @apiraino great that what we already have works. We would also need to test using a vector of 2 or more repos and check that this is aggregating the results. Other than that is ready to merge but I don't have merge rights on this repo.

nikomatsakis:
as a test, we can update the lang-agenda -- there are some points where we currently distinguished "nominated issues on the rust-lang/rust repo" from "nominated issues on the reference", I'd rather have them in one big list

spastorino (approved):

nikomatsakis:
@sladyn98 I can merge once you modify the lang-team agenda

nikomatsakis (approved):

PR #1360, opened 2021-05-17
Add a filter to the triage dashboard

giraffate:
This is a filter for Author/Assignee/Labels in the triage dashboard. It enables us to search like GitHub issue/PR, for example `assignee:giraffate`, `assignee:giraffate label:S-waiting-on-review`. I think this is useful in the rust-lang/rust repository because it has many PRs.

screenshot

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,7 +21,14 @@ async fn serve_req(req: Request<Body>, ctx: Arc<Context>) -> Result<Response<Bod
             let params = matcher.params();
             let owner = params.find("owner");
             let repo = params.find("repo");
-            return triagebot::triage::pulls(ctx, owner.unwrap(), repo.unwrap()).await;
+            let filter = if let Some(q) = req.uri.query() {
+                let pair = url::form_urlencoded::parse(q.as_bytes()).find(|(k, _)| k == "q");
+                pair.map(|(_, filter)| filter.to_string())
+            } else {
+                None
+            };
+
+            return triagebot::triage::pulls(ctx, owner.unwrap(), repo.unwrap(), filter).await;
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
I'm a little worried about allowing arbitrary queries through while using our token... I think there's probably not too much that can go wrong, but I'd prefer to hold off on doing this for now. I think I'd feel more comfortable if we shifted the actual github requests to the client, either sending the resulting json back to Rust for rendering, or moving all of the code to the client (perhaps as wasm or something). Alternatively we could move the authentication to use the user's token through some kind of github app flow...

I'm mostly worried that it already seems like this page can be readily used to make fairly arbitrary and extensive requests with our token; I'm not sure how much we should be extending that. Another option is to support more direct filtering; for example each column could have a text box associated to filter it. Then the filtering can be done either on the client-side or on the server, but either way, not something we need to worry as much about as arbitrary queries...

giraffate:
Yes, I understood your concern. Hmm, I'll try to consider the other way to do it safely. Anyway, thanks for your comment!

bjorn3:
Can you add negative filtering? Eg on github `-label:A-mir` would filter out all PR's with the label `A-mir`.

giraffate:
I updated this to filter PRs on the client side.

> Can you add negative filtering?

Yes!

giraffate:
@Mark-Simulacrum Could you take another look on this? If you still worry about anything, feel free to close this.

Mark-Simulacrum:
Sorry, slipped off my radar. Thanks for the reminder. This looks good to me now!

giraffate:
Thanks!

PR #1380, opened 2021-06-06
Automatically add beta nominated to issues fixing regressions

### LINKED ISSUE #1184: Auto-nominate PRs for beta backport
We should nominate PRs for backport if any of the following are true:
* Description mentions a issue which is labeled as regression-from-stable-to-beta
* Description mentions a issue which is milestoned with the beta version number

mibac138:
... from stable to beta

Fixes #1184


src/handlers/beta_backport.rs
```diff
@@ -0,0 +1,52 @@
...
+        return Ok(());
+    }
+
+    if issue_event.issue.pull_request.is_none() {
+        return Ok(());
+    }
+
+    for caps in CLOSES_ISSUE.captures_iter(&issue_event.issue.body) {
+        // Should never fail due to the regex
+        let issue_id = caps.get(1).unwrap().as_str().parse::<u64>().unwrap();
```
Mark-Simulacrum, line 31: I would prefer to handle failure to parse here, likely by early returning, rather than with a digit count in the regex.

src/handlers/beta_backport.rs
```diff
@@ -0,0 +1,52 @@
...
+    // See https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue
+    // Max 19 digits long to prevent u64 overflow
+    static ref CLOSES_ISSUE: Regex = Regex::new("(close[sd]|fix(e[sd])?|resolve[sd]) #(\\d{1,19})").unwrap();
+}
+
+pub(crate) async fn handle(
+    ctx: &Context,
+    event: &Event,
+) -> anyhow::Result<()> {
+    let issue_event = if let Event::Issue(event) = event {
```
Mark-Simulacrum, line 15: This feature likely wants to be implemented with the more typical handler (like major change, autolabel) so that we can enable it only on particular repositories. I think the configuration would likely benefit from being a list - we likely want to support more than one label as the condition and more than one as the resulting addition on the PR.

mibac138:
I've rebased and reworked the code. How does it look now?

Mark-Simulacrum:
This looks good -- sorry I missed your message last year(!). It looks like it needs a rebase, and then we can merge.

src/handlers/beta_backport.rs
```diff
@@ -0,0 +1,85 @@
+use crate::config::BetaBackportConfig;
+use crate::github::{IssuesAction, IssuesEvent, Label};
+use crate::handlers::Context;
+use regex::Regex;
+
+lazy_static! {
+    // See https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue
+    static ref CLOSES_ISSUE: Regex = Regex::new("(close[sd]|fix(e[sd])?|resolve[sd]) #(\\d+)").unwrap();
+}
+
```
apiraino, line 10: Is this going to trigger a beta-backport nomination for *any* PR that is fixing an issue, no matter if the originating issue is a regression and the channel (nightly, beta or stable) where the regression happened?

(sorry if I missing some context)

apiraino:
@Mark-Simulacrum I've left a new [comment](https://github.com/rust-lang/triagebot/pull/1380#discussion_r1904301890).

I'd like to have some mechanism to auto-nominate PRs for backport. Together with #1380 would be so cool for us T-compiler :-)

@mibac138 do you think you will have capacity to push the work further? I know it's awkward asking this after such long time. In case the work here is stale, I can try owning this patch and bring it to the finish line.

src/handlers/beta_backport.rs
apiraino: cc @mibac138

Issue #1339, opened 2021-05-04
`@rustbot ready` alias to switch to waiting on review?


joshtriplett:
I would love to see simpler invocations that we can recommend to PR authors, that handle switching the various "waiting on" labels.

```
@rustbot ready
@rustbot waiting-on-author (or just "author")
```

Also, could we define sets of mutually exclusive labels, such as the various "waiting on" labels, so that you only have to add one and the bot will automatically remove others?

jyn514:
> Also, could we define sets of mutually exclusive labels, such as the various "waiting on" labels, so that you only have to add one and the bot will automatically remove others?

I really want `S-waiting-on-help` for situations where *technically* it's waiting-on-author but realistically that won't happen without outside input.

Mark-Simulacrum:
I at least only regularly look at waiting-on-review; any other state is likely to get entirely lost. I think it's reasonable to ask the reviewer to take point at either reassigning or identifying and pursuing that help.

jyn514:
Duplicate of https://github.com/rust-lang/triagebot/issues/553.

nikomatsakis:
I think this is a great idea. I find maintaining these labels pretty tedious. I'd be happy to help mentor someone here. I'm going to leave some notes.

nikomatsakis:
I would also like the bot to pay attention to GH reviews. Here is what I would suggest:

* when the assigned reviewer leaves a GH review, if it is "changes requested", we should change the state of "waiting on author" (and leave a comment to that effect)
* add a `@rustbot ready` command to change the review to `S-waiting-on-reviewer` and probably add a comment pinging the assigned reviewer

The question is how easy/hard this will be to do =)

nikomatsakis:
I'm thinking about the first part first -- the GH review with "changes requested". My *guess* is that this trigger a notification?

If so, I guess we want to edit this function:

https://github.com/rust-lang/triagebot/blob/67f108c7d5e6d808d16f5c29b2f4b9649af72e9d/src/handlers/notification.rs#L16

Llandy3d:
After a first analysis, if we would want to support simpler single word abbreviations, from the example above `ready` & `author` for switching labels, we could create a `shortcut.rs` handler to keep it open for more additions down the line, what do you think ?

I also wonder if it would be better for the GH review section to be its own issue since the two are different in scope :thinking:

nikomatsakis:
@Llandy3d that makes sense to me, yes

[...]

PR #1381, opened 2021-06-07
implement the shortcut handler

Llandy3d:
#1339 

This implements the shorcut handler and already works with two shorcuts:
  - `ready` -> `S-waiting-on-review`
  - `author` -> `S-waiting-on-author`
  
It correctly remove the unnecessary label while keeping previous ones. This functionality is accompanied by unit tests.

I do wonder if the `author` command should be allowed to only team members or it's ok as it is allowed now.

A follow up to this that require more analysis, would be to ping assigned reviewers.

### DIFF
--- a/parser/src/command/shortcut.rs
+++ b/parser/src/command/shortcut.rs
@@ -10,9 +10,10 @@
 
 use crate::error::Error;
 use crate::token::{Token, Tokenizer};
+use std::collections::HashMap;
 use std::fmt;
 
-#[derive(PartialEq, Eq, Debug)]
+#[derive(PartialEq, Eq, Debug, Copy, Clone)]
 pub enum ShortcutCommand {
     Ready,
     Author,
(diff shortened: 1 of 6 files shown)


Llandy3d:
Now it will also ping the appropriate people by leaving a comment:

- `ready`  -> Assignees
- `author` -> User that opened the pull request

parser/src/command/shortcut.rs
```diff
@@ -0,0 +1,95 @@
...
+        match self {
+            ParseError::ExpectedEnd => write!(f, "expected end of command"),
+        }
+    }
+}
+
+impl ShortcutCommand {
+    pub fn parse<'a>(input: &mut Tokenizer<'a>) -> Result<Option<Self>, Error<'a>> {
+        let mut toks = input.clone();
+        if let Some(Token::Word("ready")) = toks.peek_token()? {
```
nikomatsakis, line 39: I'm wondering if we can remove some of this duplicated code. Maybe something like

```rust
let shortcuts = [
    ("ready", ShortcutCommand::Ready),
    ("author", ShortcutCommand::Author),
];

for (word, command) in shortcuts {
    if let Some(Token::Word(word)) = toks.peek_token()? { ... }
}

Ok(None)
```

nikomatsakis (approved):

nikomatsakis:
@Llandy3d ok, I'm ready to merge, but it needs a rebase

Llandy3d:
@nikomatsakis rebased & resolved conflicts

nikomatsakis (commented):
Question @Llandy3d -- do we actually *check* the config at any point? In other words, shortcuts are just always enabled, right? Are the "label editing" commands always enabled? Maybe that checking already happens in `set_labels`?

Llandy3d:
> Question @Llandy3d -- do we actually _check_ the config at any point? In other words, shortcuts are just always enabled, right? Are the "label editing" commands always enabled? Maybe that checking already happens in `set_labels`?

Features are not always enabled by default, the check happens in `handlers.rs`, it seems to happen in the `command_handlers` macro. 

To maintain the current behaviour it's mandatory to configure the shortcuts, if you don't and try to use them you will be greeted by the standard error message:
image

So the `[shortcut]` config in `triagebot.toml` is required

nikomatsakis (approved):

nikomatsakis:
Got it! Going to merge now, then @Llandy3d can you update the wiki?

Llandy3d:
@nikomatsakis sure for the wiki, somehow when running tests from the root of the repo it wasn't running parser's tests, fixed a case where it would return a `ParseError` instead of `Ok(None)`

Llandy3d:
The corresponding wiki page can be found [here](https://github.com/rust-lang/triagebot/wiki/Shortcuts)

PR #1382, opened 2021-06-07
Add disallowed labels to notify_zulip config

### LINKED ISSUE #714: Add `{include|exclude}_labels` options to `notify-zulip`
For example, auto-labeling `I-prioritize` is useful in most cases but we don't want to notify of all that on Zulip.
So it'd be great if `notify-zulip` config had `{include|exclude}_labels` options.

cc https://github.com/rust-lang/rust/issues/74531

mibac138:
Fixes #714 

I believe the proposed `include_labels` is already added as `required_labels`, correct me if I'm wrong though.

### DIFF
--- a/src/handlers/notify_zulip.rs
+++ b/src/handlers/notify_zulip.rs
@@ -1,6 +1,6 @@
 use crate::{
     config::NotifyZulipConfig,
-    github::{IssuesAction, IssuesEvent},
+    github::{IssuesAction, IssuesEvent, Label},
     handlers::Context,
 };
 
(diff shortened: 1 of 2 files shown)


nikomatsakis:
This looks good to me. The only question would be whether it should be "any" (i.e., disallowed it any of the labels match, versus all). But that seems to match the existing option.

mibac138:
I'm not sure what exactly prompted issue #714 but I feel like this version should be enough for simple cases, and like you said, it matches the existing option. If need be, it can always be expanded

src/handlers/notify_zulip.rs
```diff
@@ -21,23 +21,30 @@ pub(super) fn parse_input(
...
-                    .any(|l| pattern.matches(&l.name))
-                {
-                    // Issue misses a required label, ignore this event
-                    return Ok(None);
                 }
+                false
+            };
+
+            if !contains_label(event.issue.labels(), &config.required_labels) {
+                // Issue misses a required label, ignore this event
```
Mark-Simulacrum, line 41: This comment sounds a little misleading - I think the behavior is that if the issue is missing *all* of the required labels, then we will ignore this event. If it's just missing one of the patterns, then we will not early return here. Right?

Mark-Simulacrum, line 41:
```suggestion
                // Ignore this event when none of the required labels are present. (Only one is enough).
```

Mark-Simulacrum:
Once conflicts are resolved here I'd be happy to merge.

Urgau (commented):
Impl looks good, but this PR needs to be rebased. Once that's done we can merge it.

PR #740, opened 2020-07-31
Add lang agenda structure

nikomatsakis:
This adds a basic lang-team agenda structure so that we can start pre-generating our agendas vs me clicking on a bunch of links. 

It also adds a little code to enable getting the github API token from `.gitconfig` instead of the environment (this is used as a fallback).

Thanks to @spastorino for doing the initial legwork!

r? @Mark-Simulacrum

PR #1384, opened 2021-06-09
Fix wrong endpoint when searching in pull requests

apiraino:
The Github API are a bit confusing when the intent is to search
either through issues *or* pull requests. Their documentation mentions
the endpoint `/search/issues`, I'm not sure if `/search/pulls`
introduced in #740 was an API since then removed.
https://docs.github.com/en/rest/reference/search#search-issues-and-pull-requests

By the way, under the hood Github only has "issues" as entities and
pull requests are just "issues with a special flag".

Commit 106bbb22 unearthed this bug.

This patch should fix when searching through pull requests in the
rust-lang/rfcs repository, labels `T-compiler` and `I-nominated` were
ignored, corresponding to the `[RFC]` section of the WG-Prioritization agenda

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,14 @@ To configure the GitHub webhook, point it to the `/github-hook` path of your
 webserver (by default `http://localhost:8000`), configure the secret you chose
 in `.env`, set the content type to `application/json` and select all events.
 
+## Notes about Github issues/pulls request APIs
+
+The Github API are a bit confusing when the intent is to search either through issues _or_ pull requests. The `/{repo}/pulls` endpoint does not support filter by label, while the `/{repo}/issues` does.
+
+The search endpoint and issues/pulls in some cases can resolve the exact same kind of queries. The `/search` and `/issues` endpoints both return issues and pull requests, the endpoint `/pulls` only returns pull requests. The reason is that under the hood Github only has a "issue" entity and pull requests are just "issues with a special flag" (see [their documentation](https://docs.github.com/en/rest/reference/issues#list-repository-issues)).
+
+In some cases the Triagebot CLI ends up hitting the Github API limit of your auth token so splitting some queries into different endpoints might help. Also, the `/search` endpoint is much slower.
+
 ## License
 
 Triagebot is distributed under the terms of both the MIT license and the
--- a/src/agenda.rs
+++ b/src/agenda.rs
@@ -490,12 +490,12 @@ pub fn prioritization<'a>() -> Box<dyn Action> {
 
     let mut queries = Vec::new();
 
-    //https://github.com/rust-lang/rfcs/pulls?q=is%3Aopen+is%3Apr+label%3AI-nominated+label%3AT-compiler
+    // https://github.com/rust-lang/rfcs/pulls?q=is%3Aopen+label%3AI-nominated+label%3AT-compiler
     queries.push(QueryMap {
         name: "nominated_rfcs_t_compiler",
         query: github::Query {
             kind: github::QueryKind::List,
-            filters: vec![("state", "open"), ("is", "pr")],
+            filters: vec![("state", "open")],
             include_labels: vec!["T-compiler", "I-nominated"],
             exclude_labels: vec![],
         },
--- a/src/github.rs
+++ b/src/github.rs
@@ -767,9 +767,9 @@ impl Repository {
             .any(|&(key, value)| key == "is" && value == "pr");
         // negating filters can only be handled by the search api
         let url = if is_pr {
-            self.build_pulls_url(filters, include_labels)
+            self.build_issues_pulls_url(filters, include_labels)
         } else if use_issues {
-            self.build_issues_url(filters, include_labels)
+            self.build_issues_pulls_url(filters, include_labels)
         } else {
             self.build_search_issues_url(filters, include_labels, exclude_labels)
         };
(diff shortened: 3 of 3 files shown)


apiraino:
r? @Mark-Simulacrum 

cc: @spastorino @nikomatsakis

spastorino:
To be honest I'm not sure about those endpoints and how this should be done, would need to properly investigate but what I can say is if we merge this PR, let's better remove `build_pulls_url` and just use `build_issues_url` since from what I see the endpoint is the only difference.

spastorino:
@apiraino note that the link you're providing is about the search API and the code you're changing uses this https://docs.github.com/en/rest/reference/pulls API.

spastorino:
In any case, whatever we do those 2 methods are identical and we should not repeat the code as we are doing now.

apiraino:
thanks @spastorino for having a look. Following your hint I've checked a bit more and now I think their [/{repo}/pulls](https://docs.github.com/en/rest/reference/pulls#list-pull-requests) API we are using does not support filtering by the `label` parameter. 

On the other hand the [/{repo}/issues](https://docs.github.com/en/rest/reference/issues#list-repository-issues) does (case in point for this patch).

Good point on merging the two functions into one (they're now identical; sorry, could have paid more attention).

src/github.rs
```diff
@@ -767,9 +767,9 @@ impl Repository {
             .any(|&(key, value)| key == "is" && value == "pr");
         // negating filters can only be handled by the search api
         let url = if is_pr {
-            self.build_pulls_url(filters, include_labels)
+            self.build_issues_pulls_url(filters, include_labels)
         } else if use_issues {
-            self.build_issues_url(filters, include_labels)
+            self.build_issues_pulls_url(filters, include_labels)
```
apiraino, line 772: I'd like to remove this redundant if/else conditional in a follow-up patch after properly testing this PR

spastorino:
I've started to remember a bit of why these things existed :).

First of all, the `search` endpoint and `issues`/`pulls` in some cases can resolve the exact same kind of queries. `search` and `issues` if I'm not wrong returns both issues and prs and `pulls` only prs. If I'm not wrong in some cases it ended helping by splitting some queries into different endpoints to avoid hitting the github API limit. In other cases, I'm not sure if `is:pr`worked at all in those endpoints, it sounds to me that it does work for the `search` endpoint but for on `issues` and that explains the existence of using `pulls`. You can also resolve everything using `search` but if I also remember correctly was waaay more slow.

Sorry for the tons of doubts I have, all I'm trying to say is ... are you sure this can be removed and everything will stay the same?. Does this change decrements speed of resolution? does this change keep returning the correct results for every query?.

src/agenda.rs
```diff
@@ -490,12 +490,12 @@ pub fn prioritization<'a>() -> Box<dyn Action> {
 
     let mut queries = Vec::new();
 
-    //https://github.com/rust-lang/rfcs/pulls?q=is%3Aopen+is%3Apr+label%3AI-nominated+label%3AT-compiler
+    // https://github.com/rust-lang/rfcs/pulls?q=is%3Aopen+label%3AI-nominated+label%3AT-compiler
     queries.push(QueryMap {
         name: "nominated_rfcs_t_compiler",
         query: github::Query {
             kind: github::QueryKind::List,
-            filters: vec![("state", "open"), ("is", "pr")],
```
apiraino, line 498: I think this filtering label was not needed in the first place

apiraino:
so @spastorino I've tried gathering some comparing data about what actually changes (since we don't have tests I can only this tool live). Tests have been run on Thu. June, 24th 2021 before the weekly T-compiler meeting.

Summary: the only difference I can see in the nominated RFCs section (which was my goal for this PR).

* Here are the wg-prioritization agenda **before** and **after** the changes. The only diff I see is for the nominated RFCs, which now looks correct and doesn't output those spurious results
[agenda-after.txt](https://github.com/rust-lang/triagebot/files/6710343/agenda-after.txt)
[agenda-before.txt](https://github.com/rust-lang/triagebot/files/6710344/agenda-before.txt)
* I've printed the output of the API queries emitted. The only difference I see is the query for nominated RFCs
[queries-after.txt](https://github.com/rust-lang/triagebot/files/6710364/queries-after.txt)
[queries-before.txt](https://github.com/rust-lang/triagebot/files/6710365/queries-before.txt)
* I've timed the Triagebot cli run on both commits and the difference is negligible (like 2 or 3s)

Please feel free to verify the above data and point out anything I may have missed. Thanks :-)

apiraino:
can you pls also review https://github.com/rust-lang/triagebot/pull/1384/commits/3e2ec4c5820926705909d2d3d7110d685b6f3451 is it's consistent with your thoughts you wrote in this [comment](https://github.com/rust-lang/triagebot/pull/1384#issuecomment-857933274)? We need to start documenting this tool, I think.

nikomatsakis (approved):

spastorino:
I wonder what happens if we apply this PR, to this https://github.com/rust-lang/triagebot/blob/3e2ec4c5820926705909d2d3d7110d685b6f3451/src/agenda.rs#L535 maybe @nikomatsakis have some thought about it. The only problem I could see to this PR is if you just want pull requests and not issues. Seems like for some reason we don't need that anymore in the wg-prioritization team but in that line seems like lang needs it. Would that case work correctly?.

README.md
```diff
@@ -23,6 +23,14 @@ To configure the GitHub webhook, point it to the `/github-hook` path of your
 webserver (by default `http://localhost:8000`), configure the secret you chose
 in `.env`, set the content type to `application/json` and select all events.
 
+## Notes about Github issues/pulls request APIs
+
+The Github API are a bit confusing when the intent is to search either through issues _or_ pull requests. The `/{repo}/pulls` endpoint does not support filter by label, while the `/{repo}/issues` does.
```
spastorino, line 28: Are you sure pulls doesn't support labels?, quickly looking at our code it seems like we are using labels for PRs queries.

apiraino:
ok so apparently this patch is confusing and also doing too many things. I thought the purpose was more straightforward (my bad).

So I'm going to close it and perhaps at a later point resubmit some of these changes splitting them into smaller chunks.

thanks @spastorino for the discussion and suggestions (I'll address the comments in a separate PRs)

PR #1405, opened 2021-06-24
Update T-libs labels in agenda

apiraino:
Companion PR for https://github.com/rust-lang/rust-forge/pull/546

The team now uses `T-libs` and `T-libs-api`, so this patch renames all occurrences of `T-libs-impl` to `T-libs` and will catch both new labels

r? @spastorino


spastorino:
I'd just change the `T-libs` labels in this PR and leave the work about `build_*_url` to be addressed here https://github.com/rust-lang/triagebot/pull/1384

apiraino:
apologies @spastorino I messed up the commits from various branches. Now I forced push and should be ok

PR #1406, opened 2021-06-24
Update Tokio runtime

apiraino:
I think we should be brave and update Tokio, the Triagebot is still running v0.2 :-)

This patch only updates:
- Tokio crate + small changes to update the code for Tokio 1.x
- reqwest crate
- postgres crates (for the Triagebot DB connection)

### Notes on testing

I could test the build of the T-compiler and T-lang agenda, but I don't know how to test the Triagebot server

### Misc

Other crates need to be upgraded  but I think a separate PR is better suited (esp. because of `octocrab`+`hyper`)

I think r? @Mark-Simulacrum might be interested in this PR? Also cc: @spastorino

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ parser = { path = "parser" }
 rust_team_data = { git = "https://github.com/rust-lang/team" }
 glob = "0.3.0"
 toml = "0.5.1"
-hyper = "0.13"
+hyper = { version = "0.14.4", features = ["server", "stream"]}
 tokio = { version = "1.7.1", features = ["macros", "time", "rt"] }
 futures = { version = "0.3", default-features = false, features = ["std"] }
 async-trait = "0.1.31"
@@ -33,7 +33,7 @@ tokio-postgres = { version = "0.7.2", features = ["with-chrono-0_4"] }
 postgres-native-tls = "0.5.0"
 native-tls = "0.2"
 serde_path_to_error = "0.1.2"
-octocrab = "0.5"
+octocrab = "0.9.1"
 comrak = "0.8.2"
 route-recognizer = "0.3.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ edition = "2018"
 serde_json = "1"
 openssl = "0.10"
 dotenv = "0.15"
-reqwest = { version = "0.10", features = ["json", "blocking"] }
+reqwest = { version = "0.11.4", features = ["json", "blocking"] }
 regex = "1"
 lazy_static = "1"
 log = "0.4"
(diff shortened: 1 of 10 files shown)


apiraino:
hm ... tests work on my machine (:tm:) maybe try re-running the CI?

nikomatsakis:
Error:

```
error[E0658]: `match` is not allowed in a `const fn`
   --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/socket2-0.4.0/src/lib.rs:156:9
    |
156 | /         match address {
157 | |             SocketAddr::V4(_) => Domain::IPV4,
158 | |             SocketAddr::V6(_) => Domain::IPV6,
159 | |         }
    | |_________^
    |
    = note: see issue #49146 <https://github.com/rust-lang/rust/issues/49146> for more information
```

That doesn't look like something a re-run will fix -- maybe there are newer versions available?

JohnTitor:
We use the pinned Rust version on Docker, i.e. 1.45: https://github.com/rust-lang/triagebot/blob/327d5835c1c7ddb346d5bad5aaccf00107da17ab/Dockerfile#L8
`const_if_match` has been stabilized since 1.46 so we should update the Docker image to it or higher.

Cargo.toml
```diff
@@ -22,15 +22,15 @@ rust_team_data = { git = "https://github.com/rust-lang/team" }
 glob = "0.3.0"
 toml = "0.5.1"
 hyper = "0.13"
```
JohnTitor, line 24: I'd also update `hyper` and `octocrab` to the latest so that we can drop tokio 0.2.x completely.

apiraino:
> We use the pinned Rust version on Docker, i.e. 1.45:
> `const_if_match` has been stabilized since 1.46 so we should update the Docker image to it or higher.

ah thanks, good point. Done in https://github.com/rust-lang/triagebot/commit/9add45f2ed4fba3bcab6748e359e984e76da3727

Cargo.toml
apiraino: @JohnTitor the reason I was suggesting to hold back updating these two crates is that they have both have a conflict in a common nested dependency (the `httparse` crate). [There's a PR](https://github.com/dekellum/hyperx/pull/32) that could potentially fix this for us, so I'd like to wait and see.

Otherwise I'm not sure how can I pin the right version for `httparse` so that `octocrab` and `hyper` can get along happily (any suggestion?)

apiraino: for context
```
error: failed to select a version for `httparse`.
    ... required by package `hyperx v1.3.0`
    ... which is depended on by `octocrab v0.9.1`
    ... which is depended on by `triagebot v0.1.0 (../rust/triagebot)`
versions that meet the requirements `>=1.0, <1.4` are: 1.3.6, 1.3.5, 1.3.4, 1.3.3, 1.3.2, 1.3.1, 1.3.0, 1.2.5, 1.2.4, 1.2.3, 1.2.2, 1.2.1, 1.2.0, 1.1.2, 1.1.1, 1.1.0, 1.0.0

all possible versions conflict with previously selected packages.

  previously selected package `httparse v1.4.0`
    ... which is depended on by `hyper v0.14.9`
    ... which is depended on by `hyper-tls v0.5.0`
    ... which is depended on by `reqwest v0.11.4`
    ... which is depended on by `octocrab v0.9.1`
    ... which is depended on by `triagebot v0.1.0 (../rust/triagebot)`

failed to select a version for `httparse` which could resolve this conflict
```

apiraino: uhm ... maybe I can work around the issue by carefully pinning `hyper = "0.14.5" ` (... checking...)

JohnTitor: We can avoid it by using hyper 0.14.5 or lower, I mean, we don't have to use the latest patch version to drop tokio 0.2, a newer minor version is enough. You can specify it as follows:

```toml
hyper = { version = "=0.14.5", features = ["server"] } # Note: "server" feature is required on 0.14.x.
```

I haven't tested deeply but it should resolve the dependency issue.

spastorino:
:+1: on the idea but I guess @Mark-Simulacrum or some other maintainer that knows exactly what and how to test are better suited to merge this one.

PR #1466, opened 2021-08-28
Revamp oldest unreviewed PRs

jackh726:
Added in infrastructure to run graphql queries with cynic. After I had basically done the query side of this, I realized that it's kind of a lot of code. Might be easier to just change to `graphql-client` at some point. But for now this works.

Some further refactoring here could be nice. I didn't realize it at the time, but looking back through it, `GithubQuery` could really just have a `Box<dyn IssuesQuery>` and then a REST query just impls `IssuesQuery`.


spastorino:
@jackh726 yeah, agree that this can be improved later. Going to merge this ... cc @apiraino to be aware and properly test the results for next `T-compiler` weekly meeting.

src/github/graphql.rs
```diff
@@ -0,0 +1,347 @@
+use anyhow::Context;
+use async_trait::async_trait;
+
+#[cynic::schema_for_derives(file = "src/github/github.graphql", module = "schema")]
```
apiraino, line 4: @jackh726 I'd add a bit of documentation [about the schema file](https://docs.github.com/en/graphql/overview/public-schema), especially about [updating it and maintenance](https://docs.github.com/en/graphql/overview/breaking-changes).

(I'm not familiar with GraphQL, so I don't know what maintenance implies in terms of context needed)

src/github/graphql.rs
```diff
@@ -0,0 +1,347 @@
...
+                                )
+                            })
+                            .collect::<Vec<_>>()
+                    })
+                    .unwrap_or_default();
+                let mut comments: Vec<_> = comments
+                    .into_iter()
+                    .filter(|comment| assignees.contains(&comment.0))
+                    .collect();
+                comments.sort_by_key(|c| c.1);
```
apiraino, line 304: @jackh726 I've noticed that the last comment now reported is from an assignee. Before it was the absolute latest comment (and not always correct because it included also bots).

In case [of this issue](https://github.com/rust-lang/rust/pull/84944#issuecomment-906810332) the result is a bit confusing: the latest comment reported in the agenda is "4 months ago" but Wesley added a comment about a week ago.

How should we interpret now the latest comment? "Latest comment from assignee" or "latest comment from a human"?

cc: @spastorino

jackh726: @apiraino yeah, I think "latest comment from assignee" makes the most sense; will require a bit of changes to the template and such

We should maybe quickly note (maybe discuss) in the meeting that updates like Wesley's should probably be a "review" or should change the status to either blocked or waiting-on-author.

spastorino: Maybe comments from assignees + members of the org? :P

jackh726: @spastorino that wouldn't really work given that bors and highfive are members, as well as triage.

spastorino: Well yeah you could filter those out.

jackh726: I think it's easier/more straightforward to stick with last comment by assignee. If the goal of this is to move the PR to a resolution (merge or close), then I think *reviews* are the better metric.

Sure, some newer PRs may start with a discussion through comments, but as PRs age, that discussion should resolve, or be a blocker. At the limit of "very old" for PRs, what kind of comments are we expecting? #84944 is an interesting case; it's an old issue that *was* waiting for review, but when it was discussed during the meeting, was decided it's now blocked on a test. However, when that test is added, it's probably *still* "very old"; this comment doesn't necessarily "refresh" that.

apiraino: I can see how the "last comment by assignee" make sense in the context of PRs and issues. However the initial intent of appending the most recent comment was to check the pulse of an MCP/FCP and see if it was alive or abandoned (after ~3 months of inactivity they will be closed).

Changing now how the last comment works would perhaps require a split:
- last comment from assignee for PRs/issues
- last comment from a human for MCPs/FCPs

makes sense?

jackh726: I think we don't need to move away from last comment (by anyone) for issues or MCPs/FCPs. The simplest way to think about these is that these don't usually get assigned *anyways*.

I'm really not seeing that there is anything that needs to be changed right now, other than just a rewording to the template to make it clear that it's not the last comment, but actually "last review activity" (even "last comment by assignee" is wrong, I'm remembering, since this takes into account reviews too)

apiraino: ok, if you're happy with how it works then I'm happy, too :) 

I'll push a small patch to update the wording of the template as you suggest.

thanks for answering quickly to my doubt!

PR #1524, opened 2021-11-02
Update queries and templates for move from I-nominated to I-teamname-nominated

joshtriplett:
This updates queries and templates that referenced I-nominated.

This does not attempt to update the `nominate` command.

### DIFF
--- a/src/handlers/relabel.rs
+++ b/src/handlers/relabel.rs
@@ -185,7 +185,7 @@ mod tests {
         macro_rules! t {
             ($($member:ident { $($label:expr => $res:ident,)* })*) => {
                 let config = RelabelConfig {
-                    allow_unauthenticated: vec!["T-*".into(), "I-*".into(), "!I-nominated".into()],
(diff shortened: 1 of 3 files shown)


src/handlers/relabel.rs
```diff
@@ -185,7 +185,12 @@ mod tests {
         macro_rules! t {
             ($($member:ident { $($label:expr => $res:ident,)* })*) => {
                 let config = RelabelConfig {
-                    allow_unauthenticated: vec!["T-*".into(), "I-*".into(), "!I-nominated".into()],
+                    allow_unauthenticated: vec![
+                        "T-*".into(),
+                        "I-*".into(),
+                        "!I-*-nominated".into(),
```
Mark-Simulacrum, line 191: nit: the triagebot.toml in rust-lang/rust has `!I-*nominated`, maybe should match here.

joshtriplett: @Mark-Simulacrum Will do.

PR #1536, opened 2021-12-12
Improve the `trigger_files` implementation a bit

camelid:
Hopefully this will help with debugging a bit.


Mark-Simulacrum:
Thanks! I might integrate these though I think the root cause is failure to parse out before/after correctly.

camelid:
> Thanks! I might integrate these though I think the root cause is failure to parse out before/after correctly.

Yeah, I added these tests (which required extracting the function) to narrow down which code is under suspicion :)

Mark-Simulacrum:
Ok, directly merged in with conflicts fixed. Thanks!

PR #1562, opened 2022-01-20
feat: note command

### LINKED ISSUE #1427: add "rustbot note" command to amend OP with notes
I would like to have a `@rustbot note x` command used like:

```rust
@rustbot note summary

details details details
```

This command will add a link to the OP like

```
### Notes

* ["summary" by @username](link-to-comment)
```

This can be used to allow people to add comments or other significant notes.

chazkiker2:
### Overview

- [x] Closes #1427

### Changes

- [x] Introduces new `@rustbot note` command, allows the user to make a comment in an issue/PR that creates a Summary Entry. 

### Command Specification

The following comment would create a new summary entry

```md
@rustbot note abc    

details details details
```
    
If this was the first `@rustbot note` command to be triggered in this specific issue, it would add a new section in the Issue Body Description (or "Original Post") of the issue called "Summary Notes":
    
```md
### Summary Notes    

- ["abc" by @chazkiker2](link-to-comment)
``` 
    
If this was not the first `@rustbot note` command to be triggered in this issue, rustbot would simply append to the existing summary notes: 
    
```md
### Summary Notes
    
- ["abc" by @chazkiker2](link-to-chaz-comment)
- ["xyz" by @nikomatsakis ](link-to-niko-comment)
```


nikomatsakis:
Cool =)

Mark-Simulacrum:
Just fyi - please ping me here (and likely best on Zulip as well) when you're ready for a review or would like some help, otherwise it'll likely get lost in my notifications :)

chazkiker2:
@nikomatsakis — Here's a quick screen recording with the current functionality, please let me know if it's what you expected or what you'd like to change. More notes/questions below

https://user-images.githubusercontent.com/70249988/151492393-acbec388-552e-4686-89a1-1d547d03a471.mov

- One thing I noticed is that if I put new notes under a pre-existing `<!-- TRIAGEBOT_SUMMARY_START --> ... <!-- TRIAGEBOT_SUMMARY_END -->`. Wondering if that's expected behavior from `interactions.rs` or if that's a bug—and if so, whether I introduced it. I'll look further into it, but if that's expected functionality, please let me know 😄 
- Is the formatting of the comment markdown to your liking? Where did you want to place it in the issue body? Is "Summary Notes" an appropriate title, or would you prefer something else?

Mark-Simulacrum:
> * One thing I noticed is that if I put new notes under a pre-existing `<!-- TRIAGEBOT_START --> ... <!-- TRIAGEBOT_END -->`, triagebot starts a new section for the `EditIssueBody`. Wondering if that's expected behavior from `interactions.rs` or if that's a bug—and if so, whether I introduced that or if it was pre-existing. I'll look further into it, but if that's expected functionality, please let me know

Can you clarify what you mean by this? I'm not sure I'm following what you're saying -- maybe a link to an issue where this happened or another screencap would be useful.

I think it would be *nice* but definitely not required for the user to be able to directly edit the list, and in general feels relatively painful to get that to work well (editing markdown vs. re-generating from scratch is annoying) -- we can consider it a future improvement, but not more than that, I think.

> Would you like any restrictions on the command? Users that are capable of using it? Should the summary details meet any requirements—the simplest of which would be "to be there" in the first place?

I don't think we should impose any restrictions -- ideally, `@rustbot note <foo>`  is sufficient, and the details could be before or after that. We should also try to support multiple notes per comment -- but that might be a little annoying, not sure if the parser is well-suited to that just yet.

src/handlers/note.rs
```diff
@@ -0,0 +1,107 @@
+//! Allow users to add summary comments in Issues & Pull Requests.
+//!
+//! Users can make a new summary entry by commenting the following:
+//!
+//! ```md
+//! @rustbot note summary-title
+//!
+//! ...details details details...
```
Mark-Simulacrum, line 8: nit: let's drop this from the description of notes, since the implementation doesn't require details afterwards and we probably don't necessarily want to require that either. (If we ever start trying to capture those automatically it'll probably be by scraping the entire comment text, I suspect).

src/handlers/note.rs
```diff
@@ -0,0 +1,107 @@
...
+//! ```
+//!
+//! If this is the first summary entry, rustbot will amend the original post (the top-level comment) to add a "Notes" section:
```
Mark-Simulacrum, line 11:
```suggestion
//! If this is the first summary entry, rustbot will amend the original post (the top-level comment) to add a "Notes" section. The section should **not** be edited by hand.
```

src/handlers/note.rs
```diff
@@ -0,0 +1,107 @@
...
+    title: String,
+    comment_url: String,
+    author: String,
+}
+
+impl NoteDataEntry {
+    pub fn to_markdown(&self) -> String {
+        format!(
+            "\n- [\"{title}\" by @{author}]({comment_url})",
+            title = self.title.to_owned(),
```
Mark-Simulacrum, line 53: The `.to_owned()` here and on the other fields shouldn't be necessary.

src/handlers/note.rs
```diff
@@ -0,0 +1,107 @@
...
+//! ```
+//!
+//! If this is the first summary entry, rustbot will amend the original post (the top-level comment) to add a "Notes" section:
+//!
+//! ```md
+//! <!-- TRIAGEBOT_SUMMARY_START -->
+//!
+//! ### Summary Notes
+//!
+//! - ["summary-title" by @username](link-to-comment)
```
Mark-Simulacrum, line 18: Can we add a line below these that says something like `Generated by triagebot, see [help](wiki-page-link) for how to add more`. (And we'll want to create that wiki page https://github.com/rust-lang/triagebot/wiki -- let me know if you need help with that, not sure what permissions are like).

Mark-Simulacrum (commented):
This looks great -- left some comments on the implementation.

Mark-Simulacrum:
We may also want to add a `@rustbot note remove <foo>` command, which drops a note from the top-level description. I suspect a note called literally "remove" is unlikely, so we don't need to worry about that.

parser/src/command/note.rs
```diff
@@ -0,0 +1,39 @@
...
+        }
+    }
+}
+
+impl NoteCommand {
+    pub fn parse<'a>(input: &mut Tokenizer<'a>) -> Result<Option<Self>, Error<'a>> {
+        let mut toks = input.clone();
+        if let Some(Token::Word("note")) = toks.peek_token()? {
+            toks.next_token()?;
+            if let Some(Token::Word(title)) = toks.next_token()? {
```
Mark-Simulacrum, line 28: Let's support quoting the title as well -- i.e., Token::Quote -- which will let users supply a nicer name than just a single word.

chazkiker2:
> > * One thing I noticed is that if I put new notes under a pre-existing `<!-- TRIAGEBOT_START --> ... <!-- TRIAGEBOT_END -->`, triagebot starts a new section for the `EditIssueBody`. Wondering if that's expected behavior from `interactions.rs` or if that's a bug—and if so, whether I introduced that or if it was pre-existing. I'll look further into it, but if that's expected functionality, please let me know
> 
> Can you clarify what you mean by this? I'm not sure I'm following what you're saying -- maybe a link to an issue where this happened or another screencap would be useful.

@Mark-Simulacrum Here's a screen cast of this bug and a link to the demo issue https://github.com/chazkiker2/triagebot-playground/issues/4: 

https://user-images.githubusercontent.com/70249988/152027722-83ecad6e-b0a3-43cf-ad1b-4f9cf4a43a7c.mov

nikomatsakis:
Haven't caught up, but something that came up when I was discussing this with @joshtriplett -- it might be nice if the edits are made to a "first comment" that we add automatically to the beginning of the issue -- that would of course require adding this comment for every new issue.

chazkiker2:
> Haven't caught up, but something that came up when I was discussing this with @joshtriplett -- it might be nice if the edits are made to a "first comment" that we add automatically to the beginning of the issue -- that would of course require adding this comment for every new issue.

@nikomatsakis when you say "first comment" you mean a new comment as opposed to the issue body (or "original post")?

Mark-Simulacrum:
Hm, that seems like it's probably a bug in interactions.rs (pre-existing, almost certainly). I don't think it needs to block this PR necessarily but we should likely try to fix it. At a quick glance the code there is pretty poorly commented, so not completely sure where the bug is. It should be possible to look at just the section between the triagebot start/end comments I think, not sure how hard that would be.

re: the first comment, I'm pretty skeptical -- I don't think it's a good idea to start leaving such a comment on every issue and every PR, it seems pretty noisy. I'd start by editing the issue for now. In the future it might make sense to use the first comment where we control it (e.g., on MCPs, we already post one; on PRs we will once highfive functionality is migrated in).

chazkiker2:
As far as `@rustbot note remove XXX` functionality goes

- My latest commit supports the following functionality: 
  - `@rustbot note remove XXX` will remove the first summary entry with the title XXX from the data and the markdown
  - If there are duplicate XXX entries, only the first is removed
  - If there exists no such XXX entry, main will panic—This surely needs to be fixed... see below
- I will make updates to support: 
  - TODO: If there exists no such XXX entry instead, we will actually handle the error and make a rustbot comment noting the error
- Questions: 
  - Should we make a comment if the XXX entry doesn't exist? 
  - Should we support for more granular control over which comment is deleted? What if I wanted to delete the second XXX entry instead—we could number the list and allow for `@rustbot note remove 3` which would remove the 3rd comment?

Mark-Simulacrum:
I don't think we should expect more than one note with the same title, removing the first one seems OK for now. I would not post comments if users don't get it right, we can just silently ignore it: they should be able to edit the comment and make the deletion happen, I think.

chazkiker2:
@Mark-Simulacrum another question for ya (this is a possible bug): If a user edits the title of an existing `@rustbot note to-edit` entry and replaces the title with say `edited`, triagebot treats it like a new entry. 

Demo issue https://github.com/chazkiker2/triagebot-playground/issues/8 and screencast: 

https://user-images.githubusercontent.com/70249988/152057049-7e686543-8dbb-4de6-bdb7-6fe124072db1.mov

chazkiker2:
@Mark-Simulacrum and @nikomatsakis , I think PR is ready for another review... let me know your thoughts 😄 

- To split into separate issues: 
  - [ ] Allow user to directly edit the summary note list
  - [ ] Allow user to make multiple note entries in the same comment
  - [ ] (Unrelated bug) Editing below `<!-- TRIAGEBOT_START -->...<!-- TRIAGEBOT_END -->` makes `interactions.rs` think it has to make a whole new entry
- To do otherwise
  - [ ] Actually add documentation into [the wiki entry](https://github.com/rust-lang/triagebot/wiki/Note) for this command
- Potential edge cases:
  - [ ] What if a user _deletes_ a summary comment (either by deleting the comment itself or simply removing the `@rustbot note xxx` portion of the comment—I assume we probably want to remove that entry from the summary section?

joshtriplett:
> re: the first comment, I'm pretty skeptical -- I don't think it's a good idea to start leaving such a comment on every issue and every PR, it seems pretty noisy.

I was proposing this for the case where rustbot (or for now, rust-highfive) already posts the first comment.

Long-term, for RFCs I'm hoping that first comment includes the "rendered" link, and these notes, and links to any FCPs started, and other such things.

src/handlers/note.rs
```diff
@@ -0,0 +1,168 @@
...
+            } else {
+                let new_entry = NoteDataEntry {
+                    title: title.clone(),
+                    comment_url: comment_url.clone(),
+                    author,
+                };
+                log::debug!("New Note Entry: {:#?}", new_entry);
+                current
+                    .entries_by_url
+                    .insert(comment_url, new_entry.clone());
```
chazkiker2, line 149: can remove unnecessary clones

parser/src/command/note.rs
```diff
@@ -0,0 +1,55 @@
...
+#[derive(PartialEq, Eq, Debug)]
+pub enum NoteCommand {
+    Summary { title: String },
+    Remove { title: String },
+}
+
+#[derive(PartialEq, Eq, Debug)]
+pub enum ParseError {
+    MissingTitle,
+}
```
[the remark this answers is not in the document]
Mark-Simulacrum: I don't think that's necessary.

Mark-Simulacrum:
I'm inclined to merge this in as-is and then iterate in the future -- the changes look good to me, and I think even a partial implementation with some bugs is likely to already be quite useful. We can leave most of the nice-to-have features for future (potentially near future, but future) PRs.

@chazkiker2 Do you see any blockers to doing so?

chazkiker2:
@Mark-Simulacrum sounds good to me!

chazkiker2:
Just added an initial wiki entry for this command, @Mark-Simulacrum feel free to edit / remove / add anything you think necessary :)

chazkiker2:
> > re: the first comment, I'm pretty skeptical -- I don't think it's a good idea to start leaving such a comment on every issue and every PR, it seems pretty noisy.
> 
> I was proposing this for the case where rustbot (or for now, rust-highfive) already posts the first comment.
> 
> Long-term, for RFCs I'm hoping that first comment includes the "rendered" link, and these notes, and links to any FCPs started, and other such things.

Split to #1579

nikomatsakis:
@chazkiker2 thanks for working on this!

PR #1583, opened 2022-02-18
[fix: 1573] stop duplicating triagebot sections in EditIssueBody

### LINKED ISSUE #1573: interactions: Stop duplicating TRIAGEBOT section comments if user edits underneath initial comments
- If a user puts new notes under a pre-existing `<!-- TRIAGEBOT_START --> ... <!-- TRIAGEBOT_END -->`, triagebot starts a new section for the `EditIssueBody`. 
- It seems like this is probably a bug in `interactions.rs`.  It should be possible to look at just the section between the triagebot start/end comments I think, not sure how hard that would be.

Here's a screen cast of this bug and a link to a demo issue https://github.com/chazkiker2/triagebot-playground/issues/4: 

https://user-images.githubusercontent.com/70249988/152027722-83ecad6e-b0a3-43cf-ad1b-4f9cf4a43a7c.mov

chazkiker2:
### Description 

- [x] Fix #1573

### DIFF
--- a/src/interactions.rs
+++ b/src/interactions.rs
@@ -1,5 +1,6 @@
 use crate::github::{GithubClient, Issue};
 use std::fmt::Write;
+use tracing as log;
 
 pub struct ErrorComment<'a> {
     issue: &'a Issue,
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
This seems like a surprising fix -- I would not expect us to ever emit a carriage return (\r).

Maybe the code is not expecting 2 newlines (\n) and so this is really just changing `\n\n` -> `\n` and that is also sufficient?

I think the idea is that the two newline characters are necessary to ensure a blank line between the comment and the following text, though I don't know if that's ever *necessary*.

Mark-Simulacrum:
Hm, it may be that we need to normalize \r\n to \n, rather than adjusting the text we expect here -- the parser typically does this for us (see [this test](https://github.com/rust-lang/triagebot/blob/20158891c0eec7addd4fd1a93a584e35ab414cb6/parser/src/token.rs#L237)), but we're not using the parser here.

chazkiker2:
@Mark-Simulacrum the `\r\n` -> `\n` normalization seems to work well. Where I placed it (within `interactions.rs`) feels somewhat hackish — should this normalization happen somewhere within `github.rs` instead? Or is living within interactions appropriate as the tokenizer will handle most occurrences of this?

src/interactions.rs
```diff
@@ -56,19 +57,26 @@ pub struct EditIssueBody<'a> {
 static START_BOT: &str = "<!-- TRIAGEBOT_START -->\n\n";
 static END_BOT: &str = "<!-- TRIAGEBOT_END -->";
 
+fn normalize_body(body: &str) -> String {
+    str::replace(body, "\r\n", "\n").to_string()
```
Mark-Simulacrum, line 61: str::replace already returns a String, so the to_string() here isn't needed.

Mark-Simulacrum:
Modulo the one comment this looks good, I think keeping it to interactions.rs for now should be fine.

src/interactions.rs
chazkiker2: fixed in c500466

chazkiker2:
@Mark-Simulacrum I think this PR is ready to go in. lmk if you want any other changes :)

PR #1584, opened 2022-02-19
fix 1580: delete 'rustbot note' from summary comment

### LINKED ISSUE #1580: `@rustbot note`: Deleting the summary comment (or `rustbot note` call in comment)
A potential edge case for the note comment: what if a user deletes a summary comment (either by deleting the comment itself or simply removing the @rustbot note xxx portion of the comment—I assume we probably want to remove that entry from the summary section?

chazkiker2:
### Description

- [ ] Fix #1580

### Changes 

- [x] Allow user to delete `@rustbot note` from an existing Summary Comment — this will remove the entry from the Summary Section in the top-level comment
- [ ] Allow user to delete the Summary Comment — this will remove the entry from the Summary Section in the top-level comment
- [ ] Allow user to change the title portion of an existing Summary Comment — this will update the title in the Summary Section in the top-level comment


chazkiker2:
might need your guidance on this guy @Mark-Simulacrum. feels like the wrong place to put this functionality and yet I'm unsure where to place this.

the Note parser won't be able to parse a comment without the `rustbot note`. but we still have to trigger the note command handler when the user deleted that portion of the comment

Mark-Simulacrum:
I'm inclined to say we shouldn't really support this -- deleting comments / summary text from comments should be pretty rare, to non-existent, and we already have the rustbot remove note command which seems preferable to me.

That said, from an architecture perspective the 'right' way is probably to add another function to all the handlers for 'handle_removed' or something like that (similar to how we handle executing commands today after parsing them here), and then pipe the items filtered from the vector to there. It's not clear that many commands have a *good* reaction to deletion though, and IMO it can get pretty confusing when deleting undo-s the command (vs. editing, which is typically more expressive of the user's intent).

Mark-Simulacrum:
Not too important to support editing, but likely makes more sense. I expect note bodies to change much more than note titles.

I think we already have the infrastructure in place, at least partially, for editing to work.

chazkiker2:
yes indeed, I believe I captured that functionality in the initial PR, though it may have some bugs. can’t recall.

chazkiker2:
gonna close this given the conversation :). thanks for your input

chazkiker2:
@Mark-Simulacrum the only other version of this I meant to ask about... what if the user deletes the `rustbot note` comment entirely? At that point, would we want to remove that entry from the Summary Notes section?

Mark-Simulacrum:
No, that's pretty much equivalent from our perspective I think and is pretty rare regardless.

PR #1588, opened 2022-03-13
Rename the entire zulip topic when an MCP issue is renamed. Fixes #1228.

### LINKED ISSUE #1228: Zulip topic is only renamed for messages after MCP issue rename occurs
It happened [here](https://rust-lang.zulipchat.com/#narrow/stream/233931-t-compiler.2Fmajor-changes/topic/Lang.20Item.20for.20Transmutability.20compiler-team.23411/near/228833214). triagebot renamed the topic of the "renaming this Zulip topic" message, which effectively splits the topic in two, when it should have renamed the topic of the original "new proposal" message.

The code that is responsible for renaming the topic is here:

https://github.com/rust-lang/triagebot/blob/3272855862cf4bb2d3c1a729e335e77dd124ce42/src/handlers/major_change.rs#L138-L143

Another bug is that renaming the Zulip topic breaks [the link in the MCP issue](https://github.com/rust-lang/compiler-team/issues/411#issuecomment-790851188)—which is what #711 was originally trying to fix (see #701, the short summary is that someone manually renamed the Zulip topic, which broke the link in the MCP issue).

Fixing these issues is probably going to be difficult. I'm not sure what the right solution is.

skippy10110:
See comment [here](https://github.com/rust-lang/triagebot/issues/1228#issuecomment-1066166708). Note that I don't have a local dev instance and haven't been able to verify this change myself.

### DIFF
--- a/src/handlers/major_change.rs
+++ b/src/handlers/major_change.rs
@@ -139,7 +139,7 @@ pub(super) async fn handle_input(
             let zulip_update_req = crate::zulip::UpdateMessageApiRequest {
                 message_id: zulip_send_res.message_id,
                 topic: Some(&new_topic),
-                propagate_mode: None,
+                propagate_mode: Some("change_all"),
                 content: None,
             };
             zulip_update_req


Mark-Simulacrum:
Thanks! My guess is this is an API change upstream -- I don't recall seeing https://zulip.com/api/update-message#parameter-propagate_mode before. But this seems like the right setting.

skippy10110:
I think I may have misunderstood this bug possibly, I added an additional comment on #1228. I think the actual issue is that renaming the topic works but would end up breaking the link left on the MCP?

skippy10110:
@Mark-Simulacrum I'm not 100% sure this should be merged based on my last comment, I think my fix may have been premature. (sorry 😿)

Mark-Simulacrum:
I think this is the right fix, it may just be incomplete -- we definitely want to update the full thread on Zulip, but it may *also* be the case that the GitHub link to the Zulip topic is not currently updated. I suspect that should just require clicking through the 'stub' thread left by the renaming though; not really different to today.

skippy10110:
Ok, cool. I'm also reading through [Resolving topics breaks links by changing the topic name!](https://github.com/zulip/zulip/issues/19651) and [Provide permalinks to reference streams, topics and messages.](https://github.com/zulip/zulip/issues/19651) right now.

I wonder if just having an additional comment posted by rustbot on the github issue would be sufficient? Editing the original comment with an updated link would be nice, but would also be trickier.

Mark-Simulacrum:
An additional comment seems OK to me.

PR #1589, opened 2022-03-13
Update GitHub issue with renamed Zulip topic link

skippy10110:
Post additional comment to the GitHub issue with an updated Zulip topic link when the issue/topic is renamed. Addendum to #1228 and #1588.

### DIFF
--- a/src/handlers/major_change.rs
+++ b/src/handlers/major_change.rs
@@ -156,7 +156,7 @@ pub(super) async fn handle_input(
         &event.issue,
         zulip_msg,
         config.meeting_label.clone(),
-        cmd == Invocation::NewProposal,
+        Some(cmd),
     )
     .await
 }
@@ -204,7 +204,7 @@ pub(super) async fn handle_command(
         issue,
         zulip_msg,
         config.second_label.clone(),
-        false,
+        None,
     )
     .await
 }
@@ -215,7 +215,7 @@ async fn handle(
     issue: &Issue,
     zulip_msg: String,
     label_to_add: String,
-    new_proposal: bool,
+    cmd: Option<Invocation>,
 ) -> anyhow::Result<()> {
     let github_req = issue.add_labels(&ctx.github, vec![Label { name: label_to_add }]);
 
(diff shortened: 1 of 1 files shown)


skippy10110:
Again, please keep in mind I have no way to test this myself (and I'm also a total Rust noob...). Hopefully this change is correct. 😊

skippy10110:
cc @Mark-Simulacrum since you approved my other change

apiraino:
@skippy10110 I am totally with the rationale for this patch (it is disorienting when a user clicks a link in a github comment and lands on a "dead" Zulip  topic), but I'm slightly unsure about the implementation discussed in #1588.

Renaming a Github issue would generate comments in the issue itself that could potentially add noise. Github sometimes also compresses messages that are deemed "not relevant" to the discussion (example: some technical reports from our bot) so I wonder if these comments would end up being hidden.

Another way I could think to solve the issue of orphaned Zulip topics is: when a Github issue change name, send a message to the previous Zulip topic and forward users to the new topic ("Issue has been renamed, [new Zulip topic here](#)").
 
I'm just thinking aloud here, so maybe my suggestion is not great :) What do you think?

Mark-Simulacrum:
Doesn't Zulip already post such a message to both the old and new streams?

skippy10110:
Cool, I'll try testing it out with a dummy MCP then if you think that's ok. I already filed on the wrong repo originally (compiler-team), and then I filed incorrectly at lang-team next before I got it right, and then I ended up with extra streams from my renames.... so I was a little hesitant to make more noise. 😁

skippy10110:
Ok, I've verified that everything in the topic does get correctly moved by #1588 now, but the old topic URLs do indeed get broken.

* https://rust-lang.zulipchat.com/#narrow/stream/243200-xxx/topic/TRIAGEBOT.20TESTING.3A.20Please.20ignore.20lang-team.23148
* https://rust-lang.zulipchat.com/#narrow/stream/243200-t-lang.2Fmajor-changes/topic/TRIAGEBOT.20TESTING.3A.20Please.20ignore.20.28rename.201.29.20lang-team.23148

Accessing the old URLs doesn't provide any indication of the renamed topic:
image

In light of that, I prefer the suggestion from @apiraino, where we leave a single comment behind under the old topic name pointing to the new topic name. We could still apply the fix in this PR as well if we wanted to leave a comment with the updated link and save following the "trail of breadcrumbs", but it wouldn't be strictly necessary.

skippy10110:
I left a review comment already, but just wanted to note that this PR as-is currently has a bug that would prevent it from working. I'm planning to close this in favor of #1590 instead if that lands, so haven't bothered to fix it.

Mark-Simulacrum:
Sounds like we should close this now. Thanks!

PR #1590, opened 2022-03-14
After renaming zulip topic, post a comment under the old topic pointing to the new topic

### LINKED ISSUE #1228: Zulip topic is only renamed for messages after MCP issue rename occurs

skippy10110:
After renaming zulip topic (#1588), post an additional comment under the old (now empty) topic pointing to the newly renamed topic. Fixes #1228.

Alternative to #1589. Although #1589 could potentially still be applied. it would just be somewhat redundant.

cc @Mark-Simulacrum @apiraino

### DIFF
--- a/src/handlers/major_change.rs
+++ b/src/handlers/major_change.rs
@@ -147,6 +147,28 @@ pub(super) async fn handle_input(
                 .await
                 .context("zulip message update failed")?;
 
+            // after renaming the zulip topic, post an additional comment under the old topic with a url to the new, renamed topic
+            // this is necessary due to the lack of topic permalinks, see https://github.com/zulip/zulip/issues/15290
+            let new_topic_url = crate::zulip::Recipient::Stream {
+                id: config.zulip_stream,
+                topic: &new_topic,
+            }.url();
+            let breadcrumb_comment = format!(
+                "The associated GitHub issue has been renamed. Please see the [renamed Zulip topic]({}).",
+                new_topic_url
+            );
+            let zulip_send_breadcrumb_req = crate::zulip::MessageApiRequest {
+                recipient: crate::zulip::Recipient::Stream {
+                    id: config.zulip_stream,
+                    topic: &prev_topic,
+                },
+                content: &breadcrumb_comment,
+            };
+            zulip_send_breadcrumb_req
+                .send(&ctx.github.raw())
+                .await
+                .context("zulip post failed")?;
+
             return Ok(());
         }
     };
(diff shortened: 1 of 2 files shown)


apiraino:
that's really cool @skippy10110 , thanks!

skippy10110:
Thanks for the great suggestion!

skippy10110:
This might become fixed natively by zulip in zulip/zulip#15290, https://github.com/zulip/zulip/commit/f0c680e9c0d1a62fd414bccc82e4ac255173aaa9 !

I just tried it quickly, but I suppose I need for that commit to actually makes it way through to our zulip instance :)

Mark-Simulacrum:
Hi! @skippy10110 I'm happy to merge or not merge -- not very clear on the status here. Could you provide an update?

skippy10110:
this is all ready to go on my end if you're fine with the change (with the caveat that i have no way to test the change until it's live)

skippy10110:
i'm not sure if it's actually been reviewed by anyone, if you want to take a look also

Mark-Simulacrum (approved):

skippy10110:
@Mark-Simulacrum, do you know if this should be live yet? I just [tried it out](https://github.com/rust-lang/lang-team/issues/148) and didn't see the new comment post.

Mark-Simulacrum:
It looks like we weren't successfully deploying (unfortunately, we don't really have alarms for this currently), due to an outdated glibc version. The good news is that for once health checks worked so the previous deployment remained live; the bad news is that this patch didn't actually go out.

With https://github.com/rust-lang/triagebot/pull/1612 merged and now deployed this should also be in production now.

skippy10110:
phew 😊 thanks for letting me know, i just tested it out and it's working great!

https://rust-lang.zulipchat.com/#narrow/stream/243200-xxx/topic/TRIAGEBOT.20TESTING.3A.20Please.20ignore.20lang-team.23148

PR #1591, opened 2022-03-15
Relax parsing of labels

mbartlett21:
Let it accept more commands

### DIFF
--- a/parser/src/command.rs
+++ b/parser/src/command.rs
@@ -210,14 +210,14 @@ impl<'a> Command<'a> {
 #[test]
 fn errors_outside_command_are_fine() {
     let input =
-        "haha\" unterminated quotes @bot modify labels: +bug. Terminating after the command";
+        "haha\" unterminated quotes @bot labels +bug. Terminating after the command";
     let mut input = Input::new(input, vec!["bot"]);
     assert!(input.next().unwrap().is_ok());
 }
 
 #[test]
 fn code_1() {
-    let input = "`@bot modify labels: +bug.`";
+    let input = "`@bot modify label: +bug.`";
     let mut input = Input::new(input, vec!["bot"]);
     assert!(input.next().is_none());
 }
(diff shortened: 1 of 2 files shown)


parser/src/command.rs
```diff
@@ -242,7 +242,7 @@ fn edit_1() {
 
 #[test]
 fn edit_2() {
-    let input_old = "@bot modify label: +bug.";
+    let input_old = "@bot modify labeled: +bug.";
```
Mark-Simulacrum, line 245: Hm, I'm not entirely sure I'm following this test (before or modified), but labeled doesn't look right to me, since I don't think that parses at all right now? We probably want to write a comment here and update the test so that it has the same behavior in terms of what things parse to (and maybe assert that as well).

Mark-Simulacrum:
Sorry this slipped by my radar for so long -- I think it looks good overall, just one small comment.

parser/src/command.rs
mbartlett21: @Mark-Simulacrum It is testing an error there with the assert_ne! below.

https://github.com/rust-lang/triagebot/pull/1591/files#diff-1eb763ec24cd0509bade43f86001d94744947818b6ea7596c4e003c858327395R249

Mark-Simulacrum: There's no guarantee as to what will actually happen in that assert, though? It could just as easily be two different parses or something like that, which seems not great.

mbartlett21: Agreed. I'll take a look at rewriting a new test

mbartlett21:
@Mark-Simulacrum 
I've simplified the parsing (and allowed using `@rustbot labels to: +this -that`), and updated the tests to test various inputs.

@rustbot ready

mbartlett21:
@Mark-Simulacrum 

These are the tests that are failing now, since I allowed `@bot labels to: +this`. Should these still provide a `MisleadingTo` error?

https://github.com/rust-lang/triagebot/blob/83f7a8df8483954e8e6e59feb1a80b1f111c228c/parser/src/command/relabel.rs#L270-L292

Mark-Simulacrum:
Probably that shouldn't be an error anymore.

mbartlett21:
@Mark-Simulacrum 
Ok. This is ok to merge now.

PR #1593, opened 2022-03-17
Add cutoff env var for PRs WRF from T-compiler

apiraino:
Very small patch but very useful for me when drafting the T-compiler weekly agenda. The default cutoff value (5) for listing the PRs waiting for a review from the T-compiler is often not enough, I need to dig deeper in this [github query](https://github.com/rust-lang/rust/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-asc+label%3AS-waiting-on-review+draft%3Afalse+label%3AT-compiler+-label%3AT-lang+-label%3AT-infra+-label%3AT-release+-label%3AT-libs+-label%3AT-libs-api).

Having a dynamic cutoff value is for me very helpful. Default value (5) is as before when no override is provided.

r? @Mark-Simulacrum 

thanks!

### DIFF
--- a/src/github/graphql.rs
+++ b/src/github/graphql.rs
@@ -304,7 +304,7 @@ impl super::IssuesQuery for LeastRecentlyReviewedPullRequests {
 
         let prs: Vec<_> = prs
             .into_iter()
-            .take(5)
+            .take(50)
             .map(
                 |(updated_at, number, title, html_url, repo_name, labels, assignees)| {
                     let updated_at = crate::actions::to_human(updated_at);


nikomatsakis:
Seems fine, but couldn't we do this more elegantly-- e.g. as part of the template or something, so you don't have to set the environment variable?

Mark-Simulacrum:
Can you say more about the motivation here? I'm definitely not really happy with an arbitrary environment variable -- seems hacky -- but I'm not sure I understand yet how a dynamic value is helpful for you, and would like to know more.

apiraino:
Sure, no problem :)

@mark so the thing I am trying to solve is that there is a [GraphQL query](https://github.com/rust-lang/triagebot/blob/bc415ec002845aa0c90a28df5457b666381769f2/src/github/graphql.rs#L32-L35) pulling the first 100 PRs assigned to `T-compiler` waiting for review. Then [the first 5 elements](https://github.com/rust-lang/triagebot/blob/bc415ec002845aa0c90a28df5457b666381769f2/src/github/graphql.rs#L307) are extracted, all the rest is discarded. This is my understanding of the implementation.

Now, these PRs tends to stay in `s-waiting-for-review` for some time, not always the reviewer can immediately work on them, so the following week I would end up with the same 5 results in the `T-compiler` agenda. In order to rotate the PRs on the agenda a bit I need to skip the first results and go down the list. This is why printing a longer list helps me doing the agenda.

I could have hardcoded a `20` instead of the `5` but I'm not too fond of hardcoded values.

An alternate implementation can be to remove the `.take(5)` and try implementing the filter at the template level, like @nikomatsakis suggests (Tera, the templating engine we use, supports [for loops with indexes](https://tera.netlify.app/docs/#for)).

I didn't think of that option (so thanks @nikomatsakis!) so if it works, would that make everyone happy?

apiraino:
> It seems like we should always dump all the PRs, or perhaps randomly shuffle them and pick 5? I'm not sure yet why you want a dynamic configuration here, vs. just increasing it :)

for all puposes and intents, simply increasing `.take(5)` to `.take(50)` would be fine for me, if you prefer this way.

w.r.t. randomly picking would probably not help because (and I forgot to mention that) I need them sorted by "last activity date" (a metric calculated by looking at the latest comment from the PR reviewer/author).

Mark-Simulacrum:
I think GitHub has *some* support for last activity date, though it typically fails to reflect what I actually want :)

I would be happy to s/5/50, that seems totally fine to me. I'm not a fan of the environment variable to configure things, that's all.

apiraino:
here we go @Mark-Simulacrum I have amended this patch as per our discussion.

let me know if it's fine, thanks!

Mark-Simulacrum (approved):

PR #1596, opened 2022-03-23
[feature] Lang Agenda — Proposed FCPs: include concerns and FCP notes

### LINKED ISSUE #1592: lang agenda: include concerns and FCP notes
One of the items in the lang-agenda dumps out the proposed FCPs:

https://github.com/rust-lang/triagebot/blob/bc415ec002845aa0c90a28df5457b666381769f2/templates/lang_agenda.tt#L46-L51

It would be awesome to include registered comments and the like.

chazkiker2:
### Description

- [x] Fixes #1592 

See an example of the new lang agenda (with updated FCP details) in this [HackMD](https://hackmd.io/@chazkiker/lang-agenda-fcp-details)


src/github.rs
```diff
@@ -1046,42 +1045,94 @@ pub struct Query<'a> {
...
-                    .assignees
-                    .iter()
-                    .map(|u| u.login.as_ref())
-                    .collect::<Vec<_>>()
-                    .join(", "),
-                updated_at: crate::actions::to_human(issue.updated_at),
+            .into_iter()
+            .map(|issue| async move {
+                // TODO: this seems like a *really* bad approach, but I'm not sure how to make it better
+                let fcp_map = crate::rfcbot::get_all_fcps().await?;
```
chazkiker2, line 1072: Concern #1 — how do I get this function to only run once but still allow for it to return a `HashMap` that gets moved into an `async move`?

chazkiker2:
r? @nikomatsakis @Mark-Simulacrum 

Concerned about a couple things: 
1. See this comment: https://github.com/rust-lang/triagebot/pull/1596/files#r832791113
2. Seems like I trip GitHub's rate limit within just a handful of runs — may have just been b/c I was running locally over and over but it could also be a problem
3. Would love your guidance @Mark-Simulacrum  on the "where everything lives" aspect of this... feels like the engineering could be improved quite a bit

src/github.rs
```diff
@@ -1046,42 +1045,94 @@ pub struct Query<'a> {
...
-                    .collect::<Vec<_>>()
-                    .join(", "),
-                assignees: issue
-                    .assignees
-                    .iter()
-                    .map(|u| u.login.as_ref())
-                    .collect::<Vec<_>>()
-                    .join(", "),
-                updated_at: crate::actions::to_human(issue.updated_at),
+            .into_iter()
```
chazkiker2, line 1069: may not need `into_iter` after all, initially tried this to leave the call to `rfcbot::get_all_fcps` outside of the async move block

src/github.rs
```diff
@@ -1046,42 +1045,94 @@ pub struct Query<'a> {
...
+
+                    if let Some(fcp) = fcp_map.get(&key) {
+                        let bot_tracking_comment_html_url = format!(
+                            "{}#issuecomment-{}",
+                            issue.html_url, fcp.fcp.fk_bot_tracking_comment
+                        );
+                        let bot_tracking_comment_content = quote_reply(&fcp.status_comment.body);
+                        let fk_initiating_comment = fcp.fcp.fk_initiating_comment;
+                        let initiating_comment_html_url =
+                            format!("{}#issuecomment-{}", issue.html_url, fk_initiating_comment,);
```
chazkiker2, line 1094: could just use the `comment.html_url` since we have to call `event.get_issue` anyway 

(however, this call will likely not be necessary if we make upstream changes to rfcbot)

src/github.rs
```diff
@@ -1046,42 +1045,94 @@ pub struct Query<'a> {
...
+                        let init_comment = issue
+                            .get_comment(&client, fk_initiating_comment.try_into()?)
```
chazkiker2, line 1096: should I just make `IssueDecorators.fpc_details.fk_initiating_comment` a `usize` from the get-go?

src/github.rs
```diff
@@ -1374,6 +1425,7 @@ pub trait IssuesQuery {
     async fn query<'a>(
         &'a self,
         repo: &'a Repository,
+        include_fcp_details: bool,
```
chazkiker2, line 1428: I don't love adding this param here, to be honest... feels like this repo mostly stays away from such behavior

templates/_issues_fcps.tt
```diff
@@ -0,0 +1,19 @@
+{% import "_issue.tt" as issue %}
+
+{% macro render(issues, heading="###", empty="No issues this time.") %}
+{%- for issue in issues %}
+{{heading}} "{{issue.title}}" {{issue.repo_name}}#{{issue.number}}
+
+- **Link:** {{issue.html_url}}
+- [**Tracking Comment**]({{issue.fcp_details.bot_tracking_comment_html_url}}): {{issue.fcp_details.bot_tracking_comment_content}}
+- [**Initiating Comment**]({{issue.fcp_details.initiating_comment_html_url}}): {{issue.fcp_details.initiating_comment_content}}
```
chazkiker2, line 9: wondering what happens if `issue.fcp_details` is passed in as `None`

nikomatsakis:
> Seems like I trip GitHub's rate limit within just a handful of runs — may have just been b/c I was running locally over and over but it could also be a problem

That's not great. That said, I typically run this script once per week...

...still, I wonder if there's a way to combine this into a single graphql query or something.

src/agenda.rs
```diff
@@ -606,23 +606,17 @@ pub fn lang_planning<'a>() -> Box<dyn Action> {
...
+            queries: vec![QueryMap {
+                name: "tracking_issues",
+                kind: QueryKind::List,
+                query: Box::new(github::Query {
+                    filters: vec![("state", "open")],
+                    include_labels: vec!["C-tracking-issue"],
+                    exclude_labels: vec!["T-libs-api", "T-libs", "T-lang", "T-rustdoc"],
+                }),
+            }],
+        }],
```
nikomatsakis, line 620: this is just formatting? (do we have a mandatory rustfmt style on this repo?)

src/github.rs
nikomatsakis: can you not just move this line out of the `saync move`, like so?

```rust
let fcp_map = crate::rfcbot::get_all_fcps().await?;
let issues_decorator: Vec<_> = issues
    .into_iter()
    .map(|issue| async move {
        fcp_map[foo];
    })
    .collect();
```

nikomatsakis: I guess not-- but what error do you get when you do so?

nikomatsakis: also, do we want to compute this is `include_fcp_details` is false?

src/github.rs
```diff
@@ -1046,42 +1045,94 @@ pub struct Query<'a> {
...
+                    );
+
+                    if let Some(fcp) = fcp_map.get(&key) {
+                        let bot_tracking_comment_html_url = format!(
+                            "{}#issuecomment-{}",
+                            issue.html_url, fcp.fcp.fk_bot_tracking_comment
+                        );
+                        let bot_tracking_comment_content = quote_reply(&fcp.status_comment.body);
+                        let fk_initiating_comment = fcp.fcp.fk_initiating_comment;
+                        let initiating_comment_html_url =
```
nikomatsakis, line 1093: probably orthogonal, but I'm surprised we don't use some library for this sort of thing

src/github.rs
nikomatsakis: I think I expect this `bool` to be part of the `Self` type here

nikomatsakis (commented):
Overall, seems kind of like I expected! Left some nits and questions.

src/agenda.rs
chazkiker2: looks like that is just formatting, I will happily revert this to how it was before. 

Not sure if there's a rustfmt style on this repo or not — is there some sort of config file that would indicate whether that's true or not? 

*Could* just be my VSCode settings doing their thing — will happily change them to make them stop auto-formatting everything if need be

chazkiker2: Reverted in cdc9231

src/github.rs
chazkiker2: When I move it out, I get the following error:

```txt
error[E0507]: cannot move out of `fcp_map`, a captured variable in an `FnMut` closure
    --> src/github.rs:1072:37
     |
1069 |            let fcp_map = crate::rfcbot::get_all_fcps().await?;
     |                ------- captured outer variable
...
1072 |                .map(|issue| async move {
     |  ___________________-__________________^
     | | __________________|
     | ||
1073 | ||                 let fcp_details = if include_fcp_details {
1074 | ||                     let repository_name = if let Some(repo) = issue.repository.get() {
1075 | ||                         repo.repository.clone()
...    ||
1086 | ||                     if let Some(fcp) = fcp_map.get(&key) {
     | ||                                        -------
     | ||                                        |
     | ||                                        move occurs because `fcp_map` has type `HashMap<std::string::String, FullFCP>`, which does not implement the `Copy` trait
     | ||                                        move occurs due to use in generator
...    ||
1131 | ||                 })
1132 | ||             })
     | ||             ^
     | ||_____________|
     | |______________captured by this `FnMut` closure
     |                move out of `fcp_map` occurs here
```

I'll try to reproduce w/ a simpler use case in playground as well.

chazkiker2: let fcp_map = crate::rfcbot::get_all_fcps().await?;

chazkiker2: And we do not want to compute this if `fcp_details` is false, that's a good catch

src/github.rs
chazkiker2: are you referring to accessing the comment link or to the `quote_reply` function?

chazkiker2: unrelated but I'm similarly surprised we're using a home-baked parser rather than a command line library (like clap) to take in commands — surely wouldn't be too hard to treat a github comment like command line input

src/github.rs
chazkiker2: Was able to work around this by changing the `issues.iter().map( ... )` call into a `for issue in issues`

See changes in c7d2261... is there anything wrong with this? Seems altogether worth is to only call the rcbot endpoint one time

src/github.rs
nikomatsakis: I meant something like https://crates.io/crates/octocrab

chazkiker2: It is kind of interesting, especially b/c octocrab *is* in here, but nothing in `src/github.rs` seems to use it 

https://github.com/rust-lang/triagebot/blob/2a37c453c01f0d50d5266cab2318a729eb217f64/src/handlers.rs#L250-L255

src/github.rs
chazkiker2: Able to replicate the error in [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a4631b395d8693a21552e7a2f472e167)

chazkiker2:
> > Seems like I trip GitHub's rate limit within just a handful of runs — may have just been b/c I was running locally over and over but it could also be a problem
> 
> 
> 
> That's not great. That said, I typically run this script once per week...
> 
> 
> 
> ...still, I wonder if there's a way to combine this into a single graphql query or something.

given github's GQL, i believe this could be possible... but may be worth splitting into another story.

src/github.rs
Mark-Simulacrum: I think it seems *ok* for now, it doesn't feel like there's an obvious right thing. We can refactor in the future as needed.

Mark-Simulacrum:
@chazkiker2 Can you say if this seems ready to go on your end? My feeling is that we should probably go ahead and merge presuming it mostly works, and then if there's bugs to be fixed or rate limiting etc we can either revert or fix those depending on complexity.

Mark-Simulacrum:
Yeah, it seems a little messy, but since we don't have obvious next steps we can probably merge. With https://github.com/rust-lang/triagebot/issues/1599 there might be simplification possible but seems also a little difficult to put a finger on.

I'll merge this tomorrow to let some of the changes I merged today go live first, I think.

chazkiker2:
@Mark-Simulacrum okay sounds like a decent plan :). sorry for the slowness!

chazkiker2:
@Mark-Simulacrum btw i don’t think i have merge access on this repo, so feel free to merge in whenever you’re ready

PR #1597, opened 2022-03-27
Add shortcut '@rustbot blocked' to apply S-blocked

dtolnay:


src/handlers/shortcut.rs
```diff
@@ -25,17 +25,24 @@ pub(super) async fn handle_command(
...
+        ShortcutCommand::Blocked => blocked,
     };
 
     if !issue_labels.iter().any(|l| l.name == add) {
-        issue.remove_label(&ctx.github, remove).await?;
+        for remove in status_labels {
+            if remove != add {
+                issue.remove_label(&ctx.github, remove).await?;
+            }
+        }
```
apiraino, line 45: @dtolnay the intended behaviour here is to have three mutually excluding statuses for a PR:  either `S-waiting-on-author`, `S-waiting-on-review` or `S-blocked`, is my understanding correct?

If yes, does it apply to any scenario? I see a couple of (admittedly old and perhaps with incorrect labelling) PRs that are `S-blocked` _and_ either waiting [for review](https://github.com/rust-lang/rust/issues?q=label%3AS-blocked+label%3AS-waiting-on-review) or [for feedback from the author](https://github.com/rust-lang/rust/issues?q=label%3AS-blocked+label%3AS-waiting-on-author).

dtolnay: Whether or not they are mutually exclusive, I believe this shortcut is useful.

Notice that there is a [much larger number of PRs](https://github.com/rust-lang/rust/issues?q=label%3AS-waiting-on-author+label%3AS-waiting-on-review) which are simultaneously <kbd>S-waiting-on-author</kbd> and <kbd>S-waiting-on-review</kbd> despite rustbot's existing `author` and `ready` shortcuts treating those as "mutually exclusive". It makes perfect sense that something can be waiting on both the author and the reviewer: for example a half-reviewed PR in which the first half comments can be worked on while the second half is being reviewed.

But the point of the `rustbot author` and `rustbot ready` shortcuts is to signal a handoff between the author and the reviewer: one of them is saying "I am done now, your turn".

Similarly `rustbot blocked` signals "we are both done for now, nobody's turn".

Regarding whether the PRs in your links are mistagged, it's possible <kbd>S-blocked</kbd> is being used for 2 different purposes:

- This PR cannot make progress
- This PR can make progress but something else must happen first before it may be landed

The first one is clearly mutually exclusive with <kbd>S-waiting-on-author</kbd> and <kbd>S-waiting-on-review</kbd> because both of those are a form of progress.

The second one may be more accurately captured by a <kbd>do not merge</kbd> label rather than <kbd>S-blocked</kbd> (and @bors could learn to respect it). I've seen PRs that are ready for review, but aren't supposed to land until a PR in a different crate has been released to crates.io, or a tracking issue has been filed. I do not personally consider those "blocked".

Mark-Simulacrum (approved):

Issue #1483, opened 2021-09-01
missing rustc commits


Mark-Simulacrum:
We are not infrequently losing rustc commits from the database log. We should try to prevent this by auto-healing more reliably.

A query like `select parent_sha from rustc_commits where parent_sha not in (select sha from rustc_commits) and time > '2021-01-01';` against the database can give us the list of SHAs we're missing (potentially incomplete, but these should all definitely be in the DB). From there, we should be able to query GitHub for their information (PR, parent SHA, time) and fill the database in.

It is likely that doing this instead of / in place of the current logic which attempts to "scan back" on a git push to rust-lang/rust is preferable, and should be fairly reliable. We will need to be careful to avoid falling into any infinite loops, but that should not be terribly hard.

Implementing this should also hopefully be not too hard, and I'd be happy for someone to invest time into doing so -- please feel free to ping me on Zulip (or this issue). I'll likely get to it eventually because it's becoming a hassle to re-add commits manually, but that may take some time.

Kobzol:
I'll try to write up my understanding of the whole process, it could also serve as some basic
documentation for future reference.

So my understanding is that currently the `triagebot` stores Rust `master` commits into database
by listening for comments on PRs. If it finds a comment that is from `bors` and deals with a merge
into `master`, is gets the SHA from it, then it downloads data from GH based on this SHA and stores
the data into the database.

Why does the bot scan comments instead of listening for push events to `master`? Would it
be more difficult to find the corresponding PR from the push?

When a merge commit happens, I'm not exactly sure what are the corresponding SHAs. Is it like this?
sha: commit of the new (merged) merge commit
parent sha: parent of the merged commit, i.e. the previous merge commit

Then when the commit is recorded, its parent shas are fetched recursively. I'm not sure how the
stop condition works though - it seems that the search stops when a SHA can't be resolved or
when it's message doesn't contain a PR number (or when it's already in the database?).

Why is the recursive search performed in the first place? Do the events sometimes fail and then
the commits are missed? How can it happen that some commit is missing?

When it happens, what problems does it cause? `/bors-commit-list` shows some commit that is then
used by `rustc-perf`, but the commit does not actually exist in `master`?

Mark-Simulacrum:
> So my understanding is that currently the triagebot stores Rust master commits into database
by listening for comments on PRs. If it finds a comment that is from bors and deals with a merge
into master, is gets the SHA from it, then it downloads data from GH based on this SHA and stores
the data into the database.

[...]

PR #1598, opened 2022-03-30
Make missing commit lookup more robust

Kobzol:
Fixes: https://github.com/rust-lang/triagebot/issues/1483


src/handlers/rustc_commits.rs
```diff
@@ -70,19 +72,30 @@ pub async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
...
+    let mut pr = Some(pr);
+
+    // List of roots to be resolved. Each root and its parents will be recursively resolved
+    // until an existing commit is found.
+    let mut to_be_resolved = VecDeque::new();
+    to_be_resolved.push_back(sha.to_string());
+    to_be_resolved.extend(get_missing_commits(&db).await);
+
+    while !to_be_resolved.is_empty() {
+        let sha = to_be_resolved.pop_front().unwrap();
```
Mark-Simulacrum, line 92: `while let Some(sha) = to_be_resolved.pop_front()` probably is slightly nicer here.

I also don't know that we need a queue vs. just a simple Vec, but either seems OK.

Mark-Simulacrum:
I think this looks pretty good, so happy to merge with the nit fixed.

Kobzol:
I refactored the `while let`. `VecDeque` probably isn't strictly necessary, but it shouldn't hurt anything.

I don't have a way to test the change properly though, we'll probably have to see how it works in production.

PR #1610, opened 2022-04-13
Comment on issues when `@rustbot label` is given an invalid label

jyn514:
Previously, the label was just silently ignored, along with all other labels
in the same command. This tells the user what went wrong, and also adds all valid labels.

I'm not sure how to test this :(

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -366,16 +366,35 @@ impl IssueRepository {
         )
     }
 
-    async fn has_label(&self, client: &GithubClient, label: &str) -> bool {
+    async fn has_label(&self, client: &GithubClient, label: &str) -> anyhow::Result<bool> {
         #[allow(clippy::redundant_pattern_matching)]
         let url = format!("{}/labels/{}", self.url(), label);
-        match client.send_req(client.get(&url)).await {
-            Ok(_) => true,
-            // XXX: Error handling if the request failed for reasons beyond 'label didn't exist'
-            Err(_) => false,
+        match client._send_req(client.get(&url)).await {
+            Ok((_, _)) => Ok(true),
+            Err(e) => {
+                if e.downcast_ref::<reqwest::Error>().map_or(false, |e| e.status() == Some(StatusCode::NOT_FOUND)) {
+                    Ok(false)
+                } else {
+                    Err(e)
+                }
+            }
         }
     }
 }
+
+#[derive(Debug)]
+pub(crate) struct UnknownLabels {
+    labels: Vec<String>,
+}
+
+// NOTE: This is used to post the Github comment; make sure it's valid markdown.
+impl fmt::Display for UnknownLabels {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Unknown labels: {}", &self.labels.join(", "))
+    }
+}
+
+impl std::error::Error for UnknownLabels {}
 
 impl Issue {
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
A couple notes:

* I don't think this belongs in the github.rs API -- if we want to provide metadata, that should be returned to the caller. We can have a helper function which puts a comment, too, but I'd prefer to avoid add_labels having side effects beyond adding labels :)
* No need to test, we typically just merge and test in production for triagebot. Not critical enough to worry about.

I'm also a little uncertain about the approach. I think historically the choice to silently eat bad labels has been semi-intentional, as it means that we can avoid lots of comments if you edit a comment trying to fix any typos you made. This definitely prioritizes "advanced" users over newcomers though; which might well be the wrong tradeoff.

I'm not sure what a good way to thread the needle on that tradeoff is. It seems like either way we're not really delivering a great experience.

Regardless, my inclination is to make label adding atomic, like we did before. I think it's much less confusing what to do when we've failed to add in general, and prevents awkwardness around label-based triggering on "half" of the labels you added (since that triggering can have exclusions).

jyn514:
> Regardless, my inclination is to make label adding atomic, like we did before.

👍 This makes sense to me and will simplify moving the changes out of github.rs. I'll try and get to it this week.

jyn514:
Ah, I just realized that `add_labels` is used in many places besides `autolabel` - the tradeoffs are making more sense to me now. I changed this to return an error on unknown labels that only autolabel checks for, and all other commands just bubble up to the top-level as a log. I also ensured that no labels are added if unknown labels are present. Note that this is not *quite* the same as "atomic" because autolabel calls `add_labels(); remove_labels();` right after each other, but it matches the previous behavior.

src/github.rs
```diff
@@ -366,17 +366,36 @@ impl IssueRepository {
...
 
+#[derive(Debug)]
+pub(crate) struct UnknownLabels {
+    labels: Vec<String>,
+}
+
+impl fmt::Display for UnknownLabels {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("Unknown labels:\n- ")?;
+        write!(f, "{}", &self.labels.join("\n- "))
```
Mark-Simulacrum, line 393: I am not a fan of a multi-line error message here -- is there a reason to not just comma-separate and list them in one message? I think being concise (visually) is good here and the list doesn't give us that much value.

src/handlers/autolabel.rs
```diff
@@ -125,7 +125,18 @@ pub(super) async fn handle_input(
     event: &IssuesEvent,
     input: AutolabelInput,
 ) -> anyhow::Result<()> {
-    event.issue.add_labels(&ctx.github, input.add).await?;
+    match event.issue.add_labels(&ctx.github, input.add).await {
+        Ok(()) => {}
+        Err(e) => {
+            use crate::github::UnknownLabels;
+            if let Some(err @ UnknownLabels{ .. }) = e.downcast_ref() {
```
Mark-Simulacrum, line 132:
```suggestion
            if let Some(err @ UnknownLabels { .. }) = e.downcast_ref() {
```

src/github.rs
Mark-Simulacrum: Can we also add a comment here about this being used as a github comment text? That way it's clearer when editing this piece what syntax etc we can use.

jyn514: Sure thing, implemented both :)

PR #1622, opened 2022-06-10
Show human readable last review date

apiraino:
Fixes the last review date for various PRs.

Before
```
[T-compiler](....)
-  "Mir-Opt for copying enums with large discrepancies" [rust#85158](https://github.com/rust-lang/rust/pull/85158) (last review activity: 2022-03-28T12:45:59.324310806Z)
```

After:
```
[T-compiler](....)
-  "Mir-Opt for copying enums with large discrepancies" [rust#85158](https://github.com/rust-lang/rust/pull/85158) (last review activity: 2 months ago)
```


src/actions.rs
```diff
@@ -43,7 +43,9 @@ pub struct IssueDecorator {
     pub repo_name: String,
     pub labels: String,
     pub assignees: String,
-    pub updated_at: String,
+    pub updated_at: DateTime<Utc>,
```
jyn514, line 46: do we need to store both? can we remove the old `updated_at` field?

apiraino: good point. After a doublecheck, it seems that it can be removed :)
(force pushed the change)

jyn514 (approved):

PR #1624, opened 2022-06-22
Remove redundant queries to rfcbot.rs

apiraino:
Queries to rfcbot.rs are enabled by a boolean flag `include_fcp_details`, so when this flag is false
it's not necessary to run those queries.

r? @spastorino

thanks for the review!

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -6,6 +6,7 @@ use hyper::header::HeaderValue;
 use once_cell::sync::OnceCell;
 use reqwest::header::{AUTHORIZATION, USER_AGENT};
 use reqwest::{Client, Request, RequestBuilder, Response, StatusCode};
+use std::collections::HashMap;
 use std::convert::TryInto;
 use std::{
     fmt,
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Is there some problem with these queries (other than wasted bandwidth)? I would expect rfcbot.rs to be pretty fast, but I could be wrong.

apiraino:
No other problem to my knowledge, just redundant queries.

src/github.rs
```diff
@@ -1093,7 +1094,11 @@ impl<'q> IssuesQuery for Query<'q> {
             .await
             .with_context(|| "Unable to get issues.")?;
 
-        let fcp_map = crate::rfcbot::get_all_fcps().await?;
+        let fcp_map = if include_fcp_details {
+            crate::rfcbot::get_all_fcps().await?
+        } else {
+            HashMap::new()
+        };
```
spastorino, line 1101: Why don't you place `let fcp_map = crate::rfcbot::get_all_fcps().await?;` on line https://github.com/rust-lang/triagebot/pull/1624/files#diff-ef5fd0828e2727ce9e100f72dc4f18f0f126e393d35108269a9bea5e8a104d5cR1117

spastorino (changes requested):

PR #1625, opened 2022-06-26
Add mentions.

ehuss:
This adds a feature to add comments to PRs that touch certain files to ping interested people or to give information to the author. This is intended to take over the responsibility from highfive.

One of the key changes compared to highfive is that this can handle subsequent pushes to a PR (fixing https://github.com/rust-lang/highfive/issues/223). To avoid duplicate mentions in that scenario, this PR introduces the concept of "issue data" which stores data about an issue in the Postgres database.

### DIFF
--- a/src/handlers/mentions.rs
+++ b/src/handlers/mentions.rs
@@ -98,7 +98,7 @@ pub(super) async fn handle_input(
             None => write!(result, "Some changes occurred in {to_mention}").unwrap(),
         }
         if !reviewers.is_empty() {
-            write!(result, "\n\ncc {}", reviewers.join(",")).unwrap();
+            write!(result, "\n\ncc {}", reviewers.join(", ")).unwrap();
         }
         state.paths.push(to_mention.to_string());
     }
--- a/src/handlers/mentions.rs
+++ b/src/handlers/mentions.rs
@@ -1,0 +1,114 @@
+//! Purpose: When opening a PR, or pushing new changes, check for any paths
+//! that are in the `mentions` config, and add a comment that pings the listed
+//! interested people.
+
+use crate::{
+    config::{MentionsConfig, MentionsPathConfig},
+    db::issue_data,
+    github::{files_changed, IssuesAction, IssuesEvent},
+    handlers::Context,
+};
+use anyhow::Context as _;
+use serde::{Deserialize, Serialize};
+use std::fmt::Write;
+use std::path::Path;
+use tracing as log;
+
+const MENTIONS_KEY: &str = "mentions";
+
+pub(super) struct MentionsInput {
+    paths: Vec<String>,
+}
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+struct MentionState {
+    paths: Vec<String>,
+}
+
+pub(super) async fn parse_input(
+    ctx: &Context,
+    event: &IssuesEvent,
+    config: Option<&MentionsConfig>,
+) -> Result<Option<MentionsInput>, String> {
+    let config = match config {
+        Some(config) => config,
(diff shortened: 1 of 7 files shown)


src/handlers/mentions.rs
```diff
@@ -0,0 +1,114 @@
...
+        })
+        .unwrap_or_default()
+    {
+        let files = files_changed(&diff);
+        let file_paths: Vec<_> = files.iter().map(|p| Path::new(p)).collect();
+        let to_mention: Vec<_> = config
+            .paths
+            .iter()
+            // Only mention matching paths.
+            // Don't mention if the author is in the list.
```
Mark-Simulacrum, line 60: We'll probably want to consider mapping this through GitHub teams, but not worth doing that right now.

src/handlers/mentions.rs
```diff
@@ -0,0 +1,114 @@
...
+pub(super) async fn handle_input(
+    ctx: &Context,
+    config: &MentionsConfig,
+    event: &IssuesEvent,
+    input: MentionsInput,
+) -> anyhow::Result<()> {
+    let client = ctx.db.get().await;
+    let mut state: MentionState = issue_data::load(&client, &event.issue, MENTIONS_KEY)
+        .await?
+        .unwrap_or_default();
```
Mark-Simulacrum, line 84: Hm, so I'm a little worried that we can have pretty fast near-simultaneous events/pushes arriving, which means that the time between loading and saving the state could be significant. It seems like we'd ideally serialize around the primary key in the table -- maybe a postgres transaction is a good way to do that? Not sure.

That's the primary advantage of using the database for this -- we already have a rudimentary state mechanism via `data_section` in src/interactions.rs, which uses the issue/PR description. I think the database is probably a better mechanism for any state like this that's more automatically driven (rather than user interaction), and in general is easier to interact with and avoids polluting the issue description with garbage (even if theoretically hidden).

src/handlers/mentions.rs
```diff
@@ -0,0 +1,114 @@
...
+        let MentionsPathConfig { message, reviewers } = &config.paths[to_mention];
+        if !result.is_empty() {
+            result.push_str("\n\n");
+        }
+        match message {
+            Some(m) => result.push_str(m),
+            None => write!(result, "Some changes occurred in {to_mention}").unwrap(),
+        }
+        if !reviewers.is_empty() {
+            write!(result, "\n\ncc {}", reviewers.join(",")).unwrap();
```
Mark-Simulacrum, line 101:
```suggestion
            write!(result, "\n\ncc {}", reviewers.join(", ")).unwrap();
```

Probably want to add spaces, right?

src/handlers/mentions.rs
ehuss: Yea, that's probably wise. My postgres skills are pretty rusty, and I'm not sure if there is an elegant way to do that with this primitive JSON blob schema.  I pushed an update that uses a crude "LOCK TABLE".  There's some risk that concurrent events could cause significant delays, but I don't feel particularly worried about that.

Mark-Simulacrum: This looks okay to me. I think we could maybe use a row lock instead, but that's mostly just an optimization and doesn't seem worth doing right now (can be done later).

Mark-Simulacrum:
Thanks! I'm going to go ahead and merge this tomorrow, so I can be around to help resolve issues if something goes wrong. Not expecting any problems though.

Once this is ready we can post a PR adding/migrating high-five mentions to triagebot and after a short period of duplicate mentions remove the high-five configuration for rust-lang/rust (and other repositories eventually).

PR #1627, opened 2022-06-27
Add some more docs on running triagebot.

ehuss:
This provides a little more docs on how Triagebot works, and how you can run it locally.  I found this not *too* difficult to set up, and makes it much easier to test and iterate.


jyn514:
This is awesome, thanks!

PR #1630, opened 2022-06-28
Fix mentions elided for the author.

ehuss:
The mentions code was intended to not ping if the author is in the mention list. The author check was not working because the `@` symbol was in the way. This fixes it to strip the `@` so the author check works.


jyn514 (approved):
Thanks!! I was trying to figure out why this working but never got around to running triagebot locally :)

jyn514:
Err I may be thinking of a different part of the codebase actually... But either way this fix looks right.

PR #1631, opened 2022-06-28
Don't post mentions on rollups.

ehuss:
People don't need to be pinged on a rollup PR, so this tries to skip them.

An alternate approach would be to scan the issue body for something like `@rustbot modify labels: rollup`, but I think looking at the title should work well enough.

This won't handle beta backports, though.  I'm wondering if maybe it should ignore the beta branch?  I don't know if those pings will happen often enough to be annoying.

### DIFF
--- a/src/handlers/mentions.rs
+++ b/src/handlers/mentions.rs
@@ -42,6 +42,11 @@ pub(super) async fn parse_input(
         return Ok(None);
     }
 
+    // Don't ping on rollups.
+    if event.issue.title.starts_with("Rollup of") {
+        return Ok(None);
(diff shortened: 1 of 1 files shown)


jyn514:
Can we look for the `rollup` label instead? I think that should be more reliable.

I guess it's not added until after the PR is created though ... would be nice if we could change bors to have it in the initial PR so we can depend on it.

Mark-Simulacrum:
I'm going to merge this for now to avoid noise for contributors; we can improve it to handle the edge cases (e.g., beta, etc) later.

Issue #1637, opened 2022-07-17
Support for "silent" PRs?


RalfJung:
PRs like https://github.com/rust-lang/rust/pull/99368 cause a lot of premature pings -- it's just a perf experiment, we don't have to all know about that.

So maybe it should be possible to somehow tell triagebot to not do mentions for a particular PR? We have a special exception for rollups that does this I think, but other PRs might also want it. Ideally, if it turns out the perf experiment gives the expected results and we *do* want to land that PR, there is a way to tell triagebot to now please do the pings that have been suppressed so far.

matthiaskrgr:
Triagebot could ignore the pings as long as the ghost user is assigned and do the ping only once someone requests an actual reviewer via `r?` maybe?

RalfJung:
I guess that would somewhat replicate the old "accidentally silent" system that highfive had.

However, for Miri PRs, I do like the rustbot pings so I don't have to do them. I guess it could do the pings both on `r?` and when it sees a bors message saying "commit blah has been approved"?

Mark-Simulacrum:
I think using GitHub's draft PR status might be a good fit here. It won't prevent any interaction with bots (including bors, afaik) and seems like a reasonable "should be silent as an experiment that won't land" status. Also easy to notice and know to clear as an author, hopefully.

I think we may want a separate mechanism too (similar to special case for rollups), that's just indicative that this PR shouldn't be notified on - but I'm not sure how large the remaining territory is where pings are annoying and draft PR doesn't make sense.

ehuss:
Turning off mentions for draft PRs sounds like a good idea to me. I posted #1638 to implement that.

Unfortunately I suspect most people won't know to use draft PRs for things like performance testing, so I'm not sure it will be particularly effective.  I also like the idea of not pinging on unassigned/ghost PRs, but it sounds like that won't work for Ralf's case.  I also suspect that any "don't mention" tagging mechanism won't be particularly effective since just like draft PRs, I don't think people will know to do that.  

I'm not sure I have any ideas other than looking at the assignee to avoid it.

RalfJung:
Is it even possible to mark a PR as draft during creation (so that it applies for the initial pings)?

jyn514:
> Is it even possible to mark a PR as draft during creation (so that it applies for the initial pings)?

Yes, it shows up on the "Create Pull Request" button (there's a little drop-down that lets you pick "Create Draft Pull Request" instead).

RalfJung:
[...]

PR #1638, opened 2022-07-17
Don't post mentions for draft PRs.

ehuss:
The mentions can be quite noisy, so this provides a way to prevent them from being posted. Presumably a draft PR is under construction and may change over time, so the mentions may be premature. Once the author clicks "Ready for review", triagebot will re-scan the PR and post any mentions that haven't already been posted.

cc #1637

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -255,6 +255,8 @@ pub struct Issue {
     pub pull_request: Option<PullRequestDetails>,
     #[serde(default)]
     pub merged: bool,
+    #[serde(default)]
+    pub draft: bool,
     // API URL
     comments_url: String,
     #[serde(skip)]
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum:
We should try to see if we can ping the moment a PR exits draft status, otherwise it can move to being merged without being pinged very easily.

(Not sure what hook to listen on for that, if any).

We probably also ought to make bors error on r+ for draft PRs, but that's less important.

ehuss:
> We should try to see if we can ping the moment a PR exits draft status, otherwise it can move to being merged without being pinged very easily.

Perhaps I'm misunderstanding the question, but this should do that.  When the user clicks "Ready for Review", the webhook sends a `ReadyForReview` action, which should then immediately post any mentions.  [here](https://github.com/ehuss/triagebot-test/pull/15) is an example where immediately after the "marked this pull request as ready" event, it posts the mention comment.

Mark-Simulacrum:
Oh, yes, indeed. Sorry, must have read the diff too quickly.

This looks good to me then, thanks!

PR #1639, opened 2022-08-11
Don't stop generating releases on first failure

Mark-Simulacrum:


jyn514 (approved):
Seems fine, but when do we ever create multiple releases at once?

Mark-Simulacrum:
For context -- this is trying to debug failure to generate 1.63.0 release notes; log-diving showed us failing on generating 1.34.2 release notes. I'm not sure why we're generating those (they seem to already exist), but in any case this should let us move forward.

PR #1641, opened 2022-08-22
mentions: Only supress mention when just the author is to be CC'd.

aDotInTheVoid:
See https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Triagebot.20mentions.20not.20working.20for.20rustdoc-json for more detail.

### DIFF
--- a/src/handlers/mentions.rs
+++ b/src/handlers/mentions.rs
@@ -62,13 +62,13 @@ pub(super) async fn parse_input(
             .paths
             .iter()
             // Only mention matching paths.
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (approved):

ehuss:
@aDotInTheVoid This seems to have broken mentions that don't have a `cc` field. There are some that are configured to post a message without mentioning any specific users. Can you take a look at posting a fix for that? I think the filter just needs to not filter when `cc` is empty.

PR #1642, opened 2022-08-31
handlers: add no merge policy notifications

davidtwco:
Inspired by [discussion on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Disallowing.20PRs.20from.20fork's.20master.20branch/near/294672188).

Add a handler for issue events that checks whether a merge commit has been added to the pull request and informs the user of the project's no merge policy.

cc @Mark-Simulacrum

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -749,6 +749,33 @@ impl Issue {
         Ok(Some(String::from(String::from_utf8_lossy(&diff))))
     }
 
+    /// Returns the commits from this pull request (no commits are returned if this `Issue` is not
+    /// a pull request).
+    pub async fn commits(&self, client: &GithubClient) -> anyhow::Result<Vec<GithubCommit>> {
+        if !self.is_pr() {
+            return Ok(vec![]);
+        }
+
+        let mut commits = Vec::new();
+        let mut page = 1;
+        loop {
+            let req = client.get(&format!(
+                "{}/pulls/{}/commits?page={page}&per_page=100",
+                self.repository().url(),
+                self.number
+            ));
+
+            let new: Vec<_> = client.json(req).await?;
+            if new.is_empty() {
+                break;
+            }
+            commits.extend(new);
+
+            page += 1;
+        }
+        Ok(commits)
+    }
+
     pub async fn files(&self, client: &GithubClient) -> anyhow::Result<Vec<PullRequestFile>> {
         if !self.is_pr() {
             return Ok(vec![]);
--- a/src/handlers/no_merges.rs
+++ b/src/handlers/no_merges.rs
@@ -1,0 +1,129 @@
+//! Purpose: When opening a PR, or pushing new changes, check for merge commits
+//! and notify the user of our no-merge policy.
+
+use crate::{
+    config::NoMergesConfig,
+    db::issue_data::IssueData,
+    github::{IssuesAction, IssuesEvent},
+    handlers::Context,
+};
+use anyhow::Context as _;
+use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
+use std::fmt::Write;
+use tracing as log;
+
+const NO_MERGES_KEY: &str = "no_merges";
+
+pub(super) struct NoMergesInput {
+    /// Hashes of merge commits in the pull request.
+    merge_commits: HashSet<String>,
+}
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+struct NoMergesState {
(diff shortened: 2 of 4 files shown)


davidtwco:
This passes a `cargo check` but I haven't yet had the time to set up a proper environment to check it actually works, so opening as a draft.

src/github.rs
```diff
@@ -749,6 +749,21 @@ impl Issue {
...
 
+    /// Returns the commits from this pull request (no commits are returned if this `Issue` is not
+    /// a pull request).
+    pub async fn commits(&self, client: &GithubClient) -> anyhow::Result<Vec<Commit>> {
+        if !self.is_pr() {
+            return Ok(vec![]);
+        }
+
+        let req = client.get(&format!(
+            "{}/pulls/{}/commits",
```
Mark-Simulacrum, line 760: Note that this will only return at most 30 by default, I'd guess -- I think we should paginate here (shouldn't be too hard, can steal code from elsewhere in the file most likely). It's less important for this specific use case but could bite us later.

(Probably pages of 100 make sense).

src/github.rs
```diff
@@ -763,6 +778,12 @@ impl Issue {
     }
 }
 
+#[derive(Debug, serde::Deserialize)]
+pub struct Commit {
```
Mark-Simulacrum, line 782: Hm, any reason to not use `GithubCommit` here?

src/handlers/no_merges.rs
```diff
@@ -0,0 +1,141 @@
...
+            log::error!("failed to fetch commits: {:?}", e);
+        })
+        .unwrap_or_default();
+    for commit in commits {
+        if commit.parents.len() > 1 {
+            merge_commits.insert(commit.sha.clone());
+        }
+
+        if merge_commits.is_empty() {
+            before_first_merge_commit = Some(commit.sha);
```
Mark-Simulacrum, line 69: I'm not sure I follow this logic. Is there really an ordering guarantee on the commits returned? I would expect the answer is no, or that the answer is "yes" (but that guarantee is by commit date, which is useless here).

I think github returns the base commit via the get pull request API (https://docs.github.com/en/rest/pulls/pulls#get-a-pull-request), which might be a better mechanism for this.

Mark-Simulacrum (commented):
I wonder if we'd get a Github-native warning by adding "require linear history" to the branch protections on rust-lang/rust or whatever other repositories. Maybe worth experimenting with? We obviously don't actually have a linear history, so we'd need to bypass that in bors, but maybe then PRs would get warnings of some kind?

davidtwco:
> I wonder if we'd get a Github-native warning by adding "require linear history" to the branch protections on rust-lang/rust or whatever other repositories. Maybe worth experimenting with?

I've received feedback that GitHub will offer a "fix this conflict" button of some sort that adds a merge commit, which is immediately incorrect, so we could reduce the incidences of this happening by preventing GitHub from offering that option.

src/github.rs
davidtwco: Just didn't notice it existed, fixed.

src/github.rs
davidtwco: Added support for pagination.

src/handlers/no_merges.rs
davidtwco: I've just removed this and always suggest `master` - my thinking was that if the commits were `A -> B -> C -> D (Merge)` then it would suggest the sha for the parent commit in the pull request of the merge commit - `C` here.

Mark-Simulacrum (approved):

Mark-Simulacrum:
@davidtwco this is still marked as a draft, is that state accurate? Or are we good to go?

davidtwco:
> @davidtwco this is still marked as a draft, is that state accurate? Or are we good to go?

I haven't taken the time to set up a development environment for this locally and actually check it does what I expect it to (just because that seems like something that I'll need to block out some time to do).

Mark-Simulacrum:
Oh. Well, feel free to if you want, but we can also merge this and then enable it (perhaps not on rust-lang/rust) and test it that way. Just let me know if you'd prefer that; it's pretty typical of triagebot PRs.

davidtwco:
> Oh. Well, feel free to if you want, but we can also merge this and then enable it (perhaps not on rust-lang/rust) and test it that way. Just let me know if you'd prefer that; it's pretty typical of triagebot PRs.

Sorry for the delay here, we can test it after merging I think.

PR #1645, opened 2022-09-03
mentions: Correctly handle empty CC list

### LINKED ISSUE #1644: Messages not sent when cc is empty
@aDotInTheVoid This seems to have broken mentions that don't have a `cc` field. There are some that are configured to post a message without mentioning any specific users. Can you take a look at posting a fix for that? I think the filter just needs to not filter when `cc` is empty.

_Originally posted by @ehuss in https://github.com/rust-lang/triagebot/issues/1641#issuecomment-1236188086_

aDotInTheVoid:
Closes #1644

### DIFF
--- a/src/handlers/mentions.rs
+++ b/src/handlers/mentions.rs
@@ -61,14 +61,16 @@ pub(super) async fn parse_input(
         let to_mention: Vec<_> = config
             .paths
             .iter()
-            // Only mention matching paths.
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (approved):

ehuss:
Thanks @aDotInTheVoid!

PR #1646, opened 2022-09-05
Fix RFC rendered linkifier.

ehuss:
The RFC "Rendered" link generator wasn't working. The problem is that the `Push` event wasn't marking the payload as a PR. When the link generator was looking for the list of files, the `files` method was returning nothing.

I removed the empty `PullRequestDetails` struct since I don't foresee that ever gaining any fields (the fields usually are part of `Issue`). I think the bool makes the meaning a little clearer.

Additionally I changed the format of the link URL. I prefer to have links directly to the author's branch so that it can reliably display the latest content. Linking to the SHA blob URL risks that the link gets out of date and reviewers may be looking at stale content.  The drawback with the branch-based approach is that the link will die when the branch is deleted or reset. I think that is relatively good tradeoff, as I think the Rendered link should be updated when the RFC is closed. Perhaps that can be automated in the future.

Finally, I added some comments along the way to help understand what all these structs/fields/variants are doing.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -1256,6 +1256,9 @@ pub enum Event {
     /// - Regular comment on an issue or PR.
     /// - A PR review.
     /// - A comment on a PR review.
+    ///
+    /// These different scenarios are unified into the `IssueComment` variant
+    /// when triagebot receives the corresponding webhook event.
     IssueComment(IssueCommentEvent),
     /// Activity on an issue or PR.
     Issue(IssuesEvent),
--- a/src/github.rs
+++ b/src/github.rs
@@ -232,11 +232,6 @@ pub struct Label {
 }
 
 #[derive(Debug, serde::Deserialize)]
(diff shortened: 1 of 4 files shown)


src/github.rs
```diff
@@ -1216,11 +1246,21 @@ pub struct PushEvent {
...
 #[derive(Debug)]
 pub enum Event {
+    /// A Git branch or tag is created.
     Create(CreateEvent),
+    /// A comment on an issue or PR.
+    ///
+    /// Can be:
+    /// - Regular comment on an issue or PR.
+    /// - A PR review.
+    /// - A comment on a PR review.
```
Mark-Simulacrum, line 1259: Maybe worth a note that this is also handled in triagebot to unify these three events together.

ehuss: Sure, sounds good!

Mark-Simulacrum (approved):

PR #1649, opened 2022-09-10
Fix PR comment deserialization

Mark-Simulacrum:
Broke in https://github.com/rust-lang/triagebot/pull/1646.

```
request failed: WebhookError(IssueCommentEvent failed to deserialize

Caused by:
    0: at Path { segments: [Map { key: "issue" }, Map { key: "pull_request" }] }
    1: invalid type: map, expected a boolean at line 1 column 6086)
```


ehuss:
Apologies!  In hindsight that was an obvious mistake and I should have reviewed the API more closely.  Thanks for the fix!

PR #1652, opened 2022-09-11
Fix commands in pull request comments

### LINKED ISSUE #1651: @rustbot ready and @rustbot author erroring out in a few cases
on this PR: https://github.com/rust-lang/rust/pull/95696#issuecomment-1242892564

Tried using `@rustbot ready` shortcut to switch labels and got an error 
> Error: The "Ready" shortcut only works on pull requests.

I don't know what is different about this pull request that produces such an error.

Zulip:
https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/.40rustbot.20author.20not.20working

joboet:
Fixes #1651. Pull request comments are detected by checking the presence of the [pull_request field in the issue type](https://docs.github.com/developers/webhooks-and-events/events/issue-event-types). Broke in #1646.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -261,13 +261,10 @@ pub struct Issue {
     pub assignees: Vec<User>,
     /// This is true if this is a pull request.
     ///
-    /// Note that this field does not come from GitHub. This is manually added
-    /// when the webhook arrives to help differentiate between an event
-    /// related to an issue versus a pull request.
-    ///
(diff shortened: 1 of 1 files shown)


ehuss:
Sorry about that.  I posted an alternate solution at #1653 that just reverts the change. I think the original behavior is fine, and could potentially add fields to the empty struct in the future.

PR #1653, opened 2022-09-11
Revert pull_request field change.

### LINKED ISSUE #1651: @rustbot ready and @rustbot author erroring out in a few cases

ehuss:
The change to the `pull_request` field in #1646 is causing way more trouble than it is worth. If GitHub *sometimes* populates this field, then it is useful to know if it is a pull request or not.

Fixes #1651.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -231,6 +231,16 @@ pub struct Label {
     pub name: String,
 }
 
+/// An indicator used to differentiate between an issue and a pull request.
+///
+/// Some webhook events include a `pull_request` field in the Issue object,
+/// and some don't. GitHub does include a few fields here, but they aren't
+/// needed at this time (merged_at, diff_url, html_url, patch_url, url).
+#[derive(Debug, serde::Deserialize)]
+pub struct PullRequestDetails {
+    // none for now
+}
+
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum (approved):

apiraino:
For future notice, should this codepath will be changed: #1646 also was breaking the T-compiler agenda generation (specifically commit 3cde23dfb9a6b2e724d0c5e4257aae6cf6276eee) when running the triagebot command `cargo run --bin prioritization-agenda`.

Just mentioning this because some dependencies are not always apparent with this codebase. A good strategy could be to run all three CLI binaries of the triagebot (`compiler`,`lang`,`prioritization-agenda`) for a quick sanity check.

Anyway, no problem, this patch reverts the breaking change :)

Issue #1635, opened 2022-07-07
Scheduled events: actions in X minutes/days/hours


Mark-Simulacrum:
I believe this is somewhat distinct from https://github.com/rust-lang/triagebot/issues/1595, which is more about regularly scheduled events (e.g., every Monday at 1pm ET) -- supporting cron jobs is likely somewhat easier than scheduled events, since we can hardcode checking into triagebot and avoid a database entirely most likely. (Modulo triagebot rebooting at inopportune times). It may be that we actually end up sharing the scheduled events code, with cron jobs queueing an event to happen in X time at first startup if such an event doesn't exist yet.

----

For scheduled events, I believe our design should look something like this:

New postgres table for the events, with these columns (roughly):

* event_id (uuid or equivalent)
* event_name (text, one of a set of constants, most likely, indicating which handler to invoke for the event)
* expected_event_time (datetime, UTC)
* event_metadata (text, JSON-encoded free-form metadata)
* executed_at (timestamp)
* failed (text, can be null)

We'll have a background thread that polls the database and every minute (maybe more often, but we can start with every minute):

* With a table lock:
  * Selects all event_id's with event_time's in the past, failed is null, and executed_at at least 60 minutes ago (intended to make repeat executions rare enough that someone can investigate before we spam the issue/Zulip/whatever)
  * Sets executed_at to current time.
* Invokes handlers for each event, logging failures if any
* If a failure occurs that we can detect, then after logging it we also insert the failure message into the failed column
* on successful completion, delete the event from the table

Alongside this there's probably going to be value in a few helper methods that add events to the table and remove them.

For cron support atop this framework, we'd likely do that by running at startup a little bit of code that checks if there are any events queued for cron-task-$name, and if not, then queue up an event with the appropriate delta from now. We'd then re-queue cron-task events when handling them at the appropriate time. But this can be a separate part of the work.

cc @chazkiker2

chazkiker2:
Seems like a pretty solid game plan :). I will certainly run into questions throughout implementation (esp. cuz I haven't worked with the DB on the rust front a whole lot)

If you were to theorize about how you'd want to split this up into PRs, what would you be thinking. If it was TMF, I'd imagine it'd look something like this...

- database layer (adding table/columns/lock)
- functionality for the background thread and implementing/interacting with the table lock on the code side
- then finally adding in the interaction between github/Zulip/whatever else
- and later on build cron functionality on top of everything

chazkiker2:
[...]

PR #1658, opened 2022-10-03
[FEATURE] Scheduled Jobs

mcass19:
Closes https://github.com/rust-lang/triagebot/issues/1635

Co-authored by @julmontesdeoca.

Aims to resolve a way to run scheduled events/jobs. It follows almost all that's described in the issue.

This is a step of something we will need to implement the [Decision Process](https://lang-team.rust-lang.org/decision_process) feature.

Had a first validation/discussion on [this PR](https://github.com/mcass19/triagebot/pull/1) in the fork for further info.

We are really open to any kind of feedback, and we have a couple of questions/concerns left as comments below.


src/db.rs
```diff
@@ -179,6 +182,42 @@ pub async fn run_migrations(client: &DbClient) -> anyhow::Result<()> {
     Ok(())
 }
 
+pub async fn run_scheduled_jobs(db: &DbClient) -> anyhow::Result<()> {
+    let jobs = get_jobs_to_execute(&db).await.unwrap();
+    println!("jobs to execute: {:#?}", jobs);
+    tracing::trace!("jobs to execute: {:#?}", jobs);
+
+    for job in jobs.iter() {
+        update_job_executed_at(&db, &job.id).await?;
```
mcass19, line 191: Is `executed_at` field name clear? The intention is to store when the job starts running

src/db.rs
```diff
@@ -179,6 +182,42 @@ pub async fn run_migrations(client: &DbClient) -> anyhow::Result<()> {
...
+
+        match handle_job(&job.name, &job.metadata).await {
```
mcass19, line 193: How do you think we should run each handler async?

We tried spawning a [task](https://docs.rs/tokio/latest/tokio/task/) for each but there are some challenges propagating the errors (handling Result), so we are open to any suggestion here 😄

src/main.rs
```diff
@@ -235,7 +238,21 @@ async fn run_server(addr: SocketAddr) -> anyhow::Result<()> {
     let pool = db::ClientPool::new();
     db::run_migrations(&*pool.get().await)
         .await
-        .context("database migrations")?;
+        .context("database migrations")?;  
+
+    // spawning a background task that will run the scheduled jobs
+    // every JOB_PROCESSING_CADENCE_IN_SECS
+    task::spawn(async move {
```
mcass19, line 245: Concerned about what would happen if the task dies. We should implement a logic here to recover from that? If so, do you have any suggestions? 

(we think a posible solution is match against what [JoinHandle](https://doc.rust-lang.org/nightly/std/thread/struct.JoinHandle.html) would have)

apiraino (commented):
left a few comments after a cursory look

src/db.rs
```diff
@@ -215,5 +263,26 @@ CREATE TABLE issue_data (
...
 );
+",
+    "
+CREATE TYPE frequency_unit AS ENUM ('days', 'hours', 'minutes', 'seconds');
+",
+    "
+CREATE TABLE jobs (
+    id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
+    name TEXT NOT NULL,
+    expected_time TIMESTAMP WITH TIME ZONE NOT NULL,
```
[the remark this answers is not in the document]
apiraino: an alternate way I use to supply for that missing function is `md5(random()::text || clock_timestamp()::text)::uuid;`

src/db.rs
```diff
@@ -215,5 +263,26 @@ CREATE TABLE issue_data (
     data JSONB,
     PRIMARY KEY (repo, issue_number, key)
 );
+",
+    "
+CREATE TYPE frequency_unit AS ENUM ('days', 'hours', 'minutes', 'seconds');
+",
+    "
+CREATE TABLE jobs (
+    id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
```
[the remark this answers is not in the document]
apiraino: an alternate way I use to supply for that missing function is `md5(random()::text || clock_timestamp()::text)::uuid;`

apiraino:
As a general thought, I'm curious if the requirements for these tasks exclude _a priori_ an external library that implements deferred job scheduling (instead of writing our own).

src/db.rs
Mark-Simulacrum: This is fine. We shouldn't really have long running tasks, so awaiting inline seems fine to me.

src/db.rs
```diff
@@ -179,6 +182,51 @@ pub async fn run_migrations(client: &DbClient) -> anyhow::Result<()> {
...
+    let jobs = get_jobs_to_execute(&db).await.unwrap();
+    println!("jobs to execute: {:#?}", jobs);
+    tracing::trace!("jobs to execute: {:#?}", jobs);
+
+    for job in jobs.iter() {
+        update_job_executed_at(&db, &job.id).await?;
+
+        match handle_job(&job.name, &job.metadata).await {
+            Ok(_) => {
+                println!("job succesfully executed (id={})", job.id);
```
Mark-Simulacrum, line 195: We'll want to remove non-tracing printing. Is there a reason you added it? It should be possible to test all this out with RUST_LOG=triagebot=trace, for example.

src/db.rs
```diff
@@ -179,6 +182,51 @@ pub async fn run_migrations(client: &DbClient) -> anyhow::Result<()> {
...
+                tracing::trace!("job succesfully executed (id={})", job.id);
+
+                if let Some(frequency) = job.frequency {
```
Mark-Simulacrum, line 198: I think the front reinsert should probably be driven regardless of failure, and we'll just want to be careful about having too frequent cron tasks. Otherwise we turn transient failure into a permanent one.

src/db.rs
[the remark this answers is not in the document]
Mark-Simulacrum: I believe we have postgres 14, so this should be fine.

src/db.rs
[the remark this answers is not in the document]
Mark-Simulacrum: I think something like "execute_at" is better, yeah.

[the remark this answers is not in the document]
Mark-Simulacrum: Oh, we already have executed_at. Hm. Maybe scheduled_at?

I feel like _time makes me think it's a duration of some sort.

src/main.rs
Mark-Simulacrum: I think you can do that with a tokio task just as well? Just await it's completion and re-launch it if it fails?

Mark-Simulacrum:
re:external library, I think it's possible but I'm not convinced this problem is significant/large enough to warrant it and this way we can more closely do just what we need.

I did look at offloading the scheduling elsewhere (e.g., eventbridge), but it looks like there's not a great mechanism there for a lot of tasks (some relatively small limits) and no native support for one-off tasks.

src/db.rs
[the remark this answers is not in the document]
apiraino: FWIW I personally use the suffix `_ts` to express the timestamp of something happening in that very moment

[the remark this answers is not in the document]
mcass19: `scheduled_at` sounds good, thanks 👍🏼

src/db.rs
mcass19: Good catch, need to remove that. Thanks!

src/db.rs
mcass19: Yeah makes sense, will move that outside the successful case. Thanks !

src/main.rs
mcass19: Implemented a possible solution in https://github.com/rust-lang/triagebot/pull/1658/commits/2452683518a71627e685931a7de7953b5505601d.

We have the doubt if that's the best approach since we want to always re spawn the task in case of failure. Open to any other suggestion

src/main.rs
```diff
@@ -6,12 +6,15 @@ use futures::StreamExt;
...
 use route_recognizer::Router;
-use std::{env, net::SocketAddr, sync::Arc};
+use std::{env, net::SocketAddr, sync::Arc, time::Duration};
+use tokio::{task, time::sleep};
 use tower::{Service, ServiceExt};
 use tracing as log;
 use tracing::Instrument;
 use triagebot::{db, github, handlers::Context, notification_listing, payload, EventName};
 
+const JOB_PROCESSING_CADENCE_IN_SECS: u64 = 60;
```
apiraino, line 16: At some point in the future I'd like to take `JOB_PROCESSING_CADENCE_IN_SECS` out into an env variable with a default (I tend to keep configuration separated from the code) but it's really unimportant right now.

src/db/jobs.rs
```diff
@@ -0,0 +1,134 @@
...
+use serde::{Deserialize, Serialize};
+use tokio_postgres::Client as DbClient;
+use uuid::Uuid;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Job {
+    pub id: Uuid,
+    pub name: String,
+    pub scheduled_at: DateTime<FixedOffset>,
+    pub frequency: Option<i32>,
```
Mark-Simulacrum, line 14: It seems like this should be a u32, i.e., we should never have a *negative* duration here.

Mark-Simulacrum: (But actually see my other comment; this field is probably just going to go away entirely).

src/main.rs
```diff
@@ -237,6 +240,36 @@ async fn run_server(addr: SocketAddr) -> anyhow::Result<()> {
...
+                }
+            });
+
+            match res.await {
+                Err(err) if err.is_panic() => {
+                    /* handle panic in above task, re-launching */
+                    tracing::trace!("run_scheduled_jobs task died (error={})", err);
+                }
+                _ => {
+                    /* break in other case by default */
```
Mark-Simulacrum, line 266: This branch should be `unreachable!()`, I think -- the task we're spawning can't actually exit.

src/db.rs
```diff
@@ -179,6 +182,46 @@ pub async fn run_migrations(client: &DbClient) -> anyhow::Result<()> {
...
 
+pub async fn run_scheduled_jobs(db: &DbClient) -> anyhow::Result<()> {
+    let jobs = get_jobs_to_execute(&db).await.unwrap();
+    tracing::trace!("jobs to execute: {:#?}", jobs);
+
+    for job in jobs.iter() {
+        update_job_executed_at(&db, &job.id).await?;
+
+        if let Some(frequency) = job.frequency {
+            let duration = get_duration_from_cron(frequency, job.frequency_unit.as_ref().unwrap());
```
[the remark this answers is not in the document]
mcass19: Sure, I agree it applies much better to the concept of cron  we want.

We'll code something around those lines you mention and see how it goes, thanks!!!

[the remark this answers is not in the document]
mcass19: Implemented a solution in 44a4b9766b5b6fc24aec362de4c5a3936a8ccc64.

As you suggested:
- Use Schedule from cron
- Schedule the jobs in a separate loop from the execution loop
- Make db do not know about cron tasks, they live hardcoded in triagebot for now

src/main.rs
```diff
@@ -240,20 +239,49 @@ async fn run_server(addr: SocketAddr) -> anyhow::Result<()> {
         .await
         .context("database migrations")?;
 
+    // spawning a background task that will schedule the jobs
+    // every JOB_SCHEDULING_CADENCE_IN_SECS
+    task::spawn(async move {
+        loop {
```
mcass19, line 245: This "loop with retry" is almost exactly the same as the one for processing the jobs, changing only the cadence and the "callback" that is called (schedule_jobs or run_scheduled_jobs).

We wanted to abstract this and create a `loop_with_retry` function that receives a cadence and a callback function to avoid the duplication. Do you think it's a good idea? We haven't managed to get it working clear for now, but we'll keep trying if it makes sense

src/db.rs
```diff
@@ -179,6 +183,43 @@ pub async fn run_migrations(client: &DbClient) -> anyhow::Result<()> {
...
+pub async fn run_scheduled_jobs(db: &DbClient) -> anyhow::Result<()> {
+    let jobs = get_jobs_to_execute(&db).await.unwrap();
+    tracing::trace!("jobs to execute: {:#?}", jobs);
+
+    for job in jobs.iter() {
+        update_job_executed_at(&db, &job.id).await?;
+
+        match handle_job(&job.name, &job.metadata).await {
+            Ok(_) => {
+                tracing::trace!("job succesfully executed (id={})", job.id);
```
Mark-Simulacrum, line 210:
```suggestion
                tracing::trace!("job successfully executed (id={})", job.id);
```

src/db/jobs.rs
```diff
@@ -0,0 +1,137 @@
...
+    let mut data = Vec::with_capacity(jobs.len());
+    for job in jobs {
+        let serialized_job = serialize_job(&job);
+        data.push(serialized_job.unwrap());
+    }
+
+    Ok(data)
+}
+
+fn serialize_job(row: &tokio_postgres::row::Row) -> Result<Job> {
```
Mark-Simulacrum, line 121:
```suggestion
fn deserialize_job(row: &tokio_postgres::row::Row) -> Result<Job> {
```

I typically think of serializing as going *to* some format that's not Rust, whereas here we're doing the reverse.

src/jobs.rs
```diff
@@ -0,0 +1,49 @@
...
+//!
+//! and include it in the below vector in jobs():
+//!
+//!   jobs.push(new_job);
+//!
+//! ... fianlly, add the corresponding "send_zulip_message" handler in src/handlers/jobs.rs
+
+use crate::db::jobs::JobSchedule;
+
+// Cadence in seconds with which the jobs will be scheduled
```
Mark-Simulacrum, line 38:
```suggestion
// How often new cron-based jobs will be placed in the queue.
// This is the minimum period *between* a single cron task's executions.
```

src/jobs.rs
```diff
@@ -0,0 +1,49 @@
...
+pub const JOB_SCHEDULING_CADENCE_IN_SECS: u64 = 1800;
+
+// Cadence in seconds with which the jobs will be processed
```
Mark-Simulacrum, line 41:
```suggestion
// How often the database is inspected for jobs which need to execute. This is the granularity at which events will occur.
```

src/main.rs
Mark-Simulacrum: I wouldn't worry about it.

Mark-Simulacrum:
Essentially just nits and I think this is ready to go -- thanks for working on it!

src/db/jobs.rs
mcass19: Yeah I think the same, good catch thank you !

src/jobs.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use crate::db::jobs::JobSchedule;
+
+// How often new cron-based jobs will be placed in the queue.
+// This is the minimum period *between* a single cron task's executions.
+pub const JOB_SCHEDULING_CADENCE_IN_SECS: u64 = 1800;
+
+// How often the database is inspected for jobs which need to execute.
+// This is the granularity at which events will occur.
+pub const JOB_PROCESSING_CADENCE_IN_SECS: u64 = 60;
```
apiraino, line 44: what happens if a job (for any reason) end up taking longer than `JOB_PROCESSING_CADENCE_IN_SECS`?

src/db/jobs.rs
apiraino: Would it make sense to leverage Serde for this task? Something along the line of
```
impl TryFrom<postgres::Row> for Job {
    fn try_from(row: postgres::Row) -> Self {
        Self {
            id: Uuid = row.try_get(0)?;
            name: String = row.try_get(1)?;
            scheduled_at: DateTime<Utc> = row.try_get(2)?;
            // ...
        }
    }
}

// and then:
let j = Job::try_from(row);
```

(code is off the top of my head, not tested and contains bugs :bug: )

src/jobs.rs
julmontesdeoca: Then that could cause other scheduled jobs (if any for the next interval) to be executed later than expected. 
We did worry about this earlier and thought about processing each job asynchronously,  but got a comment from @Mark-Simulacrum (somewhere above) saying that:
> This is fine. We shouldn't really have long-running tasks, so awaiting inline seems fine to me.

perhaps we could run `db::schedule_jobs` async (from a tokio task) inside the loop to not potentially block/delay the next interval?

Mark-Simulacrum: Yeah, I wouldn't worry about this. In an ideal world we'd be emitting a metric here for each task's duration and alarming on it if it hits >5 seconds, say, but we're not really set up for that today so for now we'll just leave this as-is.

I wouldn't expect us to have more than one task in any given job execution point essentially ever, so this is pretty unlikely to become a problem. And most tasks will be "post a comment", which takes ~milliseconds.

apiraino: ok thanks for rebulding the discussion for me! I had lost it in the web interface of github :)

src/db/jobs.rs
Mark-Simulacrum: I don't see how that's leveraging serde.

But I don't think this benefits from a TryFrom impl (in fact I think it's generally a bad idea to use trait impls if you're not using them in a bound).

apiraino: Sorry I mentioned serde in a previous version of the comment but then corrected to "impl a TryFrom" - (which is what I meant). I'll move to Zulip for a follow-up on your comment (I think I need more input).

Mark-Simulacrum (approved):

PR #1664, opened 2022-10-28
Add options to silence mentions message when a particular assignment is used

WaffleLapkin:
I'm not sure if this is the right way to do it, but I think this should work.

I want to remove these long comments about `libs-api` team, when `r? libs-api` is used:

The idea is to add
```toml
[assign.label]
libs-api = ["T-libs-api"]
```

To the Rust `triagebot.toml`, which will add `T-libs-api` label when `r? libs-api` is used. Then `T-libs` won't be added because there is already a `T-*` label.

Then, to silence the message add

```toml
# [mentions."library"]
exclude_labels = ["T-libs-api"]
```

Again, the "solution" and "implementation" seem somewhat hacky. I'm open to suggestions how to better do this.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -90,6 +90,10 @@ pub(crate) struct AssignConfig {
     /// usernames, team names, or ad-hoc groups.
     #[serde(default)]
     pub(crate) owners: HashMap<String, Vec<String>>,
+    /// Labels to be added to the issue when a particular assignment is made.
+    /// Key is asignee, value is a vector of labels to be added.
+    #[serde(default)]
+    pub(crate) label: HashMap<String, Vec<String>>,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
--- a/src/config.rs
+++ b/src/config.rs
@@ -119,6 +119,8 @@ pub(crate) struct MentionsPathConfig {
     pub(crate) message: Option<String>,
     #[serde(default)]
     pub(crate) cc: Vec<String>,
+    #[serde(default)]
+    pub(crate) exclude_labels: Vec<String>,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -19,7 +19,7 @@
 
 use crate::{
     config::AssignConfig,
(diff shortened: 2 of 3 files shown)


Mark-Simulacrum:
I'm not sure I fully understand the motivation here. It seems to me like the message is generally not a big deal? It also seems like hiding it based on something obscure like picking a particular way to roll the reviewer isn't all that appealing to me.

I think I could be convinced that we should make the auto-comment message sometimes conditional on whether we consider a contributor "new" - but even there, it feels like there's a pretty high likelihood it won't trigger when we do want it to.

Overall I think my annoyance I guess with it being posted is pretty low on any given PR, so that makes it very hard to justify custom logic here.

WaffleLapkin:
Maybe this is just a "me problem", which is good, cause we don't need to do anything then :)

> It also seems like hiding it based on something obscure like picking a particular way to roll the reviewer isn't all that appealing to me.

I don't think it's obscure, if `r? libs-api` is used, then we know that this is a PR for `T-libs-api`, not `T-libs`, so we can apply the right label, at which point there is no reason to show the message.

I'm mostly annoyed that there is no way to make a `T-libs-api` PR from the beginning. IIRC `@rustbot label -T-libs +T-libs-api` doesn't work in the PR description, the labels after that are wrong (don't recall in which way though).

> I think I could be convinced that we should make the auto-comment message sometimes conditional on whether we consider a contributor "new" - but even there, it feels like there's a pretty high likelihood it won't trigger when we do want it to.

I actually don't think it's a good idea. Even if the contributor is not new (how new?) they could forget about libs/-api difference or it may be the first library PR, even though they contributed to the compiler a lot...

Mark-Simulacrum:
Hiding specifically libs-api message because you've explicitly selected libs-api does seem okay, but many of our other messages actively ping people - which we probably don't want to silence with a label.

I think in general this kind of thing is a bit hard with the current architecture, as mentions and reviewer assignment happen in different steps and ordering also isn't guaranteed (or whether they'll see a label set by a previous step).

I think the general desire to make things quieter when they're likely not helpful is a good one, but I don't think this is quite the right approach (not that I necessarily have better suggestions though).

WaffleLapkin:
I don't need/want this anymore, and it doesn't looks like anyone else does. Thus — closing.

PR #1667, opened 2022-11-01
Allow author to self-assign with r?

ehuss:
This removes the restriction that the PR author cannot self-assign with `r?`. This was an unintentional difference from highfive, and there isn't a real good reason to prevent it.


rylev (approved):
🎉

PR #1668, opened 2022-11-01
Provide a better error message when r? fails

ehuss:
Currently when `r?` fails when there are no candidates, it provides a rather obtuse error message. This PR enhances it so that the error message has a little more detail. These errors can happen with a misconfiguration (like a cyclical group), or when the PR author (or current assignee) is the only user in a group.

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -577,7 +577,7 @@ impl fmt::Display for FindReviewerError {
             FindReviewerError::NoReviewer { initial } => {
                 write!(
                     f,
-                    "No reviewers could be found from initial choice `{}`\n\
+                    "No reviewers could be found from initial request `{}`\n\
                      This repo may be misconfigured.\n\
                      Use r? to specify someone else to assign.",
                     initial.join(",")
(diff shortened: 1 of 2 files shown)


src/handlers/assign.rs
```diff
@@ -536,32 +539,61 @@ pub(super) async fn handle_command(
...
+                    "Team or group `{team}` not found.\n\
+                    \n\
+                    rust-lang team names can be found at https://github.com/rust-lang/team/tree/master/teams.\n\
+                    Reviewer group names can be found in `triagebot.toml` in this repo."
+                )
+            }
+            FindReviewerError::NoReviewer { initial } => {
+                write!(
+                    f,
+                    "No reviewers could be found from initial choice `{}`\n\
```
rylev, line 580: I personally find the term "initial choice" confusing here. Is there a way to make it more clear what that means? Perhaps: "No reviewers could be found from initial request `{}`"

ehuss: Yea, I struggled with trying to write something that anyone would understand without making the error handling too complicated (this PR is already a bit larger than I wanted). Part of the issue is that there are two ways those "initial choices" come in, either by an explicit `r?` or via auto-detection from the diff and owners table. I didn't want to carry that information around just to make this error better.

Hopefully nobody ever sees this message. I think it should only happen if there is a misconfiguration.

I applied your suggestion, but if you have other ideas I'd be happy to add them.

Mark-Simulacrum (approved):

PR #1672, opened 2022-11-13
Don't ping everyone on rebase mistakes

JakobDegen:
See Zulip discussion [here](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/I.20.28accidentally.29.20pinged.20most.20of.20the.20Rust.20org). Instead of pinging everyone, we'll send a helpful message with advice on how to correctly rebase.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -105,7 +105,9 @@ pub(crate) struct NoteConfig {
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
 pub(crate) struct MentionsConfig {
+    pub(crate) bors_commit_message: Option<String>,
     #[serde(flatten)]
     pub(crate) paths: HashMap<String, MentionsPathConfig>,
 }
@@ -397,5 +399,47 @@ mod tests {
                 no_merges: None,
             }
         );
+    }
+
+    #[test]
+    fn mentions_config() {
+        let config = r#"
+        [mentions."some_other_path"]
+        message = "foo"
+        cc = ["@someone"]
+
+        [mentions]
+        bors-commit-message = "has bors commit"
+        "#;
+        let config = toml::from_str::<Config>(config).unwrap().mentions.unwrap();
+        assert!(config.paths.len() == 1);
+        assert_eq!(
+            config.bors_commit_message.as_deref(),
+            Some("has bors commit")
+        );
+
+        let config = r#"
+        [mentions]
+        bors-commit-message = "has bors commit"
+
+        [mentions."some_other_path"]
+        message = "foo"
+        cc = ["@someone"]
+        "#;
+        let config = toml::from_str::<Config>(config).unwrap().mentions.unwrap();
+        assert!(config.paths.len() == 1);
+        assert_eq!(
+            config.bors_commit_message.as_deref(),
+            Some("has bors commit")
+        );
+
+        let config = r#"
+        [mentions."some_other_path"]
+        message = "foo"
+        cc = ["@someone"]
+        "#;
+        let config = toml::from_str::<Config>(config).unwrap().mentions.unwrap();
+        assert!(config.paths.len() == 1);
+        assert_eq!(config.bors_commit_message, None);
     }
 }
--- a/src/handlers/mentions.rs
+++ b/src/handlers/mentions.rs
@@ -17,6 +17,7 @@ use tracing as log;
 const MENTIONS_KEY: &str = "mentions";
 
 pub(super) struct MentionsInput {
+    has_bors_commit: bool,
     paths: Vec<String>,
 }
 
@@ -50,6 +51,17 @@ pub(super) async fn parse_input(
         return Ok(None);
     }
 
+    let has_bors_commit = event.action == IssuesAction::Opened
+        && config.bors_commit_message.is_some()
+        && event
+            .issue
+            .commits(&ctx.github)
+            .await
+            .map_err(|e| log::error!("failed to fetch commits: {:?}", e))
+            .unwrap_or_default()
+            .into_iter()
+            .any(|commit| commit.commit.author.name == "bors");
+
     if let Some(diff) = event
         .issue
         .diff(&ctx.github)
@@ -78,7 +90,10 @@ pub(super) async fn parse_input(
             .map(|(key, _mention)| key.to_string())
             .collect();
         if !to_mention.is_empty() {
-            return Ok(Some(MentionsInput { paths: to_mention }));
+            return Ok(Some(MentionsInput {
+                has_bors_commit,
+                paths: to_mention,
+            }));
         }
     }
     Ok(None)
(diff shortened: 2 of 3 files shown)


src/config.rs
```diff
@@ -105,7 +105,10 @@ pub(crate) struct NoteConfig {
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
 pub(crate) struct MentionsConfig {
+    #[serde(default)]
```
Mark-Simulacrum, line 110: This should be an optional field; we don't want to default to an empty string.

src/handlers/mentions.rs
```diff
@@ -95,23 +109,30 @@ pub(super) async fn handle_input(
...
-        let MentionsPathConfig { message, cc } = &config.paths[to_mention];
-        if !result.is_empty() {
-            result.push_str("\n\n");
-        }
-        match message {
-            Some(m) => result.push_str(m),
-            None => write!(result, "Some changes occurred in {to_mention}").unwrap(),
+    match input {
+        MentionsInput::HasBorsCommit => {
+            result = config.bors_commit_message.to_string();
```
Mark-Simulacrum, line 114: I think this doesn't provide a way to opt-out and get the real message (e.g., if you are pulling in miri or clippy, and making changes as part of the merge to other files).

I think one option is to escape all of the mentions and still post the message - that gives the author the ability to re-ping the right folks. Though only if they're a member of the org for teams.

I think a better option is to detect the bors merge commits and then set the PR as draft and bail out. (Probably leaving a comment is still reasonable). That means that the author can just close the PR if they don't want to ping folks or fix it and then mark as ready (which should already retrigger the logic if I'm remembering right).

JakobDegen: That seems quite annoying for the cases in which this is intentional - the current version will warn but at least it doesn't require the PR author to take any additional steps if that's what they intended to do

src/handlers/mentions.rs
```diff
@@ -50,6 +51,19 @@ pub(super) async fn parse_input(
...
+    if event
+        .issue
+        .commits(&ctx.github)
+        .await
+        .map_err(|e| log::error!("failed to fetch commits: {:?}", e))
+        .unwrap_or_default()
+        .into_iter()
+        .any(|commit| commit.commit.author.name == "bors")
+    {
+        return Ok(Some(MentionsInput::HasBorsCommit));
```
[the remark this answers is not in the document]
JakobDegen: Yeah, seems like this isn't the most unreasonable behavior for that case though. The PR author could of course ping manually, but I also don't think that it's even necessary in case of a subtree bump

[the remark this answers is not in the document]
Mark-Simulacrum: It does seem just as relevant to ping on a subtree bump if it touches code outside the subtree (as is not uncommon, I believe, though maybe not particularly desirable).

src/handlers/mentions.rs
Mark-Simulacrum: The additional step is clicking in the UI on ready for review - that feels pretty lightweight to me? It's also not strictly required - e.g., bors will listen to you in draft state iirc.

src/handlers/mentions.rs
```diff
@@ -95,23 +109,30 @@ pub(super) async fn handle_input(
...
-        let MentionsPathConfig { message, cc } = &config.paths[to_mention];
-        if !result.is_empty() {
-            result.push_str("\n\n");
-        }
-        match message {
-            Some(m) => result.push_str(m),
-            None => write!(result, "Some changes occurred in {to_mention}").unwrap(),
+    match input {
+        MentionsInput::HasBorsCommit => {
+            result = config.bors_commit_message.clone().unwrap_or_default();
```
Mark-Simulacrum, line 114: If the message isn't configured, we shouldn't post an empty comment/skip posting, but rather not have this logic at all.

src/handlers/mentions.rs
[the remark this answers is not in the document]
JakobDegen: Oh, I did not know that was possible. How does that happen?

I'll update in that case

src/handlers/mentions.rs
JakobDegen: Hmm, looking at the github API, I don't think we can mark someone else's pull request as being a draft. Escaping the pings is possible though, I'll update

src/handlers/mentions.rs
```diff
@@ -109,6 +132,11 @@ pub(super) async fn handle_input(
             None => write!(result, "Some changes occurred in {to_mention}").unwrap(),
         }
         if !cc.is_empty() {
+            let cc: Vec<String> = if input.has_bors_commit {
+                cc.iter().map(|s| format!("`{s}`")).collect()
```
RalfJung, line 136: On Zulip there was a proposal to also mark the PR as draft, to prevent it from being merge.

Ideally we'd want a scheme where the pings definitely happen when the PR gets approved or lands, but I am not sure if that is possible.

src/handlers/mentions.rs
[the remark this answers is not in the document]
RalfJung: https://github.com/rust-lang/rust/issues/101907#issuecomment-1287936891 would also help here -- it would mean we could distinguish between Miri-bors commits and rustc-bors commits.

src/handlers/mentions.rs
Mark-Simulacrum: It seems like the conversion is possible, just needs graphql rather than the rest API. 

https://docs.github.com/en/graphql/reference/mutations#convertpullrequesttodraft

src/handlers/mentions.rs
```diff
@@ -50,6 +51,17 @@ pub(super) async fn parse_input(
...
+    let has_bors_commit = event.action == IssuesAction::Opened
+        && config.bors_commit_message.is_some()
+        && event
+            .issue
+            .commits(&ctx.github)
+            .await
+            .map_err(|e| log::error!("failed to fetch commits: {:?}", e))
+            .unwrap_or_default()
+            .into_iter()
+            .any(|commit| commit.commit.author.name == "bors");
```
RalfJung, line 63: Does this even work for problematic rebases? As someone pointed out on Zulip, https://github.com/rust-lang/rust/pull/103918/commits does not have any bors-authored commits.

src/config.rs
```diff
@@ -398,4 +400,46 @@ mod tests {
...
+
+        [mentions."some_other_path"]
+        message = "foo"
+        cc = ["@someone"]
+        "#;
+        let config = toml::from_str::<Config>(config).unwrap().mentions.unwrap();
+        assert!(config.paths.len() == 1);
+        assert_eq!(
+            config.bors_commit_message.as_deref(),
+            Some("has bors commit")
```
riking, line 433: How could this middle test possibly fail?

This test should be over the overall output, not the config system

Urgau:
This PR pretty outdated now. If someone would like to pick this up feel free (or if you want to continue it Jakob).

PR #1701, opened 2023-06-11
Log failure of agenda calls

Mark-Simulacrum:


ehuss (approved):
Logging errors always seems like a good thing to me. 😄

PR #1704, opened 2023-06-14
add more config options for `no_merges`

pitaj:
- option for excluding PRs with certain labels from consideration
- option for adding labels when merge commits are detected
- option for setting a custom comment message for when merge commits are detected

rust-forge PR: https://github.com/rust-lang/rust-forge/pull/691

### Motivation

Occasionally, a merge commit like rust-lang/rust@cb5c011670ce8d073d0aae8c45e73c20593bfa11 makes it past manual review and gets merged into master.

At one point, we tried adding a check to CI to prevent this from happening (rust-lang/rust#105058), but that ended up [causing issues](https://github.com/rust-lang/rust/pull/106319#issuecomment-1368144076) and was [reverted](https://github.com/rust-lang/rust/pull/106320). This kind of check is simply too fragile for CI, and there must be a way for a human to override the bot's decision.

The capability to detect and warn about merge commits has been present in triagebot for quite some time, but was never enabled at rust-lang/rust, possibly due to concerns about false positives on rollup and subtree/submodule sync PRs. This PR intends to alleviate those concerns.

This will allow us to use the following options at rust-lang/rust:

```toml
[no_merges]
exclude_labels = ["rollup", "sync"] # hypothetical label for PRs that sync subtrees or submodules
labels = ["has-merge-commits", "S-waiting-on-author"]
```

Which will exclude rollup PRs and subtree/submodule sync PRs from merge commit detection, and post the default warning message and add the `has-merge-commits` and `S-waiting-on-author` labels when merge commits are detected on all other PRs.

The eventual vision is to have bors refuse to merge if the `has-merge-commits` label is present. A reviewer can still force the merge by removing that label if they so wish.

### Previous discussion

- [Zulip topic](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Merge.20commit.20check.20in.20CI)
- https://github.com/rust-lang/rust/pull/105058

### DIFF
--- a/src/handlers/no_merges.rs
+++ b/src/handlers/no_merges.rs
@@ -4,7 +4,7 @@
 use crate::{
     config::NoMergesConfig,
     db::issue_data::IssueData,
-    github::{IssuesAction, IssuesEvent},
+    github::{IssuesAction, IssuesEvent, Label},
     handlers::Context,
 };
 use anyhow::Context as _;
@@ -38,14 +38,21 @@ pub(super) async fn parse_input(
         return Ok(None);
     }
 
-    // Require an empty configuration block to enable no-merges notifications.
-    if config.is_none() {
+    // Require a `[no_merges]` configuration block to enable no-merges notifications.
+    let Some(config) = config else {
         return Ok(None);
-    }
+    };
 
     // Don't ping on rollups or draft PRs.
     if event.issue.title.starts_with("Rollup of") || event.issue.draft {
         return Ok(None);
+    }
+
+    // Don't trigger if the PR has any of the excluded labels.
+    for label in event.issue.labels() {
+        if config.exclude_labels.contains(&label.name) {
+            return Ok(None);
+        }
     }
 
     let mut merge_commits = HashSet::new();
@@ -73,7 +80,7 @@ pub(super) async fn parse_input(
 
 pub(super) async fn handle_input(
     ctx: &Context,
-    _config: &NoMergesConfig,
+    config: &NoMergesConfig,
     event: &IssuesEvent,
     input: NoMergesInput,
 ) -> anyhow::Result<()> {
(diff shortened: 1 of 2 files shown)


tgross35:
Not something to do in this PR, but it would be kind of cool to have merge readiness levels:

- `merge-ok`: 1-2 commits
- `merge-likely`: 3-5 commits
- `merge-iffy`: 5+ commits
- `merge-NEVER`: contains a merge commit

Just to give a quick indicator that a squash may be needed, and tie in with this

Mark-Simulacrum:
Can you say more about the motivation for this? Which PRs are we looking to exclude, how are we planning to use the has-merge-commits label, etc.?

We'll also want to document any features in the Forge (similar to https://github.com/rust-lang/rust-forge/pull/690).

pitaj:
Main motivation is to prevent merge commits like the following from being merged. For now this will just edit labels as a warning but in the future bors might refuse to merge if merge commits are present.

https://github.com/rust-lang/rust/commit/cb5c011670ce8d073d0aae8c45e73c20593bfa11

I'm just putting this here for now so I can find it later. I'll edit the PR body to include a more detailed description later.

pitaj:
Alright, I've edited the PR body with the motivation for this. And I opened a PR to rust-forge.

src/handlers/no_merges.rs
```diff
@@ -102,11 +104,25 @@ pub(super) async fn handle_input(
...
+        .to_string()
+    };
 
+    let since_last_posted = if state.data.mentioned_merge_commits.is_empty() {
+        ""
+    } else {
+        " (since this message was last posted)"
+    };
+    write!(
+        message,
```
ehuss, line 117: A lot of these strings have a lot of whitespace indentation, which ends up causing them to be rendered as code blocks, such as:

        The following commits are merge commits:

Can the strings here use `\n\` at the end of lines to remove that whitespace so that it is better formatted as markdown?

src/handlers/no_merges.rs
```diff
@@ -102,11 +104,25 @@ pub(super) async fn handle_input(
...
+        "
+    
         The following commits are merge commits{since_last_posted}:
+        "
+    )
+    .unwrap();
 
-    "
-    );
+    let mut should_send = false;
```
ehuss, line 125: GitHub won't let me comment on a line outside of the diff, but the `for` loop just below here needs to use `writeln!` instead of `write!`.

ehuss (commented):
Just some minor nits.

src/handlers/no_merges.rs
ehuss: I don't have a strong preference. Another option is to move larger strings into a const like the [assign](https://github.com/rust-lang/triagebot/blob/0cfa95a8f7382d275a536da978f24f8be14d144f/src/handlers/assign.rs#L41-L69) module. You still need to be careful about newlines since GitHub markdown is newline-sensitive.

ehuss (approved):
Thanks! I haven't tested, but I believe it should all work. I'm uncertain if there is risk about false positives, but I imagine we can adapt to them if they occur.

PR #1705, opened 2023-06-16
Process docs-update Zulip trigger in the background.

ehuss:
Zulip has a timeout of about 10 seconds by default for a webhook response. It seems like the docs-update process may take longer, so this changes it to run in the background.

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -21,19 +21,48 @@ pub struct Request {
     token: String,
 }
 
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (approved):

Mark-Simulacrum:
Looks like this needs a rebase.

ehuss:
Thanks, it should be updated.

PR #1710, opened 2023-07-20
Fix docs-update creating merge commits

ehuss:
The docs-update job has been having problems with creating merge commits that are not wanted. I think this was caused by the recent force-pushes to the master branch of rust-lang/rust. I previously thought `/merge-upstream` was required to "pull" the changes from upstream, but in my testing it seems like that is unnecessary. I have heard in the past that GitHub forks are implemented in such a way that forks use the same underlying git object store. Thus, it seems like referring to a commit from upstream seems to "just work" without needing to do a pull.


ehuss:
@Mark-Simulacrum Just checking to see if you'll have a chance to look at this?

Mark-Simulacrum (approved):

PR #1713, opened 2023-07-31
Use GraphQL to determine if a user is new.

### LINKED ISSUE #1689: rustbot always thinks @oli-obk is a new contributor
Reporting this because the bot _also_ failed to ping rust-lang/miri here: https://github.com/rust-lang/rust/pull/109938#issuecomment-1510407389

Examples of the bot thinking @oli-obk is a new contributor:

https://github.com/rust-lang/rust/pull/109938
https://github.com/rust-lang/rust/pull/109115
https://github.com/rust-lang/rust/pull/108860
https://github.com/rust-lang/rust/pull/110315

It looks like maybe all the cases where Oli doesn't get the welcome message are PRs that start with a `r? @ghost` in the PR description.

ehuss:
For some reason the `/repos/ORG/REPO/commits?author=AUTHOR` endpoint hasn't been working to see if a user has posted any commits (I don't know why). This switches it to use GraphQL which seems to be more reliable in my testing. AFAICT, the query costs 1 point, so it shouldn't be too expensive.

I could instead revert 077cdf8b11e0aa10d2cdfad655f14bbb0a4532b9 and use the `/search` endpoint instead. However, that endpoint has some issues, such as not working on forks (mainly a problem for testing), and tight rate limits (30/minute, which is very unlikely to be hit, but I think is very small and could compete with other triagebot handlers).

I decided to not use cynic for this and instead just issue JSON to the endpoint directly. I can build the cynic types, but I'm finding using cynic to be awkward. For one-off queries like this, it generates a dozen types, and it is also a pain to iterate and change the query. If you want to make any changes, you have to find the original query, find and set up the cynic generator, and regenerate the types, and then copy all that stuff back in and adjust all the code using it. Cynic is better if you are actually using the returned values, but in this case there is just one value that is needed in one place.

I'm fine with doing either alternative (use `/search` or use cynic), just let me know. I just find this approach a little more reliable and easier to write and iterate on.

Fixes #1689

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -1918,26 +1918,120 @@ impl GithubClient {
         }
     }
 
+    /// Issues an ad-hoc GraphQL query.
+    pub async fn graphql_query<T: serde::de::DeserializeOwned>(
+        &self,
+        query: &str,
+        vars: serde_json::Value,
+    ) -> anyhow::Result<T> {
+        self.json(
+            self.post(Repository::GITHUB_GRAPHQL_API_URL)
+                .json(&serde_json::json!({
+                    "query": query,
+                    "variables": vars,
+                })),
+        )
+        .await
+    }
+
+    /// Returns the object ID of the given user.
+    ///
+    /// Returns `None` if the user doesn't exist.
+    pub async fn user_object_id(&self, user: &str) -> anyhow::Result<Option<String>> {
+        let user_info: serde_json::Value = self
+            .graphql_query(
+                "query($user:String!) {
+                    user(login:$user) {
+                        id
+                    }
+                }",
+                serde_json::json!({
+                    "user": user,
+                }),
+            )
+            .await?;
+        if let Some(id) = user_info["data"]["user"]["id"].as_str() {
+            return Ok(Some(id.to_string()));
+        }
+        if let Some(errors) = user_info["errors"].as_array() {
+            if errors
+                .iter()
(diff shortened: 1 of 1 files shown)


ehuss:
@Mark-Simulacrum Just checking if you had any questions on this?

Mark-Simulacrum (approved):
Search API limits definitely hit us - e.g., lang agendas are rate limited to be generated twice a minute - so I'd avoid those. Not too familiar with cynic but the code here looks fine so fine with that choice too.

Mark-Simulacrum:
My previous approval is still good here - do we still have the merge limitations, or can you go ahead and merge?

ehuss:
Hm, I'm not entirely sure why I can't merge. It says I am not authorized. I know that GitHub does not allow an author to approve their own PR, but I've had situations on other repos where after someone else approves it, the author is allowed to merge it. But maybe the branch protection rule in this repo is slightly different. Unfortunately GitHub doesn't provide any information to help debug an authorization problem, and I don't have permissions to inspect the permissions in this repo.

Mark-Simulacrum:
Yeah, I remembered since then - we unblocked merges for a bit by manually adding the triagebot team to branch protections, but that broke sync-team since it doesn't know what that entry means. https://github.com/rust-lang/team/blob/master/repos/rust-lang/triagebot.toml could have branch protections dropped potentially...

PR #1715, opened 2023-08-28
Rustfmt

ehuss:
Rustfmt now formats let-else, which is causing CI to fail.


ehuss:
Since let-else is relatively new, I don't think there has been a lot of opportunities for people to use it, so I suspect that's why there aren't a lot of changes.

I think it would be good to merge this in order to fix CI.

As for the question about changing the rustfmt settings, I don't have an opinion. I'm happy either way, but I think it would be good to follow up on that separately so that it doesn't block other PRs.

PR #1716, opened 2023-09-06
Autoadd team label on issue/pr assignment

apiraino:
When we trigger a `r? <teamname>` the rustbot assigns the pr/issue to someone but we don't actually check the team label for that issue/pr. This is useful for tracking its progress when we periodically check the PRs of `$team`. 

This patch should also add the team label when the assignment to a team is requested in a github comment.

Requested by Felix [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Review.20for.20.23114412/near/389232591).

Tested locally ([see here](https://github.com/apiraino/test-triagebot/pull/24#issuecomment-1708598704)).

r? @Mark-Simulacrum or @spastorino

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -487,6 +487,22 @@ pub(super) async fn handle_command(
                     name.to_string()
                 } else {
                     let teams = crate::team_data::teams(&ctx.github).await?;
+                    // Determine if assignee is a team. If yes, add the corresponding label
+                    // team name here is without prefix 't-' (e.g. 'compiler', 'libs', etc.)
+                    if let Some(team) = teams.teams.get(&name) {
+                        let t_label = format!("t-{}", &team.name);
+                        if let Err(err) = issue
+                            .add_labels(&ctx.github, vec![github::Label { name: t_label }])
+                            .await
+                        {
+                            if let Some(github::UnknownLabels { .. }) = err.downcast_ref() {
+                                log::warn!("Error assigning label: {}", err);
+                            } else {
+                                return Err(err);
+                            }
+                        }
+                    }
+
                     match find_reviewer_from_names(&teams, config, issue, &[name]) {
                         Ok(assignee) => assignee,
                         Err(e) => {


src/handlers/assign.rs
```diff
@@ -487,6 +487,15 @@ pub(super) async fn handle_command(
                     name.to_string()
                 } else {
                     let teams = crate::team_data::teams(&ctx.github).await?;
+                    // Determine if assignee is a team. If yes, add the corresponding label
+                    // team name here is without prefix 't-' (e.g. 'compiler', 'libs', etc.)
+                    if let Some(team) = teams.teams.get(&name) {
+                        let _team_label = format!("t-{}", team.clone().name);
+                        issue
+                            .add_labels(&ctx.github, vec![github::Label { name: _team_label }])
```
ehuss, line 495: I think this will fail if the repository does not have a `T-*` label for the given team. Would it be possible to ignore the `UnknownLabels` error (or just log it)?

(Also, not all teams use T-* labels that match their name exactly.)

Stylistically, is there a reason to use a leading underscore?

apiraino: thanks @ehuss for giving a first look!

Yes, this solution is only useful for the main teams (compiler, libs-*, rustdoc). AFAICS, there's no way to infer the github label from the team name, it's not exposed in the [teams.json](https://team-api.infra.rust-lang.org/v1/teams.json) file. Any idea to improve the chances to catch a team label that does not respect the `T-<teamname>` structure (besides fixed mappings).

> Stylistically, is there a reason to use a leading underscore?

old habits? :sweat_smile: changed it (in case you meant it was ugly).

apiraino: > (Also, not all teams use T-* labels that match their name exactly.)

Well, now that I think of it. It wouldn't be a bad idea to map teams+labels in the team.json by retrieving this info from the .toml file. Ex. [the compiler team](https://github.com/rust-lang/team/blob/71719e35d2dd03b1741f2cf8af22f1db707524ce/teams/compiler.toml#L43C12-L43C12) has it. Sort of.

Anyway nothing I'd think about here and now.

apiraino:
Unrelated: another thing I notice in the triagebot is this Github client. Everytime an HTTP request fails [it will be retried a few times](https://github.com/rust-lang/triagebot/blob/a50b73aead327a1f5a168d6c3c91625e6018285e/src/github.rs#L34) before giving up.

In some cases, ex. when I try to set the label `https://api.github.com/repos/apiraino/test-triagebot/labels/t-android` it will obviously never work so we can give up on the first try. Or maybe not retry on some HTTP error, like 404s. Anyway, in some cases it could be slightly smarter and bail out earlier

src/handlers/assign.rs
apiraino: tentatively marking the review as resolved with the latest commit

src/handlers/assign.rs
```diff
@@ -487,6 +487,22 @@ pub(super) async fn handle_command(
                     name.to_string()
                 } else {
                     let teams = crate::team_data::teams(&ctx.github).await?;
+                    // Determine if assignee is a team. If yes, add the corresponding label
+                    // team name here is without prefix 't-' (e.g. 'compiler', 'libs', etc.)
+                    if let Some(team) = teams.teams.get(&name) {
+                        let t_label = format!("t-{}", team.clone().name);
```
tgross35, line 493:
```suggestion
                        let t_label = format!("t-{}", &team.name);
```
Minor nit, unless there's a reason to clone

src/handlers/assign.rs
```diff
@@ -487,6 +487,22 @@ pub(super) async fn handle_command(
...
+                        if let Err(err) = issue
+                            .add_labels(&ctx.github, vec![github::Label { name: t_label }])
+                            .await
+                        {
+                            if err.to_string().starts_with("Unknown") {
```
ehuss, line 498: I feel a little uncomfortable using string-based matching. I think the following might be a little more reliable:

```suggestion
                            if let Some(UnknownLabels { .. }) = err.downcast_ref() {
```

ehuss (approved):
Thanks!

src/handlers/assign.rs
spastorino: Just `team.name` should work btw

PR #1720, opened 2023-09-16
no-merges: match titles instead of labels

pitaj:
also don't re-add labels if they're manually removed

labels are not always set atomically when opening a PR example: rust-lang/miri#3059

Forge change: https://github.com/rust-lang/rust-forge/pull/701

### DIFF
--- a/src/handlers/no_merges.rs
+++ b/src/handlers/no_merges.rs
@@ -48,11 +48,13 @@ pub(super) async fn parse_input(
         return Ok(None);
     }
 
-    // Don't trigger if the PR has any of the excluded labels.
-    for label in event.issue.labels() {
-        if config.exclude_labels.contains(&label.name) {
-            return Ok(None);
-        }
+    // Don't trigger if the PR has any of the excluded title segments.
+    if config
+        .exclude_titles
+        .iter()
+        .any(|s| event.issue.title.contains(s))
+    {
+        return Ok(None);
     }
 
     let mut merge_commits = HashSet::new();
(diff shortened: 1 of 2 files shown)


src/handlers/no_merges.rs
```diff
@@ -48,11 +48,13 @@ pub(super) async fn parse_input(
...
-    // Don't trigger if the PR has any of the excluded labels.
-    for label in event.issue.labels() {
-        if config.exclude_labels.contains(&label.name) {
-            return Ok(None);
-        }
+    // Don't trigger if the PR has any of the excluded title segments.
+    if config
+        .exclude_titles
+        .iter()
+        .any(|s| event.issue.title.contains(s))
```
ehuss, line 55: Any particular reason to not make this case-insensitive? Seems like making it insensitive could make it a little less likely to have false-positives.

pitaj: Obviously it makes the implementation more complicated, but the real reason is that it reduces specificity.

Capitalization can be used to disambiguate. For example, on `rust-clippy`, using the capitalized "Rustup" disambiguates between [PRs updating from `rust`](https://github.com/rust-lang/rust-clippy/pull/11469) and ones [just mentioning "rustup"](https://github.com/rust-lang/rust-clippy/pull/7239)

Since only the first letter in a word is capitalized, generally it only takes a couple more entries to exclude the various combinations if so desired.

src/handlers/no_merges.rs
```diff
@@ -132,6 +134,22 @@ pub(super) async fn handle_input(
...
 
     if should_send {
+        if !first_time {
+            // Check if the labels are already set.
+            // Otherwise, they were probably removed manually.
+            if !event
+                .issue
+                .labels()
+                .iter()
+                .all(|label| config.labels.contains(&label.name))
```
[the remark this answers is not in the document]
pitaj: Good catch. Yeah the logic here does look wrong. What I meant to have is "if any labels from the config were removed" but it's currently turned around and upside down. Will fix.

ehuss (approved):

PR #1727, opened 2023-09-25
Learn to parse (and clean) team labels with a "t-" prefix

apiraino:
Improve a little bit the team label parser. Accept and fix wrong labels like `t-*` or `T-*` (sometimes we mix up Zulip stream names and actual team names).

I'd _not_ document this in the [triagebot wiki](https://forge.rust-lang.org/triagebot/pr-assignment.html#usage) because I'd prefer not to encourage using `t-*` labels. This patch will be our little secret 😇

Mentioned [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/Review.20for.20.23114412/near/389946378).

r? @ehuss 

thanks!

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -487,10 +487,11 @@ pub(super) async fn handle_command(
                     name.to_string()
                 } else {
                     let teams = crate::team_data::teams(&ctx.github).await?;
-                    // Determine if assignee is a team. If yes, add the corresponding label
-                    // team name here is without prefix 't-' (e.g. 'compiler', 'libs', etc.)
-                    if let Some(team) = teams.teams.get(&name) {
-                        let t_label = format!("t-{}", &team.name);
+                    // remove "t-" or "T-" prefixes before checking if it's a team name
+                    let team_name = name.trim_start_matches("t-").trim_start_matches("T-");
+                    // Determine if assignee is a team. If yes, add the corresponding GH label.
+                    if teams.teams.get(team_name).is_some() {
+                        let t_label = format!("T-{}", &team_name);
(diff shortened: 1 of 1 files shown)


src/handlers/assign.rs
```diff
@@ -487,10 +487,20 @@ pub(super) async fn handle_command(
                     name.to_string()
                 } else {
                     let teams = crate::team_data::teams(&ctx.github).await?;
-                    // Determine if assignee is a team. If yes, add the corresponding label
-                    // team name here is without prefix 't-' (e.g. 'compiler', 'libs', etc.)
-                    if let Some(team) = teams.teams.get(&name) {
-                        let t_label = format!("t-{}", &team.name);
+                    // remove "t-" or "T-" prefixes before checking if it's a team name
```
Mark-Simulacrum, line 490: Can we just trim T- and t- instead of this complicated logic?

apiraino: good point, I should have thought about it.

ehuss (approved):

ehuss:
Just being nit-picky, but it seems like this will fail if there is a legitimate user with a `t-` prefix like `t-eric`? It can be fixed if there is ever a user like that, just wanted to point it out.

apiraino:
> Just being nit-picky, but it seems like this will fail if there is a legitimate user with a `t-` prefix like `t-eric`?

Or ever a user with github handle `t-` (which exists :slightly_smiling_face:). Anyway, totally fair point and in that case the team name check must be shortcircuited before by another check if the provided full string being examined _is_ actually a team member

PR #1730, opened 2023-10-08
feat: validate proposed change(s) in triagebot.toml for PRs ✨

meysam81:
fixes rust-lang/rust#106104

A successful run is visible here: https://github.com/meysam81/triagebot-test/pull/2#issuecomment-1751986969

Some highlights I need to mention here:

- I can't be sure if the proposed change is placed in the right spot, but it works! :shrug:
- If it's not too much to ask, would you please add `hacktoberfest-accepted` to this PR, since the repo is not marked with `hacktoberfest`.
- One question of mine is: why is `cargo fmt`, nor `cargo clippy` not formatting my code? :thinking: 
- How can I write test for this? I couldn't figure this part out!

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -6,7 +6,7 @@ use std::sync::{Arc, RwLock};
 use std::time::{Duration, Instant};
 use tracing as log;
 
-static CONFIG_FILE_NAME: &str = "triagebot.toml";
+pub(crate) static CONFIG_FILE_NAME: &str = "triagebot.toml";
 const REFRESH_EVERY: Duration = Duration::from_secs(2 * 60); // Every two minutes
 
 lazy_static::lazy_static! {
@@ -36,6 +36,9 @@ pub(crate) struct Config {
     pub(crate) note: Option<NoteConfig>,
     pub(crate) mentions: Option<MentionsConfig>,
     pub(crate) no_merges: Option<NoMergesConfig>,
+    // We want this validation to run even without the entry in the config file
+    #[serde(default = "ValidateConfig::default")]
+    pub(crate) validate_config: Option<ValidateConfig>,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -165,6 +168,15 @@ pub(crate) struct ShortcutConfig {
 #[serde(deny_unknown_fields)]
 pub(crate) struct PrioritizeConfig {
     pub(crate) label: String,
+}
+
+#[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+pub(crate) struct ValidateConfig {}
+
+impl ValidateConfig {
+    fn default() -> Option<Self> {
+        Some(ValidateConfig {})
+    }
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -450,6 +462,7 @@ mod tests {
                 review_requested: None,
                 mentions: None,
                 no_merges: None,
+                validate_config: Some(ValidateConfig {}),
             }
         );
     }
--- a/src/config.rs
+++ b/src/config.rs
@@ -17,6 +17,7 @@ lazy_static::lazy_static! {
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
 #[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
 pub(crate) struct Config {
     pub(crate) relabel: Option<RelabelConfig>,
     pub(crate) assign: Option<AssignConfig>,
@@ -38,6 +39,7 @@ pub(crate) struct Config {
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
 pub(crate) struct NominateConfig {
     // team name -> label
     pub(crate) teams: HashMap<String, String>,
@@ -68,6 +70,7 @@ impl PingConfig {
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
 pub(crate) struct PingTeamConfig {
     pub(crate) message: String,
     #[serde(default)]
@@ -76,6 +79,7 @@ pub(crate) struct PingTeamConfig {
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
 pub(crate) struct AssignConfig {
     /// If `true`, then posts a warning comment if the PR is opened against a
     /// different branch than the default (usually master or main).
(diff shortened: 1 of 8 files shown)


meysam81:
r?

meysam81:
r? dev-tools

src/config.rs
```diff
@@ -15,6 +15,10 @@ lazy_static::lazy_static! {
         RwLock::new(HashMap::new());
 }
 
+pub fn config_file_name() -> &'static str {
+    CONFIG_FILE_NAME
+}
+
```
apiraino, line 21: why this utility function? Can `CONFIG_FILE_NAME` be made public and used directly in `handler.rs`?

src/handlers.rs
```diff
@@ -124,6 +124,29 @@ pub async fn handle(ctx: &Context, event: &Event) -> Vec<HandlerError> {
     errors
 }
 
+async fn validate_triagebot(ctx: &Context, event: &IssuesEvent, errors: &mut Vec<HandlerError>) {
```
apiraino, line 127: As far as I can see this function would inject an HTTP call in the `issue_handlers` macro which is called quite often. I am a bit worried about the additional overhead on every issue the triagebot is called to handle.

What is your goal here @meysam81 ? validate the triagebot config for a certain repository, right? Can this check happen _just once_ on triagebot startup?

Also the name of this function could be a bit more explicit about its purpose (e.g. `validate_triagebot_config` or something similar).

Mark-Simulacrum: This function should be placed in its own module at minimum (similar to all the other handlers). I'd emulate something like rustc_commits in terms of structure.

I would expect that if the goal is to validate PRs, then we would do this only on push events. It probably also makes sense to delegate the actual validation to the config module, not just calling `toml::from_slice` here.

meysam81:
@Mark-Simulacrum 

As you suggested, I refactored the changes into its own module.

You may also see a successful run here: https://github.com/meysam81/triagebot-test/pull/2#issuecomment-1762596593

src/handlers/validate_config.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    let diff = match event.issue.diff(&ctx.github).await {
+        Ok(Some(diff)) => diff,
+        wildcard => {
+            log::error!("Failed to fetch diff: {:?}", wildcard);
+            return Ok(None);
+        }
+    };
+
+    if diff.contains(crate::config::CONFIG_FILE_NAME) {
```
ehuss, line 43: I have a few concerns about this. IIUC, this fetches the full diff on every PR on every push. That seems like it would be expensive, and potentially slow.

Would it be possible to use the `files()` method instead? I would imagine that is more efficient, and is also structured.

Similarly, this looks like it does a string `contains` check, which means that if any text anywhere in the patch contains the bytes `triagebot.toml`, it would trigger. If it looked at the structured file list, then it should be able to look to see if that file is specifically being modified.

I'm still not sure if hitting the `files` endpoint on every push is OK, but I suspect the traffic should be low enough to be fine.

ehuss:
BTW, there is now a merge conflict.

ehuss:
I'm a bit confused about the approach taken here. It seems to validate the config on every PR opened or pushed. I would expect it to operate like it did before, where it would only validate the config on a PR that is modifying `triagebot.toml`. Can you say more about why that is changing? I would also not expect there to be any need to do any caching if it was only checking in the case where `triagebot.toml` is added or modified.

src/handlers/validate_config.rs
```diff
@@ -0,0 +1,150 @@
...
+    if let Some(triagebot_toml) = TRIAGEBOT_CACHE.lock().unwrap().get(&sha.to_string()) {
+        log::info!(
+            "Cache hit for triagebot.toml for repository {} and commit {}",
+            repo,
+            sha
+        );
+        return triagebot_toml.clone();
+    }
+
+    log::warn!(
```
ehuss, line 76: I think doing a warn! on every cache miss would pollute the logs with unactionable information. I would expect something like this to be trace or debug.  Similarly with the info! above about a hit, that information is likely to add noise.

meysam81:
> I'm a bit confused about the approach taken here. It seems to validate the config on every PR opened or pushed. I would expect it to operate like it did before, where it would only validate the config on a PR that is modifying `triagebot.toml`. Can you say more about why that is changing? I would also not expect there to be any need to do any caching if it was only checking in the case where `triagebot.toml` is added or modified.

The main driver for the new approach was to address the expensive HTTP round-trips mentioned here: https://github.com/rust-lang/triagebot/pull/1730#discussion_r1349717753

Additionally, at the expense of using a fixed size memory for holding the config (128 total `triagebot.toml` ATM), we achieve a speed up in case no new commits has been pushed and only a synchronization is triggered.

ehuss:
> The main driver for the new approach was to address the expensive HTTP round-trips mentioned here: https://github.com/rust-lang/triagebot/pull/1730#discussion_r1349717753

My concern in that comment was about using the `diff` endpoint which fetches the entire contents of every PR. Would it be possible to use the `files()` endpoint instead to check if the PR touches `triagebot.toml`, and only download the file in that case? I don't think that approach should need any sort of caching.

In the future, we could potentially cache the `files()` output, since I think there are several other handlers that need to check if a PR touches certain files. But that cache would only need to live as long as the loop through the handlers, and then can be immediately discarded. But I think that is something we could consider later, not something I would bother with in this PR.

ehuss:
Hm, so I'm taking a closer look at this, and I need the diff for something else. We're already fetching the diff 3 times in other handlers. I'll try to follow up soon with something that caches that.

ehuss:
Ok, I posted #1749 to implement caching of the diff (since I need it for something else). If/when that PR is merged, I think this PR can be updated to just call the `diff()` function to determine if it contains `triagebot.toml`, and if it does, fetch the contents and validate it (much like it did before 287b41499331d1828055acdd6ba5fcd00fff92b3).

Sorry about the back and forth on this.

ehuss (approved):

ehuss:
I hope you don't mind, I went ahead and applied some of the changes suggested above. I also made a few other changes:

- Updated toml to the latest version to get better error messages.
- Added deny_unknown_fields to show typo errors to the user.
- Switch to using SHA for the file download instead of the git ref. The git ref has issues with caching, and can take several minutes to update. This also uses the pre-existing code for fetching raw files.

Thanks for helping with this!

apiraino:
> Added deny_unknown_fields to show typo errors to the user.

@ehuss IIUC I think commit 252acb2a1f43caa67b972192a467eb97632764fa caused unexpected side-effects (see [comment](https://github.com/rust-lang/triagebot/pull/1754#issuecomment-1906769792)).

It will report errors also on PRs adding new config options to triagebot.

PR #1733, opened 2023-10-15
Allow switching prs to waiting on review when author has requested a review from an assignee

WaffleLapkin:
Adds a new config option:
```toml
[review-requested]
remove_labels = ["..."]
add_labels = ["..."]
```
When **PR author** requests a review from **a PR assignee**, triagebot will remove `remove_labels` and add `add_labels`.

Based on: https://github.com/rust-lang/rust/pull/116661#discussion_r1359619838.
Todo: document this on [forge](https://github.com/rust-lang/rust-forge/tree/master/src/triagebot).
cc @Mark-Simulacrum

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -108,20 +108,6 @@ pub async fn handle(ctx: &Context, event: &Event) -> Vec<HandlerError> {
         }
     }
 
-    if let Some(config) = config
-        .as_ref()
-        .ok()
-        .and_then(|c| c.review_requested.as_ref())
-    {
-        if let Err(e) = review_requested::handle(ctx, event, config).await {
-            log::error!(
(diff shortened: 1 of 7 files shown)


src/handlers.rs
```diff
@@ -107,6 +108,20 @@ pub async fn handle(ctx: &Context, event: &Event) -> Vec<HandlerError> {
         }
     }
 
+    if let Some(config) = config
```
Mark-Simulacrum, line 111: I would expect that this can go in as an issue handler rather than ad-hoc like this?

WaffleLapkin: Oops, yeah, didn't notice there are issue handlers. Should be fixed now.

Mark-Simulacrum (approved):

PR #1742, opened 2023-11-05
Some scheduling cleanup and automatic types planning stream opening

jackh726:
A couple cleanups for scheduling, and add job from opening types planning meeting thread and requesting updates.

(I can split the cleanups out if needed)

### DIFF
--- a/src/handlers/types_planning_updates.rs
+++ b/src/handlers/types_planning_updates.rs
@@ -1,0 +1,121 @@
+use crate::github;
+use crate::jobs::Job;
+use crate::zulip::BOT_EMAIL;
+use crate::zulip::{to_zulip_id, MembersApiResponse};
+use anyhow::{format_err, Context as _};
+use async_trait::async_trait;
+use chrono::{Duration, Utc};
+
+pub struct TypesPlanningUpdatesJob;
+
+#[async_trait]
+impl Job for TypesPlanningUpdatesJob {
+    fn name(&self) -> &'static str {
+        "types_planning_updates"
+    }
+
+    async fn run(&self, ctx: &super::Context, _metadata: &serde_json::Value) -> anyhow::Result<()> {
+        request_updates(ctx).await?;
+        Ok(())
+    }
+}
+
+const TYPES_REPO: &'static str = "rust-lang/types-team";
+
+pub async fn request_updates(ctx: &super::Context) -> anyhow::Result<()> {
+    let gh = &ctx.github;
+    let types_repo = gh.repository(TYPES_REPO).await?;
+
+    let tracking_issues_query = github::Query {
+        filters: vec![("state", "open"), ("is", "issue")],
+        include_labels: vec!["roadmap-tracking-issue"],
+        exclude_labels: vec![],
+    };
+    let issues = types_repo
+        .get_issues(&gh, &tracking_issues_query)
+        .await
+        .with_context(|| "Unable to get issues.")?;
+
+    for issue in issues {
+        let comments = issue.get_first100_comments(gh).await?;
+        if comments.len() >= 100 {
+            anyhow::bail!(
+                "Encountered types tracking issue with 100 or more comments; needs implementation."
+            );
+        }
+        let older_than_28_days = comments
+            .last()
+            .map_or(true, |c| c.updated_at < (Utc::now() - Duration::days(28)));
+        if !older_than_28_days {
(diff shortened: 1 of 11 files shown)


src/handlers/types_planning_updates.rs
```diff
@@ -0,0 +1,208 @@
...
+        .with_context(|| "Unable to get issues.")?;
+
+    let mut issues_needs_updates = vec![];
+    for issue in issues {
+        // Github doesn't have a nice way to get the *last* comment; we would have to paginate all comments to get it.
+        // For now, just bail out if there are more than 100 comments (if this ever becomes a problem, we will have to fix).
+        let comments = issue.get_first100_comments(gh).await?;
+        if comments.len() >= 100 {
+            anyhow::bail!(
+                "Encountered types tracking issue with 100 or more comments; needs implementation."
```
Mark-Simulacrum, line 107: I would recommend bailing *for that issue* -- e.g., leaving it in the list always. Debugging failures in triagebot requires infra/triagebot team involvement since this error message will be somewhere in our logs (and/or in the db in this case IIRC), which is going to be painful and means you get *zero* pings out rather than pinging excessively for small number of issues.

You could also consider looking at the last updated date on the issue rather than specifically the update date of the last comment, which doesn't run into this problem.

src/handlers/types_planning_updates.rs
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    let issue_list = issues_needs_updates.join("\n");
+
+    let message = format!("The following issues still need updates:\n\n{issue_list}");
+
+    let meeting_date_string = metadata.date_string;
+    let zulip_req = crate::zulip::MessageApiRequest {
+        recipient: crate::zulip::Recipient::Stream {
+            id: 326132,
```
Mark-Simulacrum, line 169: Can you leave a comment here indicating which stream this is supposed to be?

src/handlers/types_planning_updates.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    Ok(())
+}
+
+#[allow(unused)] // Needed for commented out bit above
+async fn zulip_id_and_email(
+    ctx: &super::Context,
+    github_id: i64,
+) -> anyhow::Result<Option<(u64, String)>> {
+    let bot_api_token = std::env::var("ZULIP_API_TOKEN").expect("ZULIP_API_TOKEN");
```
Mark-Simulacrum, line 184: Can we move this function to `zulip.rs`? That will hopefully help avoiding re-implementing it twice, and feels more consistent with the general code structure (Zulip and GitHub handling are in their own modules, handlers just call that code).

Also: I don't think this function is the right way to do this. We should use the team API and the zulip & github IDs present there, which doesn't rely on this information being available to Zulip (which AFAIK isn't guaranteed). So perhaps we can just delete it for now.

src/handlers/types_planning_updates.rs
jackh726: >  e.g., leaving it in the list always

This is a good idea.

> consider looking at the last updated date on the issue

This is a better idea and I'll switch to this.

src/handlers/types_planning_updates.rs
jackh726: Yeah, I'll just make a const at the top, since this is the same as the other job too.

src/handlers/types_planning_updates.rs
jackh726: So, I mostly copied this code from `zulip.rs` (I think, it was somewhere).

The reason I didn't just use the zulip id is because `Recipient::Private` needs to email. Not sure if this strictly required for the API, or just an artifact of the implementation here.

For now, I'll just remove this function. If I need it in the future, I'll do some related cleanups in `zulip.rs`.

Mark-Simulacrum: Sounds good. Yeah, not sure on email vs. Zulip user ID etc.

jackh726:
@Mark-Simulacrum addressed review and fixed failed cron expression parsing (guess cron doesn't support "last Monday of the month")

jackh726:
(Not that I necessary expect this to be merged hastily, but can we wait until at least Tuesday to merge so we don't accidentally trigger *this* months planning meeting job)

Mark-Simulacrum (approved):

PR #1743, opened 2023-11-05
Upgrade cynic to v3

obmarg:
I released v3 of cynic a while back, this pulls that release into triagebot.

The main changes that'll have an impact here are:

1. I've added the ability to pre parse schems in `build.rs` which is now in use here.  This makes a big difference to the compile time of crates using large schemas (such as the GitHub schema triagebot uses).
2. The `schema_for_derives` attribute macro is no longer required so I've got rid of that.
3. Query variables can now be references so I've made that change in the places where it made sense.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ serde_path_to_error = "0.1.2"
 octocrab = "0.30.1"
 comrak = { version = "0.8.2", default-features = false }
 route-recognizer = "0.3.0"
-cynic = { version = "2.0.0" }
+cynic = "3"
 itertools = "0.10.2"
 tower = { version = "0.4.13", features = ["util", "limit", "buffer", "load-shed"] }
 github-graphql = { path = "github-graphql" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ serde_path_to_error = "0.1.2"
 octocrab = "0.30.1"
 comrak = { version = "0.8.2", default-features = false }
 route-recognizer = "0.3.0"
-cynic = "3"
+cynic = "3.2.2"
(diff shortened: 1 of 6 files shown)


github-graphql/Cargo.toml
```diff
@@ -5,4 +5,7 @@ edition = "2021"
 
 [dependencies]
 chrono = { version = "0.4", features = ["serde"] }
-cynic = { version = "2.2.2" }
\ No newline at end of file
+cynic = { version = "3", features = ["rkyv"] }
```
apiraino, line 8:
```suggestion
cynic = { version = "3.2.2", features = ["rkyv"] }
```

I'd probably pin a specific version to avoid accidental updates.

Cargo.toml
```diff
@@ -36,7 +36,7 @@ serde_path_to_error = "0.1.2"
 octocrab = "0.30.1"
 comrak = { version = "0.8.2", default-features = false }
 route-recognizer = "0.3.0"
-cynic = { version = "2.0.0" }
+cynic = "3"
```
obmarg, line 39:
```suggestion
cynic = "3.2.2"
```

github-graphql/Cargo.toml
```diff
@@ -5,4 +5,7 @@ edition = "2021"
...
+
+[build-dependencies]
+cynic-codegen = { version = "3", features = ["rkyv"] }
```
obmarg, line 11:
```suggestion
cynic-codegen = { version = "3.2.2", features = ["rkyv"] }
```

ehuss (approved):
Thanks! I'm going to merge, though I have not tested this, I assume if it compiles it should be good.

PR #1719, opened 2023-09-12
 New pull request assignment proposal

apiraino:
This patchset is the implementation of a new workflow for assigning pull requests to the Rust project contributors.

**2023-11-08 UPDATE:** After a meeting with T-infra, it was remarked that this block of changes are hard to review, so we agreed a plan to split into smaller bites. This PR will stay as tracking issues for the work being split into smaller pull requests:

- [ ] Step 1: Change pull request assignment algo (#1745 ):
  - create new DB table, just store there "how busy ppl are"
  - add new mechanism to assign pull requests
  - add CLI tool to populate the new DB table and synchronize it. Make it so that it will speak HTTP instead of using public triagebot APIs.

- [ ] Step 2: Implement the backoffice where people can set themselves on time-off.
  - TBD: standalone app or inside the Triagebot codebase?

---

A slightly old design document is at [this HackMD link](https://hackmd.io/GqXA4nyTQ4m__hRZUayVRg).

This proposal was discussed separately in a number of meetings with the compiler team (also mentioned to @Mark-Simulacrum).

## Summary

The current pull request assignment is basically randomly assigned among team members. The new proposed workflow has the following features:
- each Rust project contributor belonging to a team (I'd start will a small cohort of testers) can set their own review capacity and time off using a web backoffice
- the web backoffice will persiste these data in a DB
- everytime a pull request assignment is invoked with `r? <team> or <team_member>` the triagebot will check the current availability and workload of the candidates and assign the pull request to the team member less busy, excluding automatically those that are on time off in that moment.

## Permissions and Authorization

This backoffice is a simple form listing all team members. It is served by the triagebot (from triagebot.infra.rust-lang.org/). Access to this backoffice is gated by a GitHub App. Users allowed to this backoffice:
1) must be logged into GitHub
2) must agree to the share some details with a GitHub App (only to read their profile)
3) must be a Rust project team member
4) the team they belong must be whitelisted in the env var `NEW_PR_ASSIGNMENT_TEAMS` (see [documentation](https://github.com/rust-lang/triagebot/blob/8087779fd66be25f0e30cef8cbf423dd90940d51/pr_prefs_backoffice.md))

Team leaders are considered administrators and can see the preferences of every team member. Normal team members can only see their own preferences and those of team members that agree to share theirs within the team.

## DEPLOYMENT

[...]

PR #1745, opened 2023-11-08
Add new PR review assignment

apiraino:
**UPDATE 2023-12-12**: this second proposal was discussed with @jdno, we agreed a plan detailing how to split it further. I'm going to close this PR and refactor another patch according to this plan.

----

This is the implementation of a new workflow for assigning pull requests to the Rust project contributors.

This new workflow will assign a pull request to be reviewed based on the number of assigned pull requests to team members.

Everytime a pull request assignment is invoked with `r? <team>` or `r? <team_member>` the Triagebot will check the current workload of the candidates and assign the pull request to the team member less busy.

The new workflow is DISABLED by default. It can be enabled by setting the env variable `USE_NEW_PR_ASSIGNMENT` (any value) and restarting the Triagebot.

Teams that are subject to the new PR review assignment are listed as a comma separated list in the env variable `NEW_PR_ASSIGNMENT_TEAMS`. See `.env.sample` for a usage example.

Team members workload is tracked in a new DB table `review_capacity`.

Both the initial population and the synchronization of this table are handled by a command line tool that will speak to the Triagebot through a number of HTTP endpoints. These HTTP endpoints must be PRIVATE and accessible ONLY to the sync tool. This should be handled at the infrastructure level, the endpoints are not authenticated.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,6 +8,7 @@ use crate::github::PullRequestDetails;
 use anyhow::Context;
 use handlers::HandlerError;
 use interactions::ErrorComment;
+use serde::Serialize;
 use std::fmt;
 use tracing as log;
 
@@ -25,7 +26,7 @@ pub mod notification_listing;
 pub mod payload;
 pub mod rfcbot;
 pub mod team;
-mod team_data;
+pub mod team_data;
 pub mod triage;
 pub mod zulip;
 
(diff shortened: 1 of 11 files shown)


apiraino:
cc: @Kobzol @jdno since we discussed the refactor of #1719

This PR is in a draft state and waiting for a feedback about the client that will feed and sync the database of team members workload.

To recap: I need a little nudge from T-infra if the approach of using HTTP API requests to populate the DB is correct and compatible with the current infra.

thanks 🙂

jdno:
Would it be possible to break this apart even further? I'm wondering if we can first change the assignment logic to look purely at the number of currently assigned PRs, then consider the preferences (if they exist) in the second step, and finally figure out how to set them in a third PR.

The motivation for this is to keep each individual pull request small. And as I understand it, we need to consider the case when a user hasn't provided preferences yet anyways.

apiraino:
hey @jdno yeah I get your point about splitting and in principle I totally agree (thanks for the checking my PR btw :slightly_smiling_face: ).

> first change the assignment logic to look purely at the number of currently assigned PRs

Upon reviewing again my code I am not sure it would work great. My understanding is that there is no place in the entire Rust infra where we store the information about how many PRs has a contributor assigned at any given time. The DB table I am introducing [here](https://github.com/rust-lang/triagebot/pull/1745/files#diff-ec4db46735c0a96723b82c6defbd4515b996919bd88d026007e00055006ccc20R125) would be the first place where we persist this piece of info. So far, is everything I say correct? I hope I am not missing anything.

Now, without that DB table I would be forced to retrieve dinamically the data. That means that everytime a PR is assigned (or reassigned) the triagebot should emit a number of expensive HTTP calls to Github to calculate the current number of PRs assigned for *all* members of a team.

Then find the least busy team member, assign them the PR, throw away the result and start all over again at the next PR assignment.

Am I understanding correctly your point?

> And as I understand it, we need to consider the case when a user hasn't provided preferences yet anyways.

This would be taken care by the [DB table defaults](https://github.com/rust-lang/triagebot/pull/1745/files#diff-7d7d3a6ce5c2a3e8b39ce4f49b78a71365c8e06c0d9ddb78bc2dca68bfeb4a5bR277-R289) until we have the backoffice where people can set their own prefs.

src/lib.rs
```diff
@@ -261,3 +262,68 @@ pub async fn webhook(
...
+    pub checksum: Option<String>,
+}
+
+impl ReviewCapacityUser {
+    pub(crate) fn is_available(&self) -> bool {
+        let today = chrono::Utc::today().naive_utc();
+        let is_available = (self.pto_date_end.is_some() && self.pto_date_start.is_some())
+            && (self.pto_date_end.unwrap() < today || self.pto_date_start.unwrap() > today);
+        self.active && is_available
+    }
```
spastorino, line 292:
```suggestion
match (self.active, self.pto_date_start, self.pto_date_end) {
    (true, Some(pto_date_start), Some(pto_date_end)) => {
        let today = chrono::Utc::today().naive_utc();
        today < pto_date_start || pto_date_end < today
    } 
    _ => false;
}
```

PR #1747, opened 2023-11-13
Add detail about how to raise MCP concerns

apiraino:
During the weekly T-compiler triage meetings we list all MCPs in different stages of progress. Some of them are blocked on concerns not yet resolved. We would like to list these open concerns so that meeting attendees have a glance why an MCP is not seconded or does not move forward.

In order to automate this when creating the meeting agenda it is useful if people respect a standard syntax for that so I can machine parse the issue comments and with a bit of regexps extract the relevant info.

<details><summary>Current message when opening a MCP:
</summary>
<p>

screenshot-20231113-150350

</p>
</details> 

<details><summary>The new version should look like this</summary>
<p>

screenshot-20231113-150509

</p>
</details> 

Note: the `@rustbot` syntax here won't trigger anything, It's just a psychological trick to make it look like that a command is executed but in reality it's just to facilitate the machine parsing for me.

Idea [mentioned on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.E2.9C.94.20.5Bweekly.5D.202023-10-26/near/398697369).

Documentation update for forge.rust-lang.org: https://github.com/rust-lang/rust-forge/pull/709

Thanks for looking at this patch!

r? @Mark-Simulacrum / @ehuss

### DIFF
--- a/src/handlers/major_change.rs
+++ b/src/handlers/major_change.rs
@@ -264,6 +264,27 @@ async fn handle(
         procedural comments, such as volunteering to review, indicating that you \
         second the proposal (or third, etc), or raising a concern that you would \
         like to be addressed. \
+        \n\n \
+        Concerns or objections to the proposal should be discussed on Zulip and formally registered \
+        here by adding a comment with the following syntax: \
+        \n \
+        ``` \
+        \n \
+        @rustbot concern reason-for-concern \
+        \n \
+        <description of the concern> \
+        \n \
+        ``` \
+        \n \
+        Concerns can be lifted with: \
+        \n \
+        ``` \
+        \n \
+        @rustbot resolve reason-for-concern \
+        \n \
+        ``` \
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
I guess no error. I think this is fine but maybe we can say a bit more instead of register - e.g., "formally declare an objection"? It might also be good to have docs we can link to on the actual process here - e.g., should folks clear the concern somehow?

apiraino:
I agree. I's a hack and not particularly very elegant. But if you think that providing enough context and documentation makes it palatable, I'm happy to do that!

apiraino:
ok @Mark-Simulacrum updated the text (hope I got the newlines right) and updated the documentation at the forge: https://github.com/rust-lang/rust-forge/pull/709

Issue #1753, opened 2023-12-12
[Tracking] New pull request assignment proposal


apiraino:
This issue tracks the progress of the implementation of a new workflow for assigning pull requests to the Rust project contributors.

**2023-12-12 UPDATE**: After another meeting with T-infra, we agreed a second plan detailing how to split the work further. The new plan is detailed below.

**2023-11-08 UPDATE:** After a meeting with T-infra, it was remarked that the first implementation (#1719) was hard to review so we agreed a plan to split into smaller bites (#1745)

- [ ] Step 1
  - add the DB table, strip down DB fields to user_id,assigned_prs
  - add code to keep track of assign/unassign'ed PRs
  - no migration/import/sync tooling
- [ ] Step 2
  - add logic to assign PRs based on a `count()` of your assigned PRs
  - prepare ENV VAR to switch upon restarting triagebot
- [ ] Step 3 (to be clarified)
  - HTTP endpoint to sync assignment counts with GitHub
- [ ] Step 4 (to be clarified)
  - Tackle the preferences: HTML or chat-based?
- [ ] Step 5 (to be clarified)
  - a tool (cli?) for migration/sync of PRs from Github?
  - how to sync PRs when/if bors webooks fail?
  - backoffice: add a button to the contributor to resync their own situation

---

## Summary

The current pull request assignment is basically randomly assigned among team members. The new proposed workflow has the following features:
- each Rust project contributor belonging to a team (I'd start will a small cohort of testers) can set their own review capacity and time off using a web backoffice
- the web backoffice will persiste these data in a DB
- everytime a pull request assignment is invoked with `r? <team> or <team_member>` the triagebot will check the current availability and workload of the candidates and assign the pull request to the team member less busy, excluding automatically those that are on time off in that moment.

A slightly old design document is at [this HackMD link](https://hackmd.io/GqXA4nyTQ4m__hRZUayVRg).

## Permissions and Authorization

This backoffice could be thought as a simple form listing all team members. It is served by the triagebot (from triagebot.infra.rust-lang.org). Access to this backoffice is gated by a GitHub App. Users allowed to this backoffice:
1) must be logged into GitHub
2) must agree to the share some details with a GitHub App (only to read their profile)
3) must be a Rust project team member
4) the team they belong must be whitelisted in the env var `NEW_PR_ASSIGNMENT_TEAMS`

[...]

PR #1754, opened 2023-12-13
Workflow for tracking PRs assignment

apiraino:
General overview at: https://github.com/rust-lang/triagebot/issues/1753

This patch implements the first part:
- a new DB table with just the fields to track how many PRs are assigned to a contributor at any time
- Update this table everytime a PR is assigned or unassigned

No initial sync is planned at this time. Populating the table will happen over time.

This feature is enabled by adding the following to the triagebot.toml in the Rust git repository:
```
# Enable this feature to keep track of the PR review assignment
[review-work-queue]
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -35,6 +35,7 @@ pub(crate) struct Config {
     pub(crate) note: Option<NoteConfig>,
     pub(crate) mentions: Option<MentionsConfig>,
     pub(crate) no_merges: Option<NoMergesConfig>,
+    pub(crate) review_work_queue: Option<TeamMemberWorkQueueConfig>,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -150,6 +151,13 @@ pub(crate) struct RelabelConfig {
 pub(crate) struct ShortcutConfig {
     #[serde(default)]
     _empty: (),
+}
+
+#[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub(crate) struct TeamMemberWorkQueueConfig {
+    #[serde(default)]
+    pub(crate) enabled_for_teams: Vec<String>,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -357,6 +365,9 @@ mod tests {
             [assign]
             users_on_vacation = ["jyn514"]
 
+            [review-work-queue]
+            enabled-for-teams = ["aaa", "bbb"]
+
             [note]
 
             [ping.compiler]
@@ -429,6 +440,9 @@ mod tests {
                 github_releases: None,
                 review_submitted: None,
                 review_requested: None,
+                review_work_queue: Some(TeamMemberWorkQueueConfig {
+                    enabled_for_teams: vec!["aaa".into(), "bbb".into()]
+                }),
                 mentions: None,
                 no_merges: None,
             }
--- a/src/db.rs
+++ b/src/db.rs
@@ -308,4 +308,13 @@ CREATE UNIQUE INDEX jobs_name_scheduled_at_unique_index
         name, scheduled_at
     );
 ",
+    "
+CREATE table review_prefs (
+    id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
+    user_id BIGINT REFERENCES users(user_id),
+    assigned_prs INT[] NOT NULL DEFAULT array[]::INT[],
+    num_assigned_prs INTEGER
+);
+CREATE UNIQUE INDEX review_prefs_user_id ON review_prefs(user_id);
+",
 ];
--- a/src/db.rs
+++ b/src/db.rs
@@ -312,8 +312,7 @@ CREATE UNIQUE INDEX jobs_name_scheduled_at_unique_index
 CREATE table review_prefs (
     id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
     user_id BIGINT REFERENCES users(user_id),
-    assigned_prs INT[] NOT NULL DEFAULT array[]::INT[],
-    num_assigned_prs INTEGER
+    assigned_prs INT[] NOT NULL DEFAULT array[]::INT[]
 );
 CREATE UNIQUE INDEX review_prefs_user_id ON review_prefs(user_id);
 ",
--- a/src/handlers/review_work_queue.rs
+++ b/src/handlers/review_work_queue.rs
@@ -1,0 +1,157 @@
+use crate::{
+    config::TeamMemberWorkQueueConfig,
+    github::{IssuesAction, IssuesEvent},
+    handlers::Context,
+    TeamMemberWorkQueue,
+};
+use anyhow::Context as _;
+use tokio_postgres::Client as DbClient;
+use tracing as log;
+
+// This module updates the PR work queue of team members
+// - When a PR has been assigned, adds the PR to the work queue of team members
+// - When a PR is unassigned, removes the PR from the work queue of all team members
+
+/// Get all assignees for a pull request
+async fn get_pr_assignees(
+    db: &DbClient,
+    issue_num: i32,
+) -> anyhow::Result<Vec<TeamMemberWorkQueue>> {
+    let q = "
+SELECT u.username, r.*
+FROM review_prefs r
+JOIN users u on u.user_id=r.user_id
+WHERE $1 = ANY (assigned_prs)";
+    let rows = db.query(q, &[&issue_num]).await?;
+    Ok(rows
+        .into_iter()
+        .filter_map(|row| Some(TeamMemberWorkQueue::from(row)))
+        .collect())
+}
+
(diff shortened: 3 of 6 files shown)


apiraino:
@jdno I think we can have a look at thist first step of implementation, as per the tracking issue #1753

cc: @rust-lang/infra 

thanks!

src/config.rs
```diff
@@ -357,6 +365,9 @@ mod tests {
             [assign]
             users_on_vacation = ["jyn514"]
 
+            [review-work-queue]
+            enabled-for-teams = ["aaa", "bbb"]
```
apiraino, line 369: I would like to enable this feature only for a subset of teams, rather than tracking all PRs of every contributor from the start.

But that would add a little of complexity to this PR. Maybe worth adding in a subsequent patch, when/if this one is merged?

src/handlers/review_work_queue.rs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            assignee.assigned_prs.swap_remove(index);
+        }
+        update_team_member_workqueue(&db_client, &assignee).await?;
+    }
+
+    // 2) create or increase by one the team members work queue
+    // create team members if they don't exist
+    for u in event.issue.assignees.iter() {
+        let user_id = u.id.expect("Github user was expected! Please investigate.");
```
apiraino, line 143: This is just to cover the unlikely case where we receive a `User` with an id=None

Unsure why the [User struct](https://github.com/rust-lang/triagebot/blob/92c3399ac77706649c2026279f124d4e61e6d899/src/github.rs#L21) is like that but better safe than sorry.

src/db.rs
```diff
@@ -307,5 +307,14 @@ CREATE UNIQUE INDEX jobs_name_scheduled_at_unique_index
     ON jobs (
         name, scheduled_at
     );
+",
+    "
+CREATE table review_prefs (
+    id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
+    user_id BIGINT REFERENCES users(user_id),
+    assigned_prs INT[] NOT NULL DEFAULT array[]::INT[],
+    num_assigned_prs INTEGER
```
Kobzol, line 316: Why is `num_assigned_prs` stored in the DB? Isn't that just the length of `assigned_prs`? If yes, then it should be calculated using derived SQL queries, like `SELECT array_length(schedule, 1)`.

apiraino: I usually do that so later I can add an index and have good performances on queries that involve that field (f.e. *who has most PRs assigned?*).

But that table will have ... 200 records at most? 400? So probably the perf. gain is negligible? :man_shrugging:

Kobzol: Yeah, that sounds like a big overkill :laughing: And breaks the DB normal form unnecessary, IMO. The array also seems fishy, since we don't really need the potential perf. gains here, but that at least reduces the number of requires tables.

apiraino: Removed the field in https://github.com/rust-lang/triagebot/pull/1754/commits/f8ff17d29f55e69b1fe2dbebe6febd76a12cf3fa

apiraino:
gentle ping for review @rust-lang/infra 

(maybe @ehuss?)

thanks

Kobzol:
CC @jackh726, if you could please take a look.

src/db.rs
```diff
@@ -307,5 +307,13 @@ CREATE UNIQUE INDEX jobs_name_scheduled_at_unique_index
     ON jobs (
         name, scheduled_at
     );
+",
+    "
+CREATE table review_prefs (
```
jackh726, line 312: Is the intention that this will eventually contain review preferences too? (That's what the name suggests)

My first thought was that there should instead be a table of PRs, with assigned reviewers in that list. But, maybe that doesn't make a bunch of sense.

src/handlers/review_work_queue.rs
```diff
@@ -0,0 +1,152 @@
...
+    _config: &TeamMemberWorkQueueConfig,
+    event: &IssuesEvent,
+    _inputs: ReviewPrefsInput,
+) -> anyhow::Result<()> {
+    let db_client = ctx.db.get().await;
+    let iss_num = event.issue.number as i32;
+
+    // Note: When changing assignees for a PR, we don't receive the assignee(s) removed, we receive
+    // an event `Unassigned` and the remaining assignees
+
```
jackh726, line 120: I would expect this function to "diff" the current assignees with the new assigness, and only update based on that diff...

jackh726: This is also why it makes sense to me that the table be of PRs with a list of assignees.

But, that complicates the querying. Could think about a table view that aggregates.

src/handlers/review_work_queue.rs
```diff
@@ -0,0 +1,152 @@
...
+RETURNING u.username, r.*, array_length(assigned_prs, 1) as num_assigned_prs";
+    let rec = db
+        .query_one(q, &[&assignee.user_id, &assignee.assigned_prs])
+        .await
+        .context("Update DB error")?;
+    Ok(rec.into())
+}
+
+/// Add a new user (if not existing)
+async fn maybe_create_team_member(
```
jackh726, line 51: Rename to `ensure_team_member`

triagebot.toml
```diff
@@ -3,6 +3,12 @@ allow-unauthenticated = ["bug", "invalid", "question", "enhancement"]
 
 [assign]
 
+# Enable this feature to keep track of the PR review assignment
+[review-work-queue]
+# A list of teams that will use the new PR review assignment workflow.
+# Team names match https://github.com/rust-lang/team/tree/master/teams
+enabled-for-teams=["compiler", "compiler-contributors"]
```
jackh726, line 10: Nit: spacing here (space before and after `=`) should match rest of the file.

jackh726 (changes requested):

jackh726:
I'm thinking:

It's pretty trivial to get a list of all open PRs and their assignees:

Here's the graphql query (you do still have to paginate):

```graphql
query {
  repository(owner: "rust-lang", name: "rust") {
    pullRequests(first: 100, states:OPEN) {
      pageInfo {
        hasNextPage
        endCursor
      }
      nodes {
        number
        assignees(first: 100) {
          nodes {
            login
            id
          }
        }
      }
    }
  }
}
```

It's probably worth just setting up a sync once on start, and again every 30 minutes, along with listening for events. I don't think only just waiting for the table to populate over time will really work.

I'd be curious to know how long it would take to run a whole "query the current state and update the DB". Probably not that long - even if we *way* over estimate, we should only have on the order of a 1000 PRs to deal with.

src/db.rs
apiraino: Correct. The new DB table at the end of the story will also persist the review preferences for each reviewer (thus the name). I'm trying to slip in changes one by one and as slowly as possible, as per the plan detailed in #1753

PRs in a separate joined table def. make sense in general. But in this case I think the use case is so minimal to not warrant that. _If_ in the far future the tool develops such as to make it a compelling choice, we can always migrate.

src/handlers/review_work_queue.rs
apiraino: > I would expect this function to "diff" the current assignees with the new assigness, and only update based on that diff

Hm, I'm a bit unclear on this comment :thinking: do you suggest a "3-way" comparison between two arrays (list of current assignees from the DB vs. new list of assignees from GH)? Example:
- assign to new assignees
- check if current assignees are still such
- unassign from removed assignees

(or maybe something else?)

apiraino:
> It's probably worth just setting up a sync once on start, and again every 30 minutes, along with listening for events. I don't think only just waiting for the table to populate over time will really work.

Why wouldn't it? Can you elaborate a bit on that?

I'll think about your proposal but my immediate thought I think is that I don't agree with this. We would be losing the concept of a stable "source of truth". I am afraid that querying GH every _n_-minutes would introduce a point of failure in the consistency, the DB would just become a 'cache' reflecting the GH situation. Then why not just make it a full real cache serializing a json and avoid the DB table altogether?

My reasoning about having the DB table is to have a storage persistence that can be used by the PR review assignment and not only by that. I imagine a future where it could be expanded and used also for statistics, for everything we want to know about PRs and reviews.

src/handlers/review_work_queue.rs
jackh726: Yeah, I mean, just compare the lists and find what needs to be removed and what needs to be added.

jackh726:
> We would be losing the concept of a stable "source of truth". I am afraid that querying GH every _n_-minutes would introduce a point of failure in the consistency, the DB would just become a 'cache' reflecting the GH situation.

The concern I have is that Github *is* the source of truth right now. Do you expect otherwise?

>Then why not just make it a full real cache serializing a json and avoid the DB table altogether?

I mean, this isn't a *bad* option, to keep just everything in memory. But seems unnecessary. 

>I imagine a future where it could be expanded and used also for statistics, for everything we want to know about PRs and reviews.

I think this is useful, but not with the current table schema.

PR #1756, opened 2023-12-16
make triagebot cuter

jyn514:
hi i am new rust user can one of you experienced rust people review my code please

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -39,7 +39,7 @@ mod tests {
 }
 
 const NEW_USER_WELCOME_MESSAGE: &str = "Thanks for the pull request, and welcome! \
-The Rust team is excited to review your changes, and you should hear from {who} soon.";
+The Rust team is excited to review your changes, and you should hear from {who} soon. (°꒳°)";
 
 const CONTRIBUTION_MESSAGE: &str = "Please see [the contribution \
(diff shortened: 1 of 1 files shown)


src/handlers/assign.rs
```diff
@@ -56,17 +56,17 @@ const WELCOME_WITHOUT_REVIEWER: &str = "@Mark-Simulacrum (NB. this repo may be m
 
 const RETURNING_USER_WELCOME_MESSAGE: &str = "r? @{assignee}
 
-({bot} has picked a reviewer for you, use r? to override)";
+({bot} has picked a reviewer for you, use r? to override) (˶˃ᆺ˂˶)";
```
Nilstrieb, line 59: I'm not sure whether this is appropriate here. Having a reviewer is something very positive, and they don't feel that happy to me. Maybe replace it with something more happy? The other ones fit very well though

jyn514: oh ok is this happy enough like this:

(*・ω・)ﾉ

Nilstrieb: that sounds great ferriscat

apiraino:
tbh I find this addition not really necessary. Why?
- Point 1: I (among others) try to recognize the work of contributors on both ends of a PR (author and reviewers) with an explicit "thank you" message, imo more direct and personal than a hardcoded emoji
- Point 2: I don't even know what that means and like maybe others don't

PR #1758, opened 2024-01-15
Fixes for Clippy lints

pitaj:


parser/src/command/glacier.rs
```diff
@@ -45,7 +45,7 @@ impl GlacierCommand {
                 Some(Token::Quote(s)) => {
                     let source = s.to_owned();
                     if source.starts_with("https://gist.github.com/") {
-                        return Ok(Some(GlacierCommand { source }));
+                        Ok(Some(GlacierCommand { source }))
                     } else {
                         return Err(toks.error(ParseError::InvalidLink));
```
jackh726, line 50: Why is this `return` not removed?

parser/src/command/ping.rs
```diff
@@ -47,20 +47,20 @@ impl PingCommand {
             if let Some(Token::Dot) | Some(Token::EndOfLine) = toks.peek_token()? {
                 toks.next_token()?;
                 *input = toks;
-                return Ok(Some(PingCommand { team }));
+                Ok(Some(PingCommand { team }))
             } else {
                 return Err(toks.error(ParseError::ExpectedEnd));
```
jackh726, line 52: and here

parser/src/command/glacier.rs
pitaj: Because clippy didn't say to remove it, but I suppose it may as well be removed. Will do

apiraino:
cc @pitaj in case you have time to tick these pending comments.

In general I am not a big fan of massive linting patches, takes more time to review than prepare the patch :)

(I assume changes have been tested - as far as possible with the triagebot codebase)

Thanks!

PR #1761, opened 2024-01-20
remove labels added by `no-merges` when rebase is pushed

### LINKED ISSUE #1757: Merge commit label should be removed automatically
When someone pushes merge commits, the merge commits label is added. But when they then push those merge commits away, the label is not removed, making it more annoying for the author or the reviewer to remove it. This is something that could and should be automated too.

cc @pitaj

pitaj:
closes #1757

@Nemo157 I did look into hiding prior comments but it doesn't seem supported by the existing APIs in triagebot. Looks to be pretty involved so figured I would push this in the meantime.

### DIFF
--- a/src/handlers/no_merges.rs
+++ b/src/handlers/no_merges.rs
@@ -24,6 +24,9 @@ pub(super) struct NoMergesInput {
 struct NoMergesState {
     /// Hashes of merge commits that have already been mentioned by triagebot in a comment.
     mentioned_merge_commits: HashSet<String>,
+    /// Labels that the bot added as part of the no-merges check.
+    #[serde(default)]
+    added_labels: Vec<String>,
 }
 
 pub(super) async fn parse_input(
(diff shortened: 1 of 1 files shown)


src/handlers/no_merges.rs
```diff
@@ -24,6 +24,8 @@ pub(super) struct NoMergesInput {
 struct NoMergesState {
     /// Hashes of merge commits that have already been mentioned by triagebot in a comment.
     mentioned_merge_commits: HashSet<String>,
+    /// Labels that the bot added as part of the no-merges check.
+    added_labels: Vec<String>,
```
ehuss, line 28: I believe this will cause a problem with any entries already in the database, as it will fail to deserialize, since this field is now required.

(And unfortunately it causes a panic, which will break triagebot handling for that pr.)

pitaj: Would using `Option<Vec<String>>` avoid that?

ehuss: I think `#[serde(default)]` should work.

ehuss (approved):
Thanks!

PR #1765, opened 2024-01-24
Add build profile optimized for binary size

apiraino:
This patch adds a new build profile optimized for binary size.

If we exclude the `triagebot` application, when building the other binaries under `./src/bin` I think it can make sense to optimize for binary size rathen than performance. When taken to the extreme (as in this patch), the differenze is just _huge_, see the following table (rust nightly 1.77):

```
$ la -h target/{release,opt-size}/{compiler,types,lang,prioritization-agenda}
-rwxr-xr-x 2 me me 4.0M Jan 24 17:13 target/opt-size/compiler
-rwxr-xr-x 2 me me 4.1M Jan 24 17:13 target/opt-size/lang
-rwxr-xr-x 2 me me 4.1M Jan 24 17:13 target/opt-size/prioritization-agenda
-rwxr-xr-x 2 me me 4.0M Jan 24 17:13 target/opt-size/types
-rwxr-xr-x 2 me me 140M Jan 24 17:21 target/release/compiler
-rwxr-xr-x 2 me me 141M Jan 24 17:21 target/release/lang
-rwxr-xr-x 2 me me 141M Jan 24 17:21 target/release/prioritization-agenda
-rwxr-xr-x 2 me me 140M Jan 24 17:21 target/release/types
```

I like to run the build of these ancillary tools with `cargo build --bins --profile=opt-size`.

Opinions?

Thanks for a review.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,3 +56,12 @@ default-features = false
 
 [profile.release]
 debug = 2
+
+[profile.opt-size]
+inherits = "release"
+opt-level = "z"
+lto = "on"
+codegen-units = 1
+panic = "abort"
+incremental = false
+strip = true


Cargo.toml
```diff
@@ -56,3 +56,12 @@ default-features = false
 
 [profile.release]
 debug = 2
+
+[profile.opt-size]
+inherits = "release"
+opt-level = "z"
+lto = "on"
+codegen-units = 1
+panic = "abort"
```
apiraino, line 65: This flag removes the backtraces in case of error ([docs](https://doc.rust-lang.org/cargo/reference/profiles.html#panic)).

(possibly a controversial flag, so I wanted to point it out)

Mark-Simulacrum: I don't think that's true? It only means that Drop impls won't run, you should still be able to get a backtrace (presuming there's sufficient debug info...)

Cargo.toml
```diff
@@ -56,3 +56,12 @@ default-features = false
 
 [profile.release]
 debug = 2
```
Mark-Simulacrum, line 58: Can we tweak this to = 1 which should be enough for good backtraces and see if that solves most of the size problem?

Cargo.toml
apiraino: ok. sorry for my misunderstarding

Cargo.toml
apiraino: Still a bit bigger (~61 mb) than using the profile in this patch (~4 mb).
```
$ la -h target/{release,debug1}/{compiler,types,lang,prioritization-agenda}
-rwxr-xr-x 2 me me  61M Jan 25 11:39 target/debug1/compiler
-rwxr-xr-x 2 me me  62M Jan 25 11:39 target/debug1/lang
-rwxr-xr-x 2 me me  62M Jan 25 11:39 target/debug1/prioritization-agenda
-rwxr-xr-x 2 me me  61M Jan 25 11:39 target/debug1/types
-rwxr-xr-x 2 me me 140M Jan 25 11:36 target/release/compiler
-rwxr-xr-x 2 me me 141M Jan 25 11:36 target/release/lang
-rwxr-xr-x 2 me me 141M Jan 25 11:36 target/release/prioritization-agenda
-rwxr-xr-x 2 me me 140M Jan 25 11:36 target/release/types
```

The second biggest "contributor" in the profile in this patch is `strip = true` that brings the size from ~61mb down to ~6mb so probably:
```
[profile.release]
debug = 1
strip = true
```
would bring home most of the benefits.

apiraino: Anyway: no problem, I'll just keep this patch in a stash for personal use.

Thanks for your time.

PR #1766, opened 2024-01-29
Configure team nominate label in prioritization alerts on Zulip

apiraino:
This patch tries to customize the prioritization Zulip notification (`I-{team}-nominated`) according to the team labels on the issue.

It's not complete, there are some cases not covered but I didn't feel complicating too much the logic.

And this is a brainstoming for @fmease that suggested doing so. Feel free to own this patch and bring it to completion (unsure you have permissions on this repo) or suggest how you would like the implementation to be. 

This patch completes https://github.com/rust-lang/rust/pull/119959#pullrequestreview-1820272630

Thanks!

### DIFF
--- a/src/handlers/notify_zulip.rs
+++ b/src/handlers/notify_zulip.rs
@@ -167,6 +167,7 @@ pub(super) async fn handle_input<'a>(
         for msg in msgs {
             let msg = msg.replace("{number}", &event.issue.number.to_string());
             let msg = msg.replace("{title}", &event.issue.title);
+            let msg = replace_team_to_be_nominated(&event.issue.labels, msg);
 
             crate::zulip::MessageApiRequest {
                 recipient,
@@ -178,4 +179,67 @@ pub(super) async fn handle_input<'a>(
     }
 
     Ok(())
+}
+
+fn replace_team_to_be_nominated(labels: &[Label], msg: String) -> String {
+    let teams = labels
+        .iter()
+        .map(|label| &label.name)
+        .filter_map(|label| label.strip_prefix("T-"))
+        .collect::<Vec<&str>>();
+
+    // - If a single team label is found, replace the placeholder with that one
+    // - If multiple team labels are found and one of them is "compiler", pick that one
+    //   (currently the only team handling these Zulip notification)
+    // - else, do nothing
+    if let [team] = &*teams {
+        msg.replace("{team}", team)
+    } else if teams.contains(&"compiler") {
+        msg.replace("{team}", "compiler")
+    } else {
+        msg
+    }
+}
+
+#[test]
+fn test_notification() {
+    let mut msg = replace_team_to_be_nominated(&[], "Needs `I-{team}-nominated`?".to_string());
+    assert!(msg.contains("Needs `I-{team}-nominated`?"), "{}", msg);
+
+    msg = replace_team_to_be_nominated(
+        &[Label {
+            name: "T-cooks".to_string(),
+        }],
+        "Needs `I-{team}-nominated`?".to_string(),
+    );
+    assert!(msg.contains("I-cooks-nominated"), "{}", msg);
+
+    msg = replace_team_to_be_nominated(
+        &[
+            Label {
+                name: "T-compiler".to_string(),
+            },
+            Label {
+                name: "T-libs".to_string(),
+            },
+            Label {
+                name: "T-cooks".to_string(),
+            },
+        ],
+        "Needs `I-{team}-nominated`?".to_string(),
+    );
+    assert!(msg.contains("I-compiler-nominated"), "{}", msg);
+
+    msg = replace_team_to_be_nominated(
+        &[
+            Label {
+                name: "T-libs".to_string(),
+            },
+            Label {
+                name: "T-cooks".to_string(),
+            },
+        ],
+        "Needs `I-{team}-nominated`?".to_string(),
+    );
+    assert!(msg.contains("Needs `I-{team}-nominated`?"), "{}", msg);
 }


apiraino:
hey @fmease I was looking into completing this small change.

wdyt? Does it look like what you wanted to accomplish?

fmease:
It definitely does what I had envisioned, thanks! 😊 The major issue I see is that this PR adds very rust-lang/rust-specific logic to triagebot. While triagebot still contains a lot of things specific to rust-lang/rust, I think the general vision for this bot is to move anything repo-specific to configs & to make it repo-agnostic.

cc @WaffleLapkin

fmease:
And I really don't know how one would best abstract over this feature and make it configurable.
If the maintainers of this repo are fine with merging this as is, then that's awesome, if they aren't we need to change the approach.

@Mark-Simulacrum, what do you think?

src/handlers/notify_zulip.rs
```diff
@@ -172,3 +173,69 @@ pub(super) async fn handle_input<'a>(
 
     Ok(())
 }
+
+fn get_team_to_be_nominated(labels: &Vec<Label>, msg: String) -> String {
```
Mark-Simulacrum, line 177: This doesn't sound like "get", more of a "replace" or "substitute"?

src/handlers/notify_zulip.rs
```diff
@@ -172,3 +173,69 @@ pub(super) async fn handle_input<'a>(
...
+    // - If multiple team labels are found and one of them is "compiler", pick that one
+    //   (currently the only team handling these Zulip notification)
+    // - else, return a generic text
+    if teams.len() == 1 {
+        msg.replace("{team}", teams[0])
+    } else if teams.contains(&"compiler") {
+        msg.replace("{team}", "compiler")
+    } else {
+        msg.replace(
+            "Needs `I-{team}-nominated`?",
```
Mark-Simulacrum, line 194: This is hardcoding a non-replacement string -- should we really do that? I would expect this trailing case to be just `msg` I think, i.e., no replacement is performed.

WaffleLapkin:
@fmease all zulip stuff is highly specific to rust-lang. so while I'd like it to be more configurable, this specific change does not really change much.

src/handlers/notify_zulip.rs
apiraino: > This is hardcoding a non-replacement string -- should we really do that? I would expect this trailing case to be just `msg` I think, i.e., no replacement is performed.

I thought a reasonable default message would be better than ``Needs `I-{team}-nominated`?`` which feels like someone forgot to localize a string :-)

But I am also ok with leaving it as it is. I'm sure nobody reads it anyway :smile:

apiraino:
> The major issue I see is that this PR adds very rust-lang/rust-specific logic to triagebot.

That's true, I agree. However making it _really_ generic is beyond the time I want to allocate on implementing this change (see also my first comment). So if it's not _that_ bad and it's accepted, great, otherwise I'll have to drop the mic on this :smile:

Mark-Simulacrum (approved):

src/handlers/notify_zulip.rs
```diff
@@ -172,3 +173,66 @@ pub(super) async fn handle_input<'a>(
 
     Ok(())
 }
+
+fn replace_team_to_be_nominated(labels: &Vec<Label>, msg: String) -> String {
```
fmease, line 177:
```suggestion
fn replace_team_to_be_nominated(labels: &[Label], msg: String) -> String {
```

src/handlers/notify_zulip.rs
```diff
@@ -172,3 +173,66 @@ pub(super) async fn handle_input<'a>(
...
+        .map(|label| &label.name)
+        .filter_map(|label| label.strip_prefix("T-"))
+        .collect::<Vec<&str>>();
+
+    // - If a single team label is found, replace the placeholder with that one
+    // - If multiple team labels are found and one of them is "compiler", pick that one
+    //   (currently the only team handling these Zulip notification)
+    // - else, do nothing
+    if teams.len() == 1 {
+        msg.replace("{team}", teams[0])
```
fmease, line 189:
```suggestion
    if let [team] = &*teams {
        msg.replace("{team}", team)
```

src/handlers/notify_zulip.rs
```diff
@@ -172,3 +173,66 @@ pub(super) async fn handle_input<'a>(
...
+    } else if teams.contains(&"compiler") {
+        msg.replace("{team}", "compiler")
+    } else {
+        msg
+    }
+}
+
+#[test]
+fn test_notification() {
+    let mut msg = replace_team_to_be_nominated(&vec![], "Needs `I-{team}-nominated`?".to_string());
```
fmease, line 199:
```suggestion
    let mut msg = replace_team_to_be_nominated(&[], "Needs `I-{team}-nominated`?".to_string());
```

ditto et seqq.

fmease (commented):
small stylistic nits, not blocking. sorry, forgot to submit a code review earlier.

src/handlers/notify_zulip.rs
apiraino: nit: I find the suggested syntax a bit obscure. What is the difference between the two versions? IIUC avoiding a `.len()` saves a little memory? Like one an allocation less? (trying to learn something :slightly_smiling_face: )

apiraino:
applied the suggested nits!

thanks @fmease for your time :)

src/handlers/notify_zulip.rs
fmease: I see what you mean (esp. the `&*` which I sometimes replace with `.as_slice()` for readability).

It's not about allocations but panicking. By using a slice pattern, one can quickly see that this code cannot panic (unless `-Zoom=panic` ofc ^^) but with the check and the explicit indexing, it's “less obvious”. In this case, admittedly it doesn't make a huge difference but it's a nice pattern to follow. It allows for [local reasoning](https://stackoverflow.com/questions/61207079/what-is-local-reasoning-exactly). You wouldn't write `if x.is_some() { x.unwrap() }` either for example.

fmease (approved):

src/handlers/notify_zulip.rs
apiraino: awesome, TIL! Thanks for the additional context

fmease:
needs rebase

ehuss (approved):
Thanks! I'm going to go ahead and merge based on previous reviews, and since this seems pretty simple.

@apiraino Do you mind following up and post an update to the documentation at https://github.com/rust-lang/rust-forge/blob/master/src/triagebot/zulip-notifications.md to add a brief mention of this?

PR #1767, opened 2024-01-30
Don't panic when there is no zulip token

WaffleLapkin:
triagebot currently uses `env::var().unwrap()` inside code that handles zulip things, which is not ideal (and leads to panics spamming logs for people like me, who run triagebot themselves).

I'm not _happy_ with this PR, but it probably solves the issue...

Ideally there would be a better system for enabling/disabling subsystems for the bot (via a per-instance-config and/or env vars), but I do not have energy to work on anything like that.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -25,6 +25,7 @@ lazy_static::lazy_static! {
     };
 }
 
+#[derive(Clone)]
 pub struct ClientPool {
     connections: Arc<Mutex<Vec<tokio_postgres::Client>>>,
     permits: Arc<Semaphore>,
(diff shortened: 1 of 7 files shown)


jackh726 (commented):
Don't really like the indentation changes here. Can probably make this a bit cleaner by either splitting the inner bits into new functions,  or moving the logic further up.

I do like the ZulipContext change though, happy to review that separately if you want to get that landed.

Urgau:
@WaffleLapkin, still interested in this change? If so, can you address Jack comment.

WaffleLapkin:
@Urgau no, not really. Feel free to take these changes if you want.

PR #1768, opened 2024-02-06
Add config options to change labels for draft/closed/reopened PRs

WaffleLapkin:
Adds the following options:

```toml
[autolabel."<label>"]
reopened_pr = true
new_draft_pr = true
reopened_draft_pr = true

[converted-to-draft]
remove_labels = []
add_labels = []

[ready-for-review]
remove_labels = []
add_labels = []

[closed-pr]
remove_labels = []
add_labels = []
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -32,6 +32,8 @@ pub(crate) struct Config {
     pub(crate) github_releases: Option<GitHubReleasesConfig>,
     pub(crate) review_submitted: Option<ReviewSubmittedConfig>,
     pub(crate) review_requested: Option<ReviewRequestedConfig>,
+    pub(crate) converted_to_draft: Option<ConvertedToDraftConfig>,
+    pub(crate) ready_for_review: Option<ReadyForReviewConfig>,
     pub(crate) shortcut: Option<ShortcutConfig>,
     pub(crate) note: Option<NoteConfig>,
     pub(crate) mentions: Option<MentionsConfig>,
@@ -209,6 +211,8 @@ pub(crate) struct AutolabelLabelConfig {
     #[serde(default)]
     pub(crate) new_pr: bool,
     #[serde(default)]
+    pub(crate) new_draft_pr: bool,
+    #[serde(default)]
     pub(crate) new_issue: bool,
 }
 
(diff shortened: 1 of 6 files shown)


WaffleLapkin:
This actually makes me think that using `label <-> filters` map makes little sense. Maybe a `label actions, filters` would be more appropriate?

Something like

```toml
[[autolabel]]
new_pr = true
add = ["S-waiting-on-review"]

[[autolabel]]
review-submitted = true
reviewed_label = "S-waiting-on-author"
review_labels = ["S-waiting-on-review"]

[[autolabel]]
review-requested = true
remove_labels = ["S-waiting-on-author"]
add_labels = ["S-waiting-on-review"]
```

Instead of

```toml
[autolabel."S-waiting-on-review"]
new_pr = true

[review-submitted]
reviewed_label = "S-waiting-on-author"
review_labels = ["S-waiting-on-review"]

[review-requested]
remove_labels = ["S-waiting-on-author"]
add_labels = ["S-waiting-on-review"]
```

The advantages are:
1. Doesn't require new systems for every event
2. Can bind multiple labels for the same events
3. Can bind labels to multiple events at the same time

Example of the last point:
```toml
[[autolabel]]
# Positive filters require any positive filter to fire, i.e. this is triggered on either event
review-requested = true
ready-for-review = true

add = ["S-waiting-on-review"]
remove = ["S-waiting-on-author"]
```

WaffleLapkin:
The only question is how to make this in a backwards compatible way...

jackh726:
To give feedback on the best approach, can you explain a bit on what you *want* from this?

Urgau:
@WaffleLapkin, are you still interested in moving this change forward?

Urgau:
> Currently they don't have any special handling. This is a bit awkward because drafts are marked as `S-waiting-on-review`, etc.

This is no longer the case since https://github.com/rust-lang/triagebot/pull/1968, we don't autolabel, nor auto-assign draft PRs when opened. We do now respect ready for review.

I don't think we have handling yet for "PRs converted to drafts", what would expect us to do in those cases?

Kobzol:
I would personally keep the assignee when converting to draft, but switch the labels.

Urgau:
@WaffleLapkin with https://github.com/rust-lang/triagebot/pull/2104 we now remove `S-waiting-on-review` and add `S-waiting-on-author` when converting to draft.

PR #1769, opened 2024-02-06
Explain that the assignee is not necessarily the reviewer

oli-obk:
Also give "one week" as the timeframe within any reaction (reassignment, "will take longer to review", or a review) will happen. Happy to increase the time here.

Related discussion: https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/review.20queue.20.26.20capacity

cc @rust-lang/compiler

PR #1770, opened 2024-02-08
Fix newline in message

oli-obk:
The hard newline renders badly and wasn't intended anyway. Follow up to https://github.com/rust-lang/triagebot/pull/1769


compiler-errors (approved):

ehuss (approved):
Thanks!

PR #1772, opened 2024-02-14
Update the RDS root CA list

Mark-Simulacrum:
The current root is expiring in a few months, so we need to migrate to a new one. We'll be copying similar code to perf, but we can start with making sure it works with triagebot.

I've checked that the new CA file contains the old certificate, so this should keep working with our current database (i.e. doesn't need to be synchronized deployment wise with anything).


ehuss (approved):

Mark-Simulacrum:
Merging - I have some bandwidth tonight to fix problems if we notice them, so will monitor for a bit.

PR #1773, opened 2024-02-20
Workflow for tracking PRs assignment

apiraino:
This is the third try :sweat_smile: to land part of the work on #1753 (pull requests tracking and balanced assignment for the Rust project contributors).

This patch implements:
- Added a new DB table with the fields to track how many PRs are assigned to a contributor
- Initial DB table population with a one-off job, manually run.

r? @jackh726

### DIFF
--- a/github-graphql/PullRequestsOpen.gql
+++ b/github-graphql/PullRequestsOpen.gql
@@ -1,0 +1,26 @@
+query PullRequestsOpen ($repo_owner: String!, $repo_name: String!, $after: String) {
+  repository(owner: $repo_owner, name: $repo_name) {
+    pullRequests(first: 100, after: $after, states:OPEN) {
+      pageInfo {
+        hasNextPage
+        endCursor
+      }
+      nodes {
+        number
+        updatedAt
+        createdAt
+        assignees(first: 10) {
+          nodes {
+            login
+            databaseId
+          }
+        }
+        labels(first:5, orderBy:{field:NAME, direction:DESC}) {
+          nodes {
+            name
+          }
+        }
+      }
+    }
+  }
+}
--- a/github-graphql/README.md
+++ b/github-graphql/README.md
@@ -1,0 +1,31 @@
+# How to use GraphQL with Rust
+
+# GUI Clients (Electron apps)
+
+Use a client to experiment and build your GraphQL query/mutation.
+
+https://insomnia.rest/download
+
+https://docs.usebruno.com
+
+Once you're happy with the result, save your query in a `<query>.gql` file in this directory. It will serve as
+documentation on how to reproduce the Rust boilerplate.
+
+# Cynic CLI
+
+Introspect a schema and save it locally:
+
+```sh
+cynic introspect \
+    -H "User-Agent: cynic/3.4.3" \
+    -H "Authorization: Bearer [GITHUB_TOKEN]" \
+    "https://api.github.com/graphql" \
+    -o schemas/github.graphql
+```
+
+Execute a GraphQL query/mutation and save locally the Rust boilerplate:
+
+``` sh
+cynic querygen --schema schemas/github.graphql --query query.gql
+```
+
--- a/github-graphql/src/lib.rs
+++ b/github-graphql/src/lib.rs
@@ -89,6 +89,7 @@ pub mod queries {
     #[derive(cynic::QueryFragment, Debug)]
     pub struct User {
         pub login: String,
+        pub database_id: Option<i32>,
     }
 
     #[derive(cynic::QueryFragment, Debug)]
@@ -383,5 +384,46 @@ pub mod project_items {
     #[derive(cynic::QueryFragment, Debug)]
     pub struct ProjectV2ItemFieldDateValue {
         pub date: Option<Date>,
+    }
+}
+
+/// Retrieve all pull requests waiting on review from T-compiler
+/// GraphQL query: see file github-graphql/PullRequestsOpen.gql
+pub mod pull_requests_open {
+    use crate::queries::{LabelConnection, PullRequestConnection, UserConnection};
+
+    use super::queries::DateTime;
+    use super::schema;
+
+    #[derive(cynic::QueryVariables, Clone, Debug)]
+    pub struct PullRequestsOpenVariables<'a> {
+        pub repo_owner: &'a str,
+        pub repo_name: &'a str,
+        pub after: Option<String>,
+    }
+
+    #[derive(cynic::QueryFragment, Debug)]
+    #[cynic(graphql_type = "Query", variables = "PullRequestsOpenVariables")]
+    pub struct PullRequestsOpen {
+        #[arguments(owner: $repo_owner, name: $repo_name)]
+        pub repository: Option<Repository>,
+    }
+
+    #[derive(cynic::QueryFragment, Debug)]
+    #[cynic(variables = "PullRequestsOpenVariables")]
+    pub struct Repository {
+        #[arguments(first: 100, after: $after, states: "OPEN")]
+        pub pull_requests: PullRequestConnection,
+    }
+
+    #[derive(cynic::QueryFragment, Debug)]
+    pub struct PullRequest {
+        pub number: i32,
+        pub updated_at: DateTime,
+        pub created_at: DateTime,
+        #[arguments(first: 10)]
+        pub assignees: UserConnection,
+        #[arguments(first: 5, orderBy: { direction: "DESC", field: "NAME" })]
+        pub labels: Option<LabelConnection>,
     }
 }
--- a/src/github.rs
+++ b/src/github.rs
@@ -2592,6 +2592,90 @@ async fn project_items_by_status(
     Ok(all_items)
 }
 
+/// Retrieve all pull requests in status OPEN that are not drafts
+pub async fn retrieve_pull_requests(
+    repo: &Repository,
+    client: &GithubClient,
+) -> anyhow::Result<Vec<(User, i32)>> {
+    use cynic::QueryBuilder;
+    use github_graphql::pull_requests_open::{PullRequestsOpen, PullRequestsOpenVariables};
+
+    let repo_owner = repo.owner();
+    let repo_name = repo.name();
+
+    let mut prs = vec![];
+
+    let mut vars = PullRequestsOpenVariables {
+        repo_owner,
+        repo_name,
+        after: None,
+    };
+    loop {
+        let query = PullRequestsOpen::build(vars.clone());
+        let req = client.post(&client.graphql_url);
+        let req = req.json(&query);
+
+        let data: cynic::GraphQlResponse<PullRequestsOpen> = client.json(req).await?;
+        if let Some(errors) = data.errors {
+            anyhow::bail!("There were graphql errors. {:?}", errors);
+        }
+        let repository = data
+            .data
+            .ok_or_else(|| anyhow::anyhow!("No data returned."))?
+            .repository
+            .ok_or_else(|| anyhow::anyhow!("No repository."))?;
+        prs.extend(repository.pull_requests.nodes);
+
+        let page_info = repository.pull_requests.page_info;
+        if !page_info.has_next_page || page_info.end_cursor.is_none() {
+            break;
+        }
+        vars.after = page_info.end_cursor;
+    }
+
+    let mut prs_processed: Vec<_> = vec![];
+    let _: Vec<_> = prs
+        .into_iter()
+        .filter_map(|pr| {
+            if pr.is_draft {
+                return None;
+            }
+
+            // exclude rollup PRs
+            let labels = pr
+                .labels
+                .map(|l| l.nodes)
+                .unwrap_or_default()
+                .into_iter()
+                .map(|node| node.name)
+                .collect::<Vec<_>>();
+            if labels.iter().any(|label| label == "rollup") {
+                return None;
+            }
+
+            let _: Vec<_> = pr
+                .assignees
+                .nodes
+                .iter()
+                .map(|user| {
+                    let user_id = user.database_id.expect("checked") as u64;
+                    prs_processed.push((
+                        User {
+                            login: user.login.clone(),
+                            id: Some(user_id),
+                        },
+                        pr.number,
+                    ));
+                })
+                .collect();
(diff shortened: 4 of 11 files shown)


src/main.rs
```diff
@@ -261,6 +262,12 @@ async fn run_server(addr: SocketAddr) -> anyhow::Result<()> {
         octocrab: oc,
     });
 
+    // Run all jobs that don't have a schedule (one-off jobs)
+    if !is_scheduled_jobs_disabled() {
+        spawn_job_oneoffs(ctx.clone());
+    }
```
apiraino, line 268: as a stopgap solution I've put the job trigger here

src/main.rs
```diff
@@ -310,6 +317,38 @@ async fn run_server(addr: SocketAddr) -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Spawns a background tokio task which runs all jobs having no schedule
+/// i.e. manually started
+fn spawn_job_oneoffs(ctx: Arc<Context>) {
+    // TODO: Ideally JobSchedule.schedule should become an `Option<Schedule>`
+    // and here we run all those with schedule=None
```
apiraino, line 324: @jackh726 annoted here your suggestion on how to run non-scheduled job (for a separate patch)

apiraino:
The job starts but GH returns an error, so I'll check what's happening :slightly_smiling_face:

github-graphql/PullRequestsOpen.gql
```diff
@@ -0,0 +1,26 @@
+query PullRequestsOpen ($repo_owner: String!, $repo_name: String!, $after: String) {
+  repository(owner: $repo_owner, name: $repo_name) {
+    pullRequests(first: 100, after: $after, states:OPEN, labels: ["S-waiting-on-review","T-compiler"]) {
```
jackh726, line 3: This should not be limited to these labels.

github-graphql/PullRequestsOpen.gql
```diff
@@ -0,0 +1,26 @@
...
+        number
+        updatedAt
+        createdAt
+        assignees(first: 10) {
+          nodes {
+            login
+            id
+          }
+        }
+        labels(first:5, orderBy:{field:NAME, direction:DESC}) {
```
jackh726, line 18: Looking at this, I assume you later use this to filter labels, but I don't think we should do that.

github-graphql/src/lib.rs
```diff
@@ -385,3 +386,44 @@ pub mod project_items {
...
+    #[cynic(graphql_type = "Query", variables = "PullRequestsOpenVariables")]
+    pub struct PullRequestsOpen {
+        #[arguments(owner: $repo_owner, name: $repo_name)]
+        pub repository: Option<Repository>,
+    }
+
+    #[derive(cynic::QueryFragment, Debug)]
+    #[cynic(variables = "PullRequestsOpenVariables")]
+    pub struct Repository {
+        #[arguments(first: 100, after: $after, states: "OPEN", labels: ["S-waiting-on-review","T-compiler"])]
```
jackh726, line 415: Are these needed? Seems redundant with the gql file.

jackh726: Still, like I commented above, I don't think we should limit this.

src/github.rs
```diff
@@ -2592,6 +2592,81 @@ async fn project_items_by_status(
     Ok(all_items)
 }
 
+/// Retrieve all pull requests in status OPEN, not DRAFT, with label: S-waiting-on-review
```
jackh726, line 2595: Same as above.

src/handlers/pull_requests_assignment_update.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    async fn run(&self, ctx: &super::Context, _metadata: &serde_json::Value) -> anyhow::Result<()> {
+        let db = ctx.db.get().await;
+        let gh = &ctx.github;
+
+        tracing::trace!("starting pull_request_assignment_update");
+
+        // delete everything before populating
+        init_table(&db).await?;
```
jackh726, line 22: This should be moved to the end: if we error out, we don't want to lose everything.

src/handlers/pull_requests_assignment_update.rs
```diff
@@ -0,0 +1,68 @@
...
+
+        let rust_repo = gh.repository("rust-lang/rust").await?;
+        let prs = retrieve_pull_requests(&rust_repo, &gh).await?;
+
+        // populate the table
+        for (assignees, pr_num) in prs {
+            for assignee in assignees {
+                let assignee_id = assignee.id.expect("checked");
```
jackh726, line 30: Preexisting (but maybe you can make a separate PR for this?): id on User probably shouldn't be optional.

src/handlers/pull_requests_assignment_update.rs
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// Truncate the review prefs table
+async fn init_table(db: &DbClient) -> anyhow::Result<u64> {
+    let res = db.execute("TRUNCATE review_prefs;", &[]).await?;
```
jackh726, line 42: I don't think this is right. We don't want to delete everything, just the assigned reviews.

src/handlers/pull_requests_assignment_update.rs
```diff
@@ -0,0 +1,68 @@
...
+        .await
+        .context("Insert user DB error")?;
+    Ok(rec)
+}
+
+/// Create a team member work queue
+async fn create_team_member_workqueue(
+    db: &DbClient,
+    user_id: i64,
+    pr: i32,
```
jackh726, line 62: I feel like we should aggregate by user *first*, rather than doing a db request for each PR.

src/main.rs
```diff
@@ -310,6 +317,38 @@ async fn run_server(addr: SocketAddr) -> anyhow::Result<()> {
...
+    let jobs: Vec<Box<dyn Job + Send + Sync>> = vec![Box::new(PullRequestAssignmentUpdate)];
+
+    for job in jobs {
+        let ctx = ctx.clone();
+        task::spawn(async move {
+            // Allow some spacing between starting jobs
+            let mut interval =
```
jackh726, line 331: Shouldn't this be moved outside the loop? So each job is delayed between each other?

src/main.rs
jackh726: Hmm, not quite what I was thinking, but this is good!

src/main.rs
```diff
@@ -261,6 +262,12 @@ async fn run_server(addr: SocketAddr) -> anyhow::Result<()> {
         octocrab: oc,
     });
 
+    // Run all jobs that don't have a schedule (one-off jobs)
+    if !is_scheduled_jobs_disabled() {
```
jackh726, line 266: Hmm, what is this function

jackh726: Oh, its preexisting. Okay.

github-graphql/PullRequestsOpen.gql
apiraino: hmm you're right, we should pull **all** open PRs here. Don't know what I was thinking...

github-graphql/PullRequestsOpen.gql
apiraino: right, currently I don't use them but I thought it could make sense to pull them anyway in the query (for future use?). Would it be ok to pull them but **not** use them for filtering?

src/handlers/pull_requests_assignment_update.rs
apiraino: yeah this is something that confuses me, don't know exactly why our https://github.com/rust-lang/triagebot/blob/c52016ab56eddb5ce2ac51a7bde6a47b8653719a/src/github.rs#L19-L23 has the Id optional :shrug: 

I agree to remove the `Option` is a separate PR.

src/handlers/pull_requests_assignment_update.rs
apiraino: do you suggest to simply delete all assignments? What if we have GH users that left team? I am not sure the `users` table of the triagebot is ever purged (didn't check yet).

My reasoning is: when invoking a job to initialize the table I thought it was easier to start from a clean slate. This `init_table` should only be called when the intent is to initialize the table.

No strong opinion, though, about just deleting the assignments. It will just require adding a bit of logic to `create_team_member_workqueue` to turn it into an `upsert`

github-graphql/PullRequestsOpen.gql
jackh726: That's fine

src/handlers/pull_requests_assignment_update.rs
jackh726: Well, I think this should at least be after the web requests. If those fail for whatever reason, we shouldn't just lose the data (even though it's easier to get back).

src/handlers/pull_requests_assignment_update.rs
jackh726: The table is also supposed to store preferences, right? If we delete table, preferences are lost.

I guess we could periodically cleanup users, but don't think it's that big of an issue at the moment.

github-graphql/README.md
```diff
@@ -0,0 +1,30 @@
+# How to use GraphQL with Rust
```
apiraino, line 1: I will note down is a separate issue to make a little refactor of the GraphQL code. It has grown a bit organically but now I think we can establish some guidelines on how to do that

src/github.rs
```diff
@@ -2592,6 +2592,93 @@ async fn project_items_by_status(
...
+            let labels = pr
+                .labels
+                .map(|l| l.nodes)
+                .unwrap_or_default()
+                .into_iter()
+                .map(|node| node.name)
+                .collect::<Vec<_>>();
+            if labels.iter().any(|label| label == "rollup") {
+                return None;
+            }
```
apiraino, line 2655: maybe I've found a good use for the labels we pull in this query: exclude the rollups because they don't follow the same lifecycle of other PRs

src/handlers/pull_requests_assignment_update.rs
```diff
@@ -0,0 +1,70 @@
...
+/// Truncate the review prefs table
+async fn init_table(db: &DbClient) -> anyhow::Result<u64> {
+    let res = db
+        .execute("UPDATE review_prefs SET assigned_prs='{}';", &[])
+        .await?;
+    Ok(res)
+}
+
+/// Add a new user (if not existing)
+async fn ensure_team_member(db: &DbClient, user_id: i64, username: &str) -> anyhow::Result<u64> {
```
jackh726, line 49: nit: we should probably de-duplicate this with the existing `notifications::record_username`

jackh726 (commented):
Okay, seems very close. There are still a couple open comments from me and a couple new ones.

I don't think anything here is *blocking*. If you can get to it now, good. Otherwise, a TODO will be okay.

apiraino:
> Okay, seems very close. There are still a couple open comments from me and a couple new ones.
> 
> I don't think anything here is _blocking_. If you can get to it now, good. Otherwise, a TODO will be okay.

just e bit more time, I'm working on resolving the open comments because they do make sense :)

src/github.rs
apiraino: (good point)

src/handlers/pull_requests_assignment_update.rs
```diff
@@ -0,0 +1,72 @@
...
+        init_table(&db).await?;
+
+        // aggregate by user first
+        let aggregated = prs.into_iter().fold(HashMap::new(), |mut acc, (user, pr)| {
+            let (_, prs) = acc
+                .entry(user.id.unwrap())
+                .or_insert_with(|| (user, Vec::new()));
+            prs.push(pr);
+            acc
+        });
```
apiraino, line 37: now pull requests are aggregated by user before executing queries (this snippet was a bit tricky to get right, I was helped!)

src/handlers/pull_requests_assignment_update.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        // populate the table
+        for (_user_id, (assignee, prs)) in &aggregated {
+            let assignee_id = assignee.id.expect("checked");
+            let _ = record_username(&db, assignee_id, &assignee.login).await;
```
apiraino, line 42: I am now using `record_username` and removed the deuplicate function. I've slightly changed the signature of `record_username`.

src/handlers/pull_requests_assignment_update.rs
jackh726: This looks right, but in the future I find it much easier to just use an imperative style (for loop) with a mutable map.

src/main.rs
apiraino: ok, now the job runner should space a bit each job execution. Added a comment to clarify that there is no guarantee they start in sequence and that the delay is precise. But I think it should do.

apiraino:
I think all the comments are now resolved :+1: 

@rustbot ready

src/main.rs
jackh726: I'm not going to block this PR on it, but this should be able to be in the same if block as below. Please fix next PR

src/main.rs
```diff
@@ -310,6 +319,35 @@ async fn run_server(addr: SocketAddr) -> anyhow::Result<()> {
...
+/// i.e. manually executed at the end of the triagebot startup
+// - jobs are not guaranteed to start in sequence (care is to be taken to ensure thet are completely independent one from the other)
+// - the delay between jobs start is not guaranteed to be precise
+async fn spawn_job_oneoffs(ctx: Arc<Context>) {
+    let jobs: Vec<Box<dyn Job + Send + Sync>> = vec![Box::new(PullRequestAssignmentUpdate)];
+
+    for (idx, job) in jobs.into_iter().enumerate() {
+        let ctx = ctx.clone();
+        task::spawn(async move {
+            // Allow some spacing between starting jobs
```
jackh726, line 332: Also not going to block on this, but seems like we could also just spawn one task that runs through all the jobs. Or just not worry about the delay (we can always add it later if it's a problem).

jackh726 (approved):
A couple comments that I'd like for you to followup on in another PR, but let's land this.

PR #1780, opened 2024-03-02
Add an issue transfer command

### LINKED ISSUE #166: Support `@rustbot move $repo`
Does GitHub support this?
- https://developer.github.com/changes/2017-11-09-repository-transfer-api-preview/

https://discordapp.com/channels/442252698964721669/443148290352218115/624290071985782784

ehuss:
This adds a command to transfer an issue from one repository to another. People occasionally file issues in the wrong repo, and it is convenient to transfer them. Unfortunately you need write access to *both* repositories to transfer. Often that isn't the case (like transferring between cargo and clippy or vice-versa). 

The command is `@rustbot transfer reponame`. The source repository needs a `[transfer]` table to enable the feature. 

It is restricted to any team member. I'm not sure if that's the right balance of permissions, but I think it is a good starting place, since this is a fairly destructive command.

Unfortunately due to the way the API works, there is no visual indication that the issue has transferred (unlike the Web UI button, which shows a big notice). I think we'll just need to emphasize in the documentation that after the running the command you need to manually reload the page to view it in the new location.

Closes #166

### DIFF
--- a/parser/src/command.rs
+++ b/parser/src/command.rs
@@ -13,6 +13,7 @@ pub mod prioritize;
 pub mod relabel;
 pub mod second;
 pub mod shortcut;
+pub mod transfer;
 
 #[derive(Debug, PartialEq)]
 pub enum Command<'a> {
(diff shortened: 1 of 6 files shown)


ehuss:
The purpose of this PR is to explicitly allow people who don't have write permissions in both repos to use it. If they already have write permissions, they can use the UI to transfer.

jackh726:
Right - my thought is that to start it might be useful for only those people to use this for now (so they can fix things if something goes wrong). But maybe I'm just being to cautious :)

jackh726:
I can merge this with a rebase

ehuss:
> I can merge this with a rebase

Thanks, done!

jackh726 (approved):

PR #1781, opened 2024-03-06
Add webhook handler to update PR workload queues

apiraino:
This is the third step to implement tracking the pull request workload designed in #1753

In this patch we enable handling GH webhooks when a PR is assigned or unassigned.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -15,6 +15,10 @@ lazy_static::lazy_static! {
         RwLock::new(HashMap::new());
 }
 
+// This struct maps each possible option of the triagebot.toml.
+// See documentation of options at: https://forge.rust-lang.org/triagebot/pr-assignment.html#configuration
+// When adding a new config option to the triagebot.toml, it must be also mapped here
+// Will be used by the `issue_handlers!()` or `command_handlers!()` macros.
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
 #[serde(rename_all = "kebab-case")]
 #[serde(deny_unknown_fields)]
@@ -39,6 +43,7 @@ pub(crate) struct Config {
     // We want this validation to run even without the entry in the config file
     #[serde(default = "ValidateConfig::default")]
     pub(crate) validate_config: Option<ValidateConfig>,
+    pub(crate) pr_tracking: Option<ReviewPrefsConfig>,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -317,6 +322,12 @@ pub(crate) struct GitHubReleasesConfig {
     pub(crate) changelog_branch: String,
 }
 
+#[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+pub(crate) struct ReviewPrefsConfig {
+    #[serde(default)]
+    _empty: (),
+}
+
 fn get_cached_config(repo: &str) -> Option<Result<Arc<Config>, ConfigurationError>> {
     let cache = CONFIG_CACHE.read().unwrap();
     cache.get(repo).and_then(|(config, fetch_time)| {
@@ -463,6 +474,7 @@ mod tests {
                 mentions: None,
                 no_merges: None,
                 validate_config: Some(ValidateConfig {}),
+                pr_tracking: None,
             }
         );
     }
--- a/src/db.rs
+++ b/src/db.rs
@@ -332,6 +332,7 @@ CREATE table review_prefs (
     assigned_prs INT[] NOT NULL DEFAULT array[]::INT[]
 );",
     "
+CREATE EXTENSION intarray;
 CREATE UNIQUE INDEX review_prefs_user_id ON review_prefs(user_id);
  ",
 ];
--- a/src/github.rs
+++ b/src/github.rs
@@ -286,8 +286,11 @@ pub struct Issue {
     ///
     /// Example: `https://github.com/octocat/Hello-World/pull/1347`
     pub html_url: String,
+    // User performing an `action`
     pub user: User,
     pub labels: Vec<Label>,
+    // Users assigned to the issue/pr after `action` has been performed
+    // These are NOT the same as `IssueEvent.assignee`
     pub assignees: Vec<User>,
     /// Indicator if this is a pull request.
     ///
@@ -954,8 +957,14 @@ pub enum IssuesAction {
     Unpinned,
     Closed,
     Reopened,
-    Assigned,
-    Unassigned,
+    Assigned {
+        /// Github users assigned to the issue / pull request
+        assignee: User,
+    },
+    Unassigned {
+        /// Github users removed from the issue / pull request
+        assignee: User,
+    },
     Labeled {
         /// The label added from the issue
         label: Label,
--- a/src/handlers/pr_tracking.rs
+++ b/src/handlers/pr_tracking.rs
@@ -1,0 +1,111 @@
+//! This module updates the PR workqueue of the Rust project contributors
+//!
+//! Purpose:
+//!
+//! - Adds the PR to the workqueue of one team member (when the PR has been assigned)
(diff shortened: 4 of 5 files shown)


src/db.rs
```diff
@@ -332,6 +332,7 @@ CREATE table review_prefs (
     assigned_prs INT[] NOT NULL DEFAULT array[]::INT[]
 );",
     "
+CREATE EXTENSION intarray;
```
apiraino, line 335: I enabled this Postgres extension to be able to use `sort()` and `uniq()` when working on the arrays. We need to keep these INSERTs idempotent (see comment in `pr_tracking.rs`)

src/handlers/pr_tracking.rs
```diff
@@ -0,0 +1,137 @@
...
+            event.action,
+            IssuesAction::Assigned | IssuesAction::Unassigned
+        )
+    {
+        return Ok(None);
+    }
+    Ok(Some(ReviewPrefsInput {}))
+}
+
+pub(super) async fn handle_input<'a>(
```
apiraino, line 51: I didn't find another way around doing this other than:
- removing the PR from everyone
- adding the PR again to the assignee(s) - if any

We need to handle ~three~ four cases:
- PR is unassigned from everyone
- PR is unassigned to one of possibly many assignee(s) - unlikely but allowed by the API
- PR is assigned to one person
- PR is assigned to multiple assignees at the same time - unlikely but allowed by the API

src/handlers/pr_tracking.rs
```diff
@@ -0,0 +1,137 @@
...
+//! Purpose:
+//!
+//! - Adds the PR to the work queue of one or more team member (when the PR has been assigned)
+//! - Removes the PR from the work queue of one or more team members (when the PR is unassigned or closed)
+//!
+//! Notes:
+//!
+//! - When assigning or unassigning a PR, we only receive the list of current assignee(s)
+//! - In case of a complete PR unassignment the list will be empty (i.e. does not have assignee(s) removed)
+//! - The code in this handler MUST be idempotent because GH emits a webhook trigger for every assignee
```
apiraino, line 12: > ... GH emits a webhook trigger for every assignee

This took me a bit to figure out. I'm not sure if the webhook could be configured differently.

Even we work with a single assignee, I think it's good to test everything to handling multiple assignees.

apiraino:
@jackh726 I've removed a lot of code ^^, now I have a better understanding. PR is ready for a first review when you have time. Thanks!

src/github.rs
```diff
@@ -990,6 +993,8 @@ pub struct IssuesEvent {
     pub repository: Repository,
     /// The GitHub user that triggered the event.
     pub sender: User,
+    /// Assignee affected by this `action`
+    pub assignee: User,
```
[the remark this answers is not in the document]
jackh726: This isn't right. Not every action has an `assignee`. This should be under `Assigned` and `Unassigned` in `IssuesAction`.

src/config.rs
```diff
@@ -39,6 +39,7 @@ pub(crate) struct Config {
     // We want this validation to run even without the entry in the config file
     #[serde(default = "ValidateConfig::default")]
     pub(crate) validate_config: Option<ValidateConfig>,
+    pub(crate) pr_tracking: Option<ReviewPrefsConfig>,
```
jackh726, line 42: What is this for?

apiraino: I wish there was documentation for that!

IIUC it's used to ensure that the `Config` struct maps correctly the content of the `triagebot.toml`. In other words, when we will add `[pr-tracking]` in the `triagebot.toml`, it must be also mapped here (will be used by the `issue_handler!()` macro).

jackh726: Hmm, it's pretty unfortunate it's so hard-coded as an issue handler. I guess it makes sense because we only want to get webhooks for the rust-lang/rust repo, but we can't really support other repos anyways with the current setup.

jackh726: Can you please add a comment about this?

src/github.rs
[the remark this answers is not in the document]
apiraino: I'm sorry, I'm not sure i understand this comment: what do you mean that this `assignee` should go in `IssuesAction`? That's the enum describing all possible actions, right?

My understanding was: when the action is `Assigned` or `Unassigned`, then the `issue.assignee` field is the user object of the action (either they are unassigned the PR or become the new assignee).

Maybe `assignee: Option<User>` since it can be null?

thanks :slightly_smiling_face:

src/config.rs
apiraino: ok, I'll add some wording about that. However the issue handler (well, any handler in that macro) can be enabled by adding a table thing `[something]` in the .toml files. Without that handlers will be just ignored when a webhook payload comes.

(iiuc what you mean)

Supporting other repositories (I assume you mean for PR tracking), I think it should be relatively easy and involve small changes at the DB table level (like adding a field).

src/github.rs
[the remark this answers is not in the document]
jackh726: For example, the `Labeled` and `Unlabeled` variants have a `label` field, because `label` is only present on those actions.

Similarly, `assignee` is only present on `Assigned` and `Unassigned` actions.

`assignee: Option<User>` could work, but is not appropriate; for other actions, we never expect assignee, for example. And for assigned and unassigned we always expect it.

src/config.rs
apiraino: added some wording. hope it clarifies :slightly_smiling_face:

src/github.rs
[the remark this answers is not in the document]
apiraino: ok @jackh726 I've refactored the code as per your suggestion, seems to work.

As an aside the code now looks a bit strange because of this handling of enums variants with structs inside. I have to squint to understand what it does (and I wrote it). Any suggestion on how to improve the readability?

src/handlers/pr_tracking.rs
```diff
@@ -0,0 +1,111 @@
...
+    else {
+        return Ok(());
+    };
+
+    // ensure the team member object of this action exists in the `users` table
+    record_username(&db_client, assignee.id.unwrap(), &assignee.login)
+        .await
+        .context("failed to record username")?;
+
+    if let IssuesAction::Unassigned { assignee: _ } = &event.action {
```
GuillaumeGomez, line 66:
```suggestion
    if matches!(event.action, ssuesAction::Unassigned { .. }) {
```

src/handlers/pr_tracking.rs
```diff
@@ -0,0 +1,111 @@
...
+        delete_pr_from_workqueue(&db_client, assignee.id.unwrap(), event.issue.number)
+            .await
+            .context("Failed to remove PR from workqueue")?;
+    }
+
+    if let IssuesAction::Assigned { assignee: _ } = &event.action {
```
GuillaumeGomez, line 72:
```suggestion
    if matches!(event.action, IssuesAction::Assigned { .. }) {
```

src/handlers/pr_tracking.rs
```diff
@@ -0,0 +1,111 @@
...
+        return Ok(None);
+    }
+
+    // ... and if the action is an assignment or unassignment with an assignee
+    let (IssuesAction::Assigned { assignee: _ } | IssuesAction::Unassigned { assignee: _ }) =
+        &event.action
+    else {
+        return Ok(None);
+    };
+    Ok(Some(ReviewPrefsInput {}))
```
GuillaumeGomez, line 41:
```suggestion
    match event.action {
        IssuesAction::Assigned { .. } | IssuesAction::Unassigned { .. } => Ok(Some(ReviewPrefsInput {})),
        _ => Ok(None),
    }
```

apiraino:
thank you @GuillaumeGomez it looks much better now :)

jackh726 (approved):
Looks good now!

jackh726:
@apiraino can you make a PR to rust-lang/rust to add the config option to the `triagebot.toml`

PR #1783, opened 2024-03-19
Quote review suggestions in comments

apiraino:
This looks like an innocent cosmetic change now but it become critical when we will implement the new review assignment.

In case a pull request review cannot be assigned (for any reason), the triagebot will post a comment on GitHub. The text of the comment does not quote correctly the "r?" wording, example "Use r? to specify someone else to assign.".

That could trigger an infinite loop where the words "r? to" are interpreted by the triagebot receiving the webhook after posting that command.

Quoting with markdown will prevent this loop.

I am not sure this patch solves 100% of the all cases but it should cover most of them.

r? @jackh726

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -61,10 +61,10 @@ const RETURNING_USER_WELCOME_MESSAGE: &str = "r? @{assignee}
 They will have a look at your PR within the next two weeks and either review your PR or \
 reassign to another reviewer.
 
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
We should already be ignoring any messages from triagebot for command parsing, but extra quoting shouldn't hurt.

Mark-Simulacrum (approved):

PR #1784, opened 2024-03-23
Use rfcbot style for FCPs in the prioritization agenda

Urgau:
This PR add the necessary information to replicate the rfcbot style when showing FCPs and uses it in the prioritization agenda.

~~One thing at isn't perfect is that we are using the Github login and not the Zulip login, using the Zulip username would require accessing the team database which I didn't wanted to do yet but could be done as a follow up PR.~~ (fixed, https://github.com/rust-lang/triagebot/pull/1784#issuecomment-2016833448)

r? @apiraino 
 
--------

They would now look like this (modulo the mentions). ~~Full example [here](https://hackmd.io/@Urgau/rytauC3Aa).~~

- merge: [Add a new `--build-id` flag to rustc (compiler-team#635)](https://github.com/rust-lang/compiler-team/issues/635#issuecomment-1777545767)
    * `@wesleywiser` `@davidtwco` `@Aaron1011` `@petrochenkov`
    * [other-existing-options (by petrochenkov)](https://github.com/rust-lang/compiler-team/issues/635#issuecomment-1779597156) [option-name (by wesleywiser)](https://github.com/rust-lang/compiler-team/issues/635#issuecomment-1785349936)
- merge: [Retire the mailing list and make all decisions on zulip (compiler-team#649)](https://github.com/rust-lang/compiler-team/issues/649#issuecomment-1618781780)
    * no pending checkboxs
    * [automatic-sync (by compiler-errors)](https://github.com/rust-lang/compiler-team/issues/649#issuecomment-1618902660) [single-point-of-failure-via-stream-archival (by pnkfelix)](https://github.com/rust-lang/compiler-team/issues/649#issuecomment-1664056989)

### DIFF
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -74,6 +74,7 @@ pub struct FCPDetails {
     pub initiating_comment_html_url: String,
     pub initiating_comment_content: String,
     pub disposition: String,
+    pub should_mention: bool,
     pub pending_reviewers: Vec<FCPReviewerDetails>,
     pub concerns: Vec<FCPConcernDetails>,
 }
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -9,6 +9,7 @@ use tera::{Context, Tera};
 use crate::{
     github::{self, GithubClient, Repository},
     http_client::{CompilerMeeting, HttpClient},
+    rfcbot,
 };
 
 #[async_trait]
@@ -55,11 +56,21 @@ pub struct IssueDecorator {
 }
 
 #[derive(Serialize, Deserialize, Debug)]
+pub struct FCPConcernDetails {
+    pub name: String,
+    pub reviewer_login: String,
+    pub concern_url: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
 pub struct FCPDetails {
     pub bot_tracking_comment_html_url: String,
     pub bot_tracking_comment_content: String,
     pub initiating_comment_html_url: String,
     pub initiating_comment_content: String,
+    pub disposition: String,
+    pub pending_reviewers: Vec<rfcbot::Reviewer>,
+    pub concerns: Vec<FCPConcernDetails>,
 }
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -130,6 +141,7 @@ impl<'a> Action for Step<'a> {
                     let query = query.clone();
                     handles.push(tokio::task::spawn(async move {
                         let _permit = semaphore.acquire().await?;
+                        let fcps_groups = ["proposed_fcp", "in_pre_fcp", "in_fcp"];
                         let mcps_groups = [
                             "mcp_new_not_seconded",
                             "mcp_old_not_seconded",
@@ -140,7 +152,7 @@ impl<'a> Action for Step<'a> {
                         let issues = query
                             .query(
                                 &repository,
-                                name == "proposed_fcp",
+                                fcps_groups.contains(&name.as_str()),
                                 mcps_groups.contains(&name.as_str())
                                     && repository.full_name.contains("rust-lang/compiler-team"),
                                 &gh,
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -9,7 +9,6 @@ use tera::{Context, Tera};
 use crate::{
     github::{self, GithubClient, Repository},
     http_client::{CompilerMeeting, HttpClient},
-    rfcbot,
 };
 
 #[async_trait]
@@ -63,13 +62,19 @@ pub struct FCPConcernDetails {
 }
 
 #[derive(Serialize, Deserialize, Debug)]
+pub struct FCPReviewerDetails {
+    pub github_login: String,
+    pub zulip_id: Option<u64>,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
 pub struct FCPDetails {
     pub bot_tracking_comment_html_url: String,
     pub bot_tracking_comment_content: String,
     pub initiating_comment_html_url: String,
     pub initiating_comment_content: String,
     pub disposition: String,
-    pub pending_reviewers: Vec<rfcbot::Reviewer>,
+    pub pending_reviewers: Vec<FCPReviewerDetails>,
     pub concerns: Vec<FCPConcernDetails>,
 }
 
--- a/src/github.rs
+++ b/src/github.rs
@@ -1712,7 +1712,9 @@ impl<'q> IssuesQuery for Query<'q> {
             .with_context(|| "Unable to get issues.")?;
 
         let fcp_map = if include_fcp_details {
-            crate::rfcbot::get_all_fcps().await?
+            crate::rfcbot::get_all_fcps()
+                .await
+                .with_context(|| "Unable to get all fcps from rfcbot.")?
         } else {
             HashMap::new()
         };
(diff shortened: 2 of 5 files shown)


Urgau:
I've found a pretty neat trick. Zulip as a de-ambiguator syntax when pinging some, it works by specifying the Zulip ID after the name, like so `@**Urgau|123456**`; but if we omit the name and just provide the ID it will ping the right person. Simplifying the greatly the process.

I've added this is the last commit.

src/actions.rs
```diff
@@ -140,7 +156,7 @@ impl<'a> Action for Step<'a> {
                         let issues = query
                             .query(
                                 &repository,
-                                name == "proposed_fcp",
+                                name.contains("fcp"),
```
apiraino, line 159: I would probably revert this change. It will run an additional HTTP query also for FCPs that don't need the additional details. In fact we're interested to show these details only for FCPs with label `proposed-final-comment-period`.

src/github.rs
```diff
@@ -1750,6 +1758,42 @@ impl<'q> IssuesQuery for Query<'q> {
                         bot_tracking_comment_content,
                         initiating_comment_html_url: init_comment.html_url.clone(),
                         initiating_comment_content: quote_reply(&init_comment.body),
+                        disposition: fcp
+                            .fcp
+                            .disposition
+                            .as_deref()
+                            .unwrap_or("<unknown>")
+                            .to_string(),
+                        pending_reviewers: fcp
```
apiraino, line 1767: TIL using the Zulip ID for pings, that's a neat trick! 

Though I think in this context we don't want to emit ping every week, I'm afraid it will just become annoying. I would rather instead remove the `@` mentions and just list the people the FCP is waiting on.

templates/prioritization_agenda.tt
apiraino, line 1: I think the new Jinja template needs some tweaks to look like the current Markdown we produce (a bit more compact).

Also, the changes to the headings in this template `prioritization_agenda.tt` could probably be reverted (again, to keep the template as compact as possible)

apiraino (changes requested):
hey @Urgau thanks for working on this! :slightly_smiling_face: 

I like the new layout you suggest for the FCPs.

I left a few comments (mostly nits), let me know what you think!

templates/prioritization_agenda.tt
Urgau: IMO the current template is already too compact and difficult to read. With this new style it became unreadable. I can try to make it more compact but it's already very compact.

src/github.rs
Urgau: We may not want to ping every week but I think we should ping from time to time, and I think pinging T-compiler member in the meeting at the specific things waiting for them is acceptable to me.

Especially considering that they are already being pinging at the start of the meeting.

templates/prioritization_agenda.tt
Urgau: After some discussions on Zulip, I decided to revert those stylistic changes.

Urgau: One thing I just saw is that those bullet points are now so long that Zulip is collapsing the message by default. This is unfortunate.

src/github.rs
Urgau: After some careful thinking I think I may have found a compromise.

I propose that we only mention under those circumstance:
  - we are 2 weeks into the FCP *(even if the are concerns registered, so that team members can voice there opinions on the FCP and concerns)*
  - or when there are no concerns and we are at least 4 weeks into the FCP *(no concerns + long after the start: probably didn't see the ping or forgot; and if the are not happy about the proposal they can just put up a concern)*

Btw, I used your `_` trick to do the no-but-still-looking-mention.

apiraino:
Enabling mentions to remind team members checking their boxes in FCPs could have been a follow-up PR focused because it is unclear how it should work. However, for the sake of merging this change, I will approve this PR. For now it is disabled and will be discussed with T-compiler if and how that could work.

Thanks @Urgau for suggesting this formatting for the FCPs.

r=me

templates/prioritization_agenda.tt
```diff
@@ -28,9 +29,9 @@ note_id: xxx
 - Old MCPs (not seconded, take a look)
 {{-issues::render(issues=mcp_old_not_seconded, indent="  ", with_age=true, empty="No old proposals this time.")}}
 - Pending FCP requests (check your boxes!)
-{{-issues::render(issues=in_pre_fcp, indent="  ", empty="No pending FCP requests this time.")}}
+{{-issues_rfcbot::render(issues=in_pre_fcp, indent="  ", empty="No pending FCP requests this time.")}}
 - Things in FCP (make sure you're good with it)
```
apiraino, line 33: @Urgau The items in this section lost the links to the links to the pull requests

probably worth checking if your new macro is not eating other links around

apiraino (changes requested):

templates/prioritization_agenda.tt
Urgau: Sorry for that, based on the extra information you gave in DM, I was able to fix it.

To make in short, when we don't have any details for an FCP we use a fallback and that fallback was manual instead of using the `_issue.tt` template 🤦

I pushed a commit using the proper fallback *as well as fixing a weird newline issue*.

src/github.rs
```diff
@@ -1745,11 +1755,60 @@ impl<'q> IssuesQuery for Query<'q> {
...
+                    //  - or when there are no concerns and we are at least 4 weeks into the FCP.
+                    //
+                    // FIXME: This should get T-compiler approval before being enabled by default
+                    let should_mention = std::env::var("TRIAGEBOT_COMPILER_MENTION").is_ok() && {
+                        let now = chrono::offset::Utc::now();
+                        let time_diff = now - init_comment.created_at;
+                        time_diff.num_weeks() == 2
+                            || (time_diff.num_weeks() >= 4 && fcp.concerns.is_empty())
+                    };
+
```
apiraino, line 1769:
```suggestion
                    // TODO: agree with the team(s) a policy to emit actual mentions to remind FCP
                    // voting member to cast their vote
                    let should_mention = false;
```

apiraino: @Urgau can we disable this for the moment and keep it as a reminder (which is useful)

Urgau: I applied your patch *(which removes, not disable, it was already disabled)*.

I hope we can get a policy on this in the near future as IMO not having mention doesn't give much benefits.

apiraino: yep, that's on my radar, don't worry :slightly_smiling_face: as I mentioned I just think that that needs a brief discussion involving people on the other end of these mentions and I don't want your work here to be blocked on that

apiraino:
r=me

apiraino (approved):

apiraino:
@jackh726 we're done here, I think. When you have a sec feel free to have a look (and ofc leave a comment). thanks!

templates/_issues_rfcbot.tt
```diff
@@ -0,0 +1,15 @@
+{% import "_issue.tt" as issue %}
+
+{% macro render(issues, indent="", empty="None.") %}
+{%- for issue in issues %}
+{%- if issue.fcp_details is object %}
+{{indent}}- {{issue.fcp_details.disposition}}: [{{issue.title}} ({{issue.repo_name}}#{{issue.number}})]({{issue.fcp_details.bot_tracking_comment_html_url}})
+{{indent}}{{indent}}-{% for reviewer in issue.fcp_details.pending_reviewers %} @{% if issue.fcp_details.should_mention %}{% else %}_{% endif %}**|{{reviewer.zulip_id}}**{%else%} no pending checkboxs{% endfor %}
+{{indent}}{{indent}}-{% for concern in issue.fcp_details.concerns %} [{{concern.name}} (by {{concern.reviewer_login}})]({{concern.concern_url}}){%else%} no pending concerns{% endfor -%}
```
jackh726, line 8: It might be nice to write something like `concerns: ...`

jackh726 (commented):
One small nit

jackh726 (approved):

templates/_issues_rfcbot.tt
Urgau: Sure, pushed the change.

PR #1786, opened 2024-04-03
Assign PR assignment based on work queue

apiraino:
This implements the new way of assigning pull requests reviews (designed in #1753). When this PR is merged, PRs in `rust-lang/rust` will be assigned based on user preferences. These can be queried from Zulip (see [docs](https://forge.rust-lang.org/triagebot/pr-assignment-tracking.html)).

If no preferences are set, PRs will be just assigned (like always).

There are multiple points where PRs are assigned (via comment or using the GH UI), I've tried to cover all cases, took some time to test them all. The amount of changes reflect that it is not so easy to catch all spots where this happens. I wish I could make a smaller patch but I think this is an _all-or-nothing_ situation where all cases should be handled at once.

Also modified the zulip triagebot command `work show` to return the # work queue capacity set by the user. This is helpful in case a PR assignment is rejected. A PR to update the [documentation](https://forge.rust-lang.org/triagebot/pr-assignment-tracking.html) will follow-up after merge of this.

If a PR assignment is rejected we will return a message with some context. Examples:

Failed PR assignment to a user:
screenshot-20240403-104629

Failed self-assignment:
screenshot-20240403-104717

r? @jackh726

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -7,6 +7,9 @@
 //! * `@rustbot release-assignment`: Removes the commenter's assignment.
 //! * `r? @user`: Assigns to the given user (PRs only).
 //!
+//! Note: this module does not handle review assignments issued from the
+//! GitHub "Assignees" dropdown menu
+//!
 //! This is capable of assigning to any user, even if they do not have write
 //! access to the repo. It does this by fake-assigning the bot and adding a
 //! "claimed by" section to the top-level comment.
@@ -20,7 +23,7 @@
 use crate::{
     config::AssignConfig,
     github::{self, Event, FileDiff, Issue, IssuesAction, Selection},
-    handlers::{Context, GithubClient, IssuesEvent},
+    handlers::{pr_tracking::has_user_capacity, Context, GithubClient, IssuesEvent},
     interactions::EditIssueBody,
 };
 use anyhow::{bail, Context as _};
@@ -30,6 +33,7 @@ use rand::seq::IteratorRandom;
 use rust_team_data::v1::Teams;
 use std::collections::{HashMap, HashSet};
 use std::fmt;
+use tokio_postgres::Client as DbClient;
 use tracing as log;
 
 #[cfg(test)]
@@ -74,6 +78,27 @@ const NON_DEFAULT_BRANCH: &str =
      Please double check that you specified the right target!";
 
 const SUBMODULE_WARNING_MSG: &str = "These commits modify **submodules**.";
+
+pub const SELF_ASSIGN_HAS_NO_CAPACITY: &str = "
+You have insufficient capacity to be assigned the pull request at this time. PR assignment has been reverted.
+
+Please choose another assignee or increase your assignment limit.
+
+(see [documentation](https://forge.rust-lang.org/triagebot/pr-assignment-tracking.html))";
+
+pub const REVIEWER_HAS_NO_CAPACITY: &str = "
+`{username}` has insufficient capacity to be assigned the pull request at this time. PR assignment has been reverted.
+
+Please choose another assignee.
+
+(see [documentation](https://forge.rust-lang.org/triagebot/pr-assignment-tracking.html))";
+
+const NO_REVIEWER_HAS_CAPACITY: &str = "
+Could not find a reviewer with enough capacity to be assigned at this time. This is a problem.
+
+Please contact us on [#t-infra](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra) on Zulip.
+
+cc: @jackh726 @apiraino";
 
 fn on_vacation_msg(user: &str) -> String {
     ON_VACATION_WARNING.replace("{username}", user)
@@ -272,6 +297,8 @@ async fn set_assignee(issue: &Issue, github: &GithubClient, username: &str) {
 /// Determines who to assign the PR to based on either an `r?` command, or
 /// based on which files were modified.
 ///
+/// Will also check if candidates have capacity in their work queue.
+///
 /// Returns `(assignee, from_comment)` where `assignee` is who to assign to
 /// (or None if no assignee could be found). `from_comment` is a boolean
 /// indicating if the assignee came from an `r?` command (it is false if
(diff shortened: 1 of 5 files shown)


src/handlers/assign.rs
```diff
@@ -693,14 +765,57 @@ fn find_reviewer_from_names(
...
+    if filtered_candidates.is_empty() {
+        return Err(FindReviewerError::AllReviewersFiltered {
+            initial: names.to_vec(),
+            filtered: names.to_vec(),
+        });
+    }
+
+    log::debug!("Filtered list of candidates: {:?}", filtered_candidates);
+
+    Ok(filtered_candidates
```
jackh726, line 783: Quickly skimming on my phone. I'm curious how we currently users on vacation? I don't see any logic here or in the function below (but I may be missing it).

src/handlers/assign.rs
```diff
@@ -7,6 +7,9 @@
 //! * `@rustbot release-assignment`: Removes the commenter's assignment.
 //! * `r? @user`: Assigns to the given user (PRs only).
 //!
+//! Note: this module does not handle review assignments issued from the
```
jackh726, line 10: Do you handle this elsewhere?

src/handlers/assign.rs
apiraino: Users on vacation are listed in the [triagebot file](https://github.com/rust-lang/rust/blob/ca7d34efa94afe271accf2bd3d44152a5bd6fff1/triagebot.toml#L771) and then filtered out [here](https://github.com/rust-lang/triagebot/blob/c6fa165f057475e34d5e968c89c25b8962b42a43/src/handlers/assign.rs#L726-L738), just right before (inside `candidate_reviewers_from_names`). Then we apply a second pass and filter out team members after reading their work queue.

My plan is to deprecate the "on vacation" thing altogether and revert the patch that introduced it (https://togithub.com/rust-lang/triagebot/pull/1712). To clarify: the "on vacation" feature is sound but should be handled in a more fine-grained way

src/handlers/assign.rs
apiraino: yes, exactly [here](https://github.com/rust-lang/triagebot/pull/1786/files#diff-63f676f3f26949c577c742ff956f9c696a8a640529b4a268068ac205508e4c11R76). 

At this time we don't handle PR assignment done using the GH web UI, so what happens is a bit ugly; in that piece of code I handle the webhook _post-assignment_ and when the assignee(s) don't have capacity I _revert_ the assignment and emit a comment.

(ideally we should handle those assignments in a webhook but I felt this a topic for another pr :slightly_smiling_face: )

src/handlers/assign.rs
```diff
@@ -75,6 +79,27 @@ const NON_DEFAULT_BRANCH: &str =
...
+Please choose another assignee or if you *really* want `{username}` to review this pull request, please ask them to increase their assignment limit.
+
+(see [documentation](https://forge.rust-lang.org/triagebot/pr-assignment-tracking.html))";
+
+const NO_REVIEWER_HAS_CAPACITY: &str = "
+Could not find a reviewer with enough capacity to be assigned at this time. This is a problem.
+
+Please contact us on [#t-infra](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra) on Zulip.
+
+cc: @Mark-Simulacrum";
```
jackh726, line 101: better to cc me (or you)

src/handlers/assign.rs
```diff
@@ -75,6 +79,27 @@ const NON_DEFAULT_BRANCH: &str =
...
+You have insufficient capacity to be assigned the pull request at this time. PR assignment has been reverted.
+
+Please choose another assignee or increase your assignment limit.
+
+(see [documentation](https://forge.rust-lang.org/triagebot/pr-assignment-tracking.html))";
+
+pub const REVIEWER_HAS_NO_CAPACITY: &str = "
+`{username}` has insufficient capacity to be assigned the pull request at this time. PR assignment has been reverted.
+
+Please choose another assignee or if you *really* want `{username}` to review this pull request, please ask them to increase their assignment limit.
```
jackh726, line 92: I think just leaving it at "Please choose another assignee" is fine

src/lib.rs
```diff
@@ -132,17 +132,26 @@ pub struct ReviewPrefs {
...
+    pub max_assigned_prs: Option<i32>,
 }
 
 impl ReviewPrefs {
     fn to_string(&self) -> String {
+        let capacity = if self.max_assigned_prs.is_some() {
+            format!("{}", self.max_assigned_prs.expect("This can't fail"))
+        } else {
+            String::from("Not set (i.e. unlimited)")
+        };
```
jackh726, line 144: Why not just
```rust
let capacity = match self.max_assigned_prs {
  Some(max) =>  format!("{}",max),
  None => String::from("Not set (i.e. unlimited)"),
};
```

jackh726 (commented):
A couple small comments - looks good though. Importantly, this doesn't yet enable people to change their max PRs, which is good. In theory, nothing should change.

apiraino:
@jackh726 I've applied your suggestions, thanks! As discussed, let's leave this here for a moment until we first merge a patch allowing people to set their own workload. Going to prepare it in short.

apiraino:
#1790 is up for review

jackh726:
I think you may have misunderstood. I would like to merge this *before* we enable allowing people to change their setting.

src/handlers/assign.rs
```diff
@@ -75,6 +79,27 @@ const NON_DEFAULT_BRANCH: &str =
...
+Please choose another assignee.
+
+(see [documentation](https://forge.rust-lang.org/triagebot/pr-assignment-tracking.html))";
+
+const NO_REVIEWER_HAS_CAPACITY: &str = "
+Could not find a reviewer with enough capacity to be assigned at this time. This is a problem.
+
+Please contact us on [#t-infra](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra) on Zulip.
+
+cc: @**Jack Huey** @**apiraino**";
```
jackh726, line 101: I don't think this is right for github.

Should be `cc @jackh726 @apiraino`

apiraino: what was I thinking :skull:

apiraino:
rebased in #1877

PR #1791, opened 2024-04-15
Allow a single notify-zulip notification to send multiple Zulip messages

fmease:
This would unblock rust-lang/rust#116957:

**Hard Requirements**:

We would like to

1. **ping** the rustdoc team (via `@*T-rustdoc*`),
2. **link** to the nominated PR (via the short link `#NNN`),
3. create a Zulip **poll**.

**Problems**:

* Unfortunately, neither pings nor short links inside the “question” (title) of a poll get recognized by Zulip.
  Example: [Here](https://rust-lang.zulipchat.com/#narrow/stream/266220-t-rustdoc/topic/Nominate.20.23123905.20for.20beta/near/433147678), you can see that neither `@*T-rustdoc*` nor `#123905` got autolinked.
* Furthermore, it's not possible to sidestep the issue above by moving the ping and the PR ref “out of” the poll and into a separate — preceding — paragraph. If we tried that, Zulip would not detect the poll. This seems to be a known limitation: [Create a poll | Zulip help center § Troubleshooting \[archive link\]](https://web.archive.org/web/20231102162048/https://zulip.com/help/create-a-poll#troubleshooting) ([the live site](https://zulip.com/help/create-a-poll) no longer mentions this explicitly, only implicitly).

**Solution**:

Send two separate Zulip messages. This requires modifications to **triagebot**.

---

This PR enables users to specify a *sequence* of strings for the fields `notify-zulip.⟨label⟩.message_on_⟨add|remove|close|reopen⟩` in order to make triagebot post a sequence of Zulip messages. For backward compatibility, you can still specify a single string. I haven't renamed any of the fields (like `message_on_$` to `messages_on_$`) to avoid churn. A bare string gets treated as a sequence of length 1 containing that very string.

In TypeScript syntax, I've relaxed the type of the optional (!) fields `message_on_$` from `string | null` to `string | string[] | null`.

---

r? @Mark-Simulacrum or infra

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -228,10 +228,26 @@ pub(crate) struct NotifyZulipConfig {
 pub(crate) struct NotifyZulipLabelConfig {
     pub(crate) zulip_stream: u64,
     pub(crate) topic: String,
-    pub(crate) message_on_add: Option<String>,
-    pub(crate) message_on_remove: Option<String>,
-    pub(crate) message_on_close: Option<String>,
-    pub(crate) message_on_reopen: Option<String>,
+    #[serde(rename = "message_on_add", default, deserialize_with = "string_or_seq")]
+    pub(crate) messages_on_add: Vec<String>,
+    #[serde(
+        rename = "message_on_remove",
+        default,
+        deserialize_with = "string_or_seq"
+    )]
+    pub(crate) messages_on_remove: Vec<String>,
+    #[serde(
+        rename = "message_on_close",
+        default,
+        deserialize_with = "string_or_seq"
+    )]
+    pub(crate) messages_on_close: Vec<String>,
+    #[serde(
+        rename = "message_on_reopen",
+        default,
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -228,10 +228,26 @@ pub(crate) struct NotifyZulipConfig {
...
+    )]
+    pub(crate) messages_on_remove: Vec<String>,
+    #[serde(
+        rename = "message_on_close",
+        default,
+        deserialize_with = "string_or_seq"
+    )]
+    pub(crate) messages_on_close: Vec<String>,
+    #[serde(
+        rename = "message_on_reopen",
```
Mark-Simulacrum, line 246: Should all the renames be alias? I guess it doesn't matter for just Deserialize?

Mark-Simulacrum (approved):
Seems fine to me, though I have not yet reviewed in detail.

PR #1792, opened 2024-04-16
Revert "Assign PR assignment based on work queue"

apiraino:
Reverts rust-lang/triagebot#1786

That patch seems to stop PRs being assigned. Reverting while looking into it

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -335,5 +335,4 @@ CREATE table review_prefs (
 CREATE EXTENSION intarray;
 CREATE UNIQUE INDEX review_prefs_user_id ON review_prefs(user_id);
  ",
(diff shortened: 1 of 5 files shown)


src/db.rs
```diff
@@ -335,5 +335,4 @@ CREATE table review_prefs (
 CREATE EXTENSION intarray;
 CREATE UNIQUE INDEX review_prefs_user_id ON review_prefs(user_id);
  ",
-    "ALTER TABLE review_prefs ADD COLUMN max_assigned_prs INTEGER DEFAULT NULL;",
```
Mark-Simulacrum, line 338: This looks iffy, but I think should be ok to revert. We'll need to be careful when reapplying this change.

Mark-Simulacrum (approved):

src/db.rs
apiraino: good point, I'll keep it in mind, thanks

PR #1793, opened 2024-04-16
PR assignment based on work queue availability (take #2)

apiraino:
This is a version of #1786 that fixes the two following bugs:

Fixes 2 bugs:

1. the initial migration wasn't formatted correctly and the new schema changes weren't applied

2. The SELECT to find a reviewer would return wrong results when a team member had NULL in the table `review_prefs.max_assigned_prs`

r? @jackh726

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -332,7 +332,11 @@ CREATE table review_prefs (
     assigned_prs INT[] NOT NULL DEFAULT array[]::INT[]
 );",
     "
-CREATE EXTENSION intarray;
-CREATE UNIQUE INDEX review_prefs_user_id ON review_prefs(user_id);
+CREATE EXTENSION IF NOT EXISTS intarray;",
+    "
+CREATE UNIQUE INDEX IF NOT EXISTS review_prefs_user_id ON review_prefs(user_id);
  ",
+    "
+ALTER TABLE review_prefs ADD COLUMN IF NOT EXISTS max_assigned_prs INTEGER DEFAULT NULL;
+",
 ];
--- a/src/handlers/pr_tracking.rs
+++ b/src/handlers/pr_tracking.rs
@@ -1,18 +1,22 @@
 //! This module updates the PR workqueue of the Rust project contributors
+//! Runs after a PR has been assigned or unassigned
 //!
 //! Purpose:
 //!
-//! - Adds the PR to the workqueue of one team member (when the PR has been assigned)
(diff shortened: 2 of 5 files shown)


src/db.rs
```diff
@@ -332,7 +332,11 @@ CREATE table review_prefs (
...
     "
-CREATE EXTENSION intarray;
-CREATE UNIQUE INDEX review_prefs_user_id ON review_prefs(user_id);
+CREATE EXTENSION IF NOT EXISTS intarray;",
+    "
+CREATE UNIQUE INDEX IF NOT EXISTS review_prefs_user_id ON review_prefs(user_id);
  ",
+    "
+ALTER TABLE review_prefs ADD COLUMN IF NOT EXISTS max_assigned_prs INTEGER DEFAULT NULL;
+",
```
apiraino, line 341: The migration should now be correctly formatted.

Adding the clause `IF NOT EXISTS` should address the concern addressed in [this comment](https://github.com/rust-lang/triagebot/pull/1792/commits/70dd15e02a30932af4691f5fc0f020c273faeab6#r1567313497) about re-running them.

src/handlers/pr_tracking.rs
```diff
@@ -66,18 +70,60 @@ pub(super) async fn handle_input<'a>(
...
+pub async fn has_user_capacity(
+    db: &crate::db::PooledClient,
+    assignee: &str,
+) -> anyhow::Result<ReviewPrefs, FindReviewerError> {
+    let q = "
+SELECT username, r.*
+FROM review_prefs r
+JOIN users ON users.user_id = r.user_id
+WHERE username = $1
+AND ( CARDINALITY(r.assigned_prs) < max_assigned_prs OR max_assigned_prs IS NULL );";
```
apiraino, line 117: The old version of this SQL query ([here](https://github.com/rust-lang/triagebot/pull/1786/files#diff-63f676f3f26949c577c742ff956f9c696a8a640529b4a268068ac205508e4c11R117)) would fail to find a reviewer if the candidate had max_assignment_prs was NULL

apiraino (commented):
I'm going to add a bit of documentation about the migrations.

Apart from that, hopefully this should be fine.

jackh726 (approved):

jackh726:
Okay, let's try this again.

jackh726:
For sure if this is still problematic, we need to consider splitting this up.

PR #1794, opened 2024-04-16
Allow ghost GH user to skip PR workqueue check

apiraino:
Follow up to #1793 

This fix should allow the ghost user to skip some checks when being assigned a PR.

We use [ghost](https://github.com/ghost) as a special GH user to "assign to no one" when we just need a PR to go through the CI. Therefore this user needs no work queue check.

r? @jackh726

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -766,16 +766,22 @@ async fn find_reviewer_from_names(
     // These are all ideas for improving the selection here. However, I'm not
     // sure they are really worth the effort.
 
-    // filter out team members without capacity
-    let filtered_candidates = filter_by_capacity(db, &candidates)
-        .await
-        .expect("Error while filtering out team members");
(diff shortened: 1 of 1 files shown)


src/handlers/assign.rs
```diff
@@ -766,16 +766,22 @@ async fn find_reviewer_from_names(
...
-            initial: names.to_vec(),
-            filtered: names.to_vec(),
-        });
+    // If we are trying to assign to the ghost GitHub user, bypass every check
+    let mut filtered_candidates: HashSet<String> = HashSet::new();
+    if candidates.contains("ghost") {
+        filtered_candidates.insert("ghost".to_string());
+    } else {
+        // filter out team members without capacity
+        filtered_candidates = filter_by_capacity(db, &candidates)
```
jackh726, line 775: Nit for a followup PR: creation of the HashSet can be moved into the if.

jackh726 (approved):

PR #1796, opened 2024-04-16
Revert last two prs

apiraino:
This should revert #1793 and #1794 and return to a stable situation

r? @jackh726


jackh726 (approved):
Let's take this in parts next time - would be good to find out the best way to test too.

PR #1797, opened 2024-04-19
Fix DB schema migrations

apiraino:
- Fix a wrong migration introduced in #1781
- Clarify how DB migrations should be added to the triagebot

r? @jackh726

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -272,6 +272,9 @@ async fn handle_job(
     Ok(())
 }
 
+// Important notes when adding migrations:
+// - Each DB change is an element in this array and must be a single SQL instruction
+// - The total # of items in this array must be equal to the value of `database_versions.migration_counter`
 static MIGRATIONS: &[&str] = &[
     "
 CREATE TABLE notifications (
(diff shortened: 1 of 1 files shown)


src/db.rs
```diff
@@ -332,7 +335,8 @@ CREATE table review_prefs (
     assigned_prs INT[] NOT NULL DEFAULT array[]::INT[]
 );",
     "
-CREATE EXTENSION intarray;
-CREATE UNIQUE INDEX review_prefs_user_id ON review_prefs(user_id);
+CREATE EXTENSION IF NOT EXISTS intarray;",
+    "
+CREATE UNIQUE INDEX IF NOT EXISTS review_prefs_user_id ON review_prefs(user_id);
```
apiraino, line 340: I can't check the schema of the triagebot in prod but the clauses `IF NOT EXISTS` should ensure nothing bad happens when applied.

jackh726:
I think we should get eyes on this from @Mark-Simulacrum as I'm not very familiar with how migrations work and what the current state of the db is.

ehuss:
Will this fix the errors about `function sort(integer[]) does not exist`?

apiraino:
> Will this fix the errors about `function sort(integer[]) does not exist`?

@ehuss Yes and now I know why. Without this patch the migration would fail:
```
$ cargo run --bin triagebot
...
Failed to run server: database migrations

Caused by:
    0: executing 12th migration
    1: db error: ERROR: cannot insert multiple commands into a prepared statement
    2: ERROR: cannot insert multiple commands into a prepared statement
```

unsure why it didn't happen (or wasn't catched) during the triagebot deployment at the time :man_shrugging:

jackh726 (approved):
After talking with @Kobzol seems like this is fine

PR #1798, opened 2024-04-22
Rename GitHub token env var

### LINKED ISSUE #1789: Rename github token env var
For better consistency with [the Github docs](https://docs.github.com/en/actions/security-guides/automatic-token-authentication), I'd like the triagebot to also use GITHUB_TOKEN as env var to pick up the github API token.

Renaming all occurrences of `GITHUB_API_TOKEN` into `GITHUB_TOKEN` should do the job.

Even if the triagebot is mostly used internally, this change _will_ likely break someone workflow :) so let's ensure to make it prominent ... somewhere (in `#triagebot` on Zulip?).

@rustbot claim

apiraino:
closes #1789 

r? @jackh726

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,5 +1,5 @@
-# if `GITHUB_API_TOKEN` is not set here, the token can also be stored in `~/.gitconfig`
-GITHUB_API_TOKEN=MUST_BE_CONFIGURED
+# if `GITHUB_TOKEN` is not set here, the token can also be stored in `~/.gitconfig`
+GITHUB_TOKEN=MUST_BE_CONFIGURED
 DATABASE_URL=MUST_BE_CONFIGURED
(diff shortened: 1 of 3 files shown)


ehuss:
This will likely require updating simpleinfra to handle the secret there. Would it be possible to read both env vars to make the transition easier?

apiraino:
ok @ehuss I've added again the old env var when looking for a GH token but I'd keep the rest as it is to discourage using the old one :)

should be ready for reaview again, I think.

ehuss (approved):
Looks good to me. I'll leave this to @Mark-Simulacrum to actually click the merge button in case there are problems or it needs to be reverted, but I don't expect there to be any issues.

Mark-Simulacrum:
Feel free to merge especially now that you have logs :)

If those aren't enough let me know what else would help.

Otherwise I'll try to get to this but it'll probably be next weekend.

PR #1799, opened 2024-04-22
Github user id is not optional

### LINKED ISSUE #1775: Github userId should not be an `Option<T>`
Don't know exactly why our https://github.com/rust-lang/triagebot/blob/c52016ab56eddb5ce2ac51a7bde6a47b8653719a/src/github.rs#L19-L23 has the Id optional :shrug: 

_Originally posted by @apiraino in https://github.com/rust-lang/triagebot/pull/1773#discussion_r1496324399_
            

@rustbot claim

apiraino:
This patch removes the Optional id in the GitHub User struct.

Unsure why it was this way, but I think it is reasonable to expect that any GitHub User has an id.

Closes #1775 

r? @jackh726

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -18,7 +18,7 @@ use tracing as log;
 #[derive(Debug, PartialEq, Eq, serde::Deserialize)]
 pub struct User {
     pub login: String,
-    pub id: Option<u64>,
+    pub id: u64,
 }
 
 impl GithubClient {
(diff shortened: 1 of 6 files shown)


src/github.rs
```diff
@@ -200,17 +200,20 @@ impl User {
...
-        let permission = crate::team_data::teams(client).await?;
-        let map = permission.teams;
-        Ok(map["all"]
-            .members
-            .iter()
-            .find(|g| g.github == self.login)
-            .map(|u| u.github_id))
-    }
+// Returns the ID of the given user, if the user is in the `all` team.
+pub async fn get_id_for_username<'a>(
```
apiraino, line 206: I have replaced the `User.get_id()` with one that just takes a `&str`. I think after this patch the previous version wasn't used anywhere else.

jackh726 (approved):

PR #1800, opened 2024-05-01
Fix some webhook deserialization errors.

ehuss:
This fixes a couple of deserialization errors I was seeing in the logs:
* PR enqueued/dequeued events
* Review requested of a team sets requested_reviewer to None
* PR review comments can sometimes have `submitted_at` as null

These events aren't actually used by triagebot, but I figured it would be good to help clean up the logs.

This also adds a log message to make it easier to see what is wrong in the future.


ehuss:
Oops, switching to draft, there are some things I missed.

ehuss:
OK, pushed another change to fix an error deserializing pr review comments with a null `submitted_at`.

jackh726 (approved):

Issue #1802, opened 2024-05-02
validate_config should not run on closed PRs


ehuss:
The `validate_config` handler seems to trigger on events hitting old, closed PRs. It loads the `triagebot.toml` from the PR itself, which might be very old, and fail. https://github.com/rust-lang/rust/pull/73206#issuecomment-2089379975 is an example.

I think at a minimum it probably should not run on closed PRs. I'm not sure if it should ignore other events, but that seems like a start.

PR #1804, opened 2024-05-02
Limit validate_config actions.

ehuss:
This limits which actions the validate_config handler will handle. There was a problem where it was validating the config on every action. This means that if there is an invalid config, and *any* action was performed on the PR (like adding a label, or assigning someone), it would post a comment about the invalid config.

This is particularly a problem with old closed PRs that modified `triagebot.toml`. Minor actions like adding milestones or labels shouldn't re-trigger validation.

Fixes https://github.com/rust-lang/triagebot/issues/1802

### DIFF
--- a/src/handlers/validate_config.rs
+++ b/src/handlers/validate_config.rs
@@ -4,6 +4,7 @@
 
 use crate::{
     config::{ValidateConfig, CONFIG_FILE_NAME},
+    github::IssuesAction,
     handlers::{Context, IssuesEvent},
 };
 use tracing as log;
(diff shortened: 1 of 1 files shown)


src/handlers/validate_config.rs
```diff
@@ -13,6 +14,12 @@ pub(super) async fn parse_input(
     event: &IssuesEvent,
     _config: Option<&ValidateConfig>,
 ) -> Result<Option<()>, String> {
+    if !matches!(
+        event.action,
+        IssuesAction::Opened | IssuesAction::Reopened | IssuesAction::Synchronize
```
jackh726, line 19: Huh, what is `Synchronize`?

jackh726 (approved):

src/handlers/validate_config.rs
ehuss: Whenever commits are pushed.

PR #1808, opened 2024-05-06
Use a safer command for rebase

Kobzol:
Related PR: https://github.com/rust-lang/rust-forge/pull/742

### DIFF
--- a/src/handlers/no_merges.rs
+++ b/src/handlers/no_merges.rs
@@ -88,7 +88,7 @@ so these commits will need to be removed for this pull request to be merged.
 You can start a rebase with the following commands:
 ```shell-session
 $ # rebase
-$ git rebase -i master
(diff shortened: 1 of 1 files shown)


jackh726 (approved):

src/handlers/no_merges.rs
```diff
@@ -88,7 +88,7 @@ so these commits will need to be removed for this pull request to be merged.
 You can start a rebase with the following commands:
 ```shell-session
 $ # rebase
-$ git rebase -i master
+$ git pull --rebase https://github.com/rust-lang/rust.git master
```
RalfJung, line 91: This will be wrong (and dangerously so) for other repositories, won't it?

Kobzol: Damn, I completely forgot that we also use triagebot for other repos. Let me revert this.

PR #1813, opened 2024-05-16
Rework backport template compiler triage meeting

apiraino:
This remove a bit of manual work when rendering the weekly compiler triage meeting agenda:
- Acknowledges backports PR author (:heart:)
- Automate with the Tera templating engine a little bit of manual chore

Visually, nothing changes but according to [xkcd](xkcd.com/1205) I will save about 4 hrs work across 5 years :-)


templates/_issue.tt
```diff
@@ -1,4 +1,18 @@
-{% macro render(issue, with_age="") %}"{{issue.title}}" [{{issue.repo_name}}#{{issue.number}}]({{issue.html_url}}){% if issue.mcp_details.zulip_link %} ([Zulip]({{issue.mcp_details.zulip_link}})){% endif %}{% if with_age %} (last review activity: {{issue.updated_at_hts}}){%- endif -%}
+{% macro render(issue, with_age="", branch="") %}"{{issue.title}}" [{{issue.repo_name}}#{{issue.number}}]({{issue.html_url}}){% if issue.mcp_details.zulip_link %} ([Zulip]({{issue.mcp_details.zulip_link}})){% endif %}{% if with_age %} (last review activity: {{issue.updated_at_hts}}){%- endif -%}
+{%- if branch != "" %}
+    - Authored by {{ issue.author }}
+{%- endif -%}
 {% if issue.mcp_details.concerns %}{%- for concern in issue.mcp_details.concerns %}
     - concern: [{{- concern.0 -}}]({{- concern.1 -}})
-{%- endfor -%}{% endif -%}{% endmacro %}
+{%- endfor -%}{%- endif -%}
+{%- if branch  %}
```
jackh726, line 8: This seems like not the right place for this?

jackh726: Matching on branch seems less than ideal overall

apiraino: hmm ... what I am trying to do is twofold. Write a line that says "Authored by xxx" and append the markup code that Zulip uses to open a poll, after every backport nomination. The final result I'd like to see for every backport item is:
```
- :beta: "Title" [rust#123456](link)
  - Authored by xxx
<!--
/poll Approve beta backport of #123456?
approve
decline
don't know
-->
```

Are you suggesting that the _if branch_ conditional should be moved in the parent `_issues.tt` template? Or changing the conditional itself (ex. rename the `branch` variable into something like `backport`)?

jackh726: Renaming the variable to backport is fine, since apparently we inline things here for things like MCPs - this seems not great but not worth changing here.

apiraino: Renamed the variable to `backport_branch` (sounds even better)

PR #1815, opened 2024-05-21
small: Add documentation link on unknown commands

blyxyas:
It was hard to find the Rust forge page dedicated to documenting triagebot's Zulip commands, and there was no `help` command.

With this PR, triagebot links to [this page](https://forge.rust-lang.org/triagebot/notifications.html) on unknown commands.

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -189,13 +189,19 @@ fn handle_command<'a>(
                                 .map_err(|e| format_err!("Failed to await at this time: {e:?}"))
                             }
                             Some("docs-update") => return trigger_docs_update(message_data),
-                            _ => {}
+                            _ => {
+                                return Ok(Some(
+                                    String::from(
+                                        "Unknown command, check triagebot's documentation at https://forge.rust-lang.org/triagebot/index.html#common-command-summary"
+                                    )
+                                ))
+                            }
                         }
                     }
                     next = words.next();
                 }
 
-                Ok(Some(String::from("Unknown command")))
+                Ok(Some(String::from("Unknown command, check triagebot's documentation at https://forge.rust-lang.org/triagebot/index.html")))
             }
         }
     })


src/zulip.rs
```diff
@@ -195,7 +195,7 @@ fn handle_command<'a>(
                     next = words.next();
                 }
 
-                Ok(Some(String::from("Unknown command")))
+                Ok(Some(String::from("Unknown command, check triagebot's documentation at <https://forge.rust-lang.org/triagebot/notifications.html>")))
```
apiraino, line 198: Are the brackets around the link needed?

blyxyas: It's kinda the default formatting for links over the web, it's just a way to signify it. It isn't necessary for Zulip to parse it as a clickable link.

If it's too ugly I'll just modify it :)

apiraino (approved):
apart from a small cosmetic nit, I think it 's reasonable as a stopgap solution. And, yes, better would be a `help` command (but that's a bit more work)

src/zulip.rs
ehuss: Why does this link to the notifications page? There are other features that accept Zulip commands, and I'm concerned it could be confusing for the user since it could imply that notifications are the only commands it accepts.

blyxyas: I was having problems finding the notifications page, so I didn't notice that that implementation is the one that also handles meetings. Should be fixed now (I added another message for manual triggers @ing triagebot)

src/zulip.rs
```diff
@@ -189,13 +189,19 @@ fn handle_command<'a>(
...
                             }
                             Some("docs-update") => return trigger_docs_update(message_data),
-                            _ => {}
+                            _ => {
+                                return Ok(Some(
+                                    String::from(
+                                        "Unknown command, check triagebot's documentation at https://forge.rust-lang.org/triagebot/index.html#common-command-summary"
+                                    )
+                                ))
+                            }
```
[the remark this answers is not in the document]
Mark-Simulacrum: That sounds like a plausible theory; it also seems like this is just duplicate effort - we'd presumably fall through regardless to the trailing error if there wasn't another command found in the message.

PR #1817, opened 2024-05-29
Add more logging for unexpected status codes.

ehuss:
This adds some more logging in the saga to determine why diffing is not working as expected. For the cargo milestoning, all I see in the logs is:

```
Retrying after 0 seconds, remaining attepts 2
Retrying after 0 seconds, remaining attepts 1
Retrying after 0 seconds, remaining attepts 0
```

and then nothing. My theory is that it is returning some unexpected status code. The retry logic checks for `!is_success()` which only checks for codes 2xx, and then error reporting code only checks for `error_for_status_ref`, which is only 4xx and 5xx. My guess is that some other status code is slipping through? I'm really shooting in the dark since I can't reproduce the issue locally.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -35,13 +35,19 @@ impl GithubClient {
             resp = self.retry(req, sleep, MAX_ATTEMPTS).await?;
         }
         let maybe_err = resp.error_for_status_ref().err();
+        let status = resp.status();
         let body = resp
             .bytes()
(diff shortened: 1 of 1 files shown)


ehuss:
I'm going to close since I'm pretty confident that https://github.com/rust-lang/triagebot/pull/1819 will resolve this.

PR #1819, opened 2024-06-02
Fix cargo milestoning timeout.

ehuss:
This fixes an issue where not all cargo PRs were getting milestoned.

The process of milestoning all the cargo PRs can take more than 10 seconds. This was causing the thread to get cancelled due to the webhook timeout.

This fixes the problem by spawning a background task to handle the milestoning process.


ehuss:
It needs to hit the `set_milestone` endpoint possibly a few dozen times. That endpoint seems fairly slow.

Mark-Simulacrum (approved):

PR #1821, opened 2024-06-12
Guard against overflowing GitHub ID comments

apiraino:
Recently we overflowed the ID counter for comments on GitHub.

This causes some comment IDs from this URL https://rfcbot.rs/api/all to be negative numbers.

This patch is a follow-up to this commit https://github.com/rust-lang/rfcbot-rs/commit/f9d84fa85b54296896bcbb2cf04049f783912ea6 to avoid some tooling (notably meetings agenda generators) crashing when retrieving comments with invalid IDs. It will not retrieve a comment that has an invalid ID (e.g. `-2132153390`) 

More info about this incident at: https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/rfcbot.20asleep/near/442898636

cc: @traviscross (we discussed this)

r? @jackh726 (thanks!)

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -510,7 +510,7 @@ impl Issue {
         self.state == IssueState::Open
     }
 
-    pub async fn get_comment(&self, client: &GithubClient, id: u64) -> anyhow::Result<Comment> {
+    pub async fn get_comment(&self, client: &GithubClient, id: i32) -> anyhow::Result<Comment> {
         let comment_url = format!("{}/issues/comments/{}", self.repository().url(client), id);
         let comment = client.json(client.get(&comment_url)).await?;
         Ok(comment)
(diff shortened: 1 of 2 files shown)


apiraino:
review plssss :bowing_man: 

who can have a loook at this?

src/github.rs
```diff
@@ -1825,20 +1825,34 @@ impl<'q> IssuesQuery for Query<'q> {
...
                     let fk_initiating_comment = fcp.fcp.fk_initiating_comment;
-                    let init_comment = issue
-                        .get_comment(&client, fk_initiating_comment.try_into()?)
-                        .await?;
+                    let (initiating_comment_html_url, initiating_comment_content) =
+                        if u32::try_from(fk_initiating_comment).is_err() {
+                            // We blew out the GH comment incremental counter (a i32 on their end)
+                            // See: https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/rfcbot.20asleep
+                            log::debug!("Ignoring overflowed comment id from GitHub");
+                            ("".to_string(), "".to_string())
```
ehuss, line 1834: I don't fully understand why this is using empty strings. But I'll trust since I assume you do most of the agenda stuff that it works for you.

ehuss (approved):
Thanks!

src/github.rs
apiraino: yes it's a bit of a hack and I am not either very fond of this. `initiating_comment_html_url` and `initiating_comment_content` are mandatory fields in `FCPDetails`, so the minimal change I could think of was to set them to an empty string rather than making them `Option<String>` and deal with a widerspread change in the codebase.

PR #1828, opened 2024-07-25
ping goal owners

nikomatsakis:
Add a zulip command to ping goal owners. I intend this to run once a week eventually but right now it's still executed manually.

### DIFF
--- a/src/handlers/project_goals.rs
+++ b/src/handlers/project_goals.rs
@@ -1,0 +1,153 @@
+use crate::github::Issue;
+use crate::github::{self, GithubClient, User};
+use crate::jobs::Job;
+use crate::zulip::to_zulip_id;
+use anyhow::Context as _;
+use async_trait::async_trait;
+use chrono::Utc;
+use tracing::{self as log};
+
+const MAX_ZULIP_TOPIC: usize = 60;
+const RUST_PROJECT_GOALS_REPO: &'static str = "rust-lang/rust-project-goals";
+const GOALS_STREAM: u64 = 435869; // #project-goals
+
+const MESSAGE: &str = r#"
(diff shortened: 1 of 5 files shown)


src/zulip.rs
```diff
@@ -188,6 +189,12 @@ fn handle_command<'a>(
                                 .await
                                 .map_err(|e| format_err!("Failed to await at this time: {e:?}"))
                             }
+                            Some("ping-goals") => {
```
jackh726, line 192: So...I feel like it may be prudent to gate this so that only you (@nikomatsakis) can run this.

jackh726 (commented):
Mostly good, one thought.

src/zulip.rs
nikomatsakis: ...good idea, at least for now. :)

nikomatsakis:
@jackh726 added the simplest of all possible ACLs :)

src/handlers/project_goals.rs
```diff
@@ -39,6 +39,18 @@ impl Job for ProjectGoalsUpdateJob {
     }
 }
 
+/// Returns true if the user with the given github id is allowed to ping all group people
+/// and do other "project group adminstrative" tasks.
+pub async fn check_project_goal_acl(_gh: &GithubClient, gh_id: u64) -> anyhow::Result<bool> {
+    /// Github ID of the user allowed to ping all group people.
+    ///
+    /// FIXME: We should create a team for the person/people managing the goals program
+    /// and check that the zulip person is on it, but I'm too
```
jackh726, line 48: Left out a word here, but its funnier this way

jackh726 (approved):

PR #1835, opened 2024-08-24
Initial implementation of relnotes handler

Mark-Simulacrum:
This follows up on recent discussion with T-libs-api (in particular) to make it easier for T-release and teams to propose specific text for inclusion in release notes. The strategy proposed here is a dedicated issue for release note text discussion + revision. Follow-up work to this will automate pulling the relevant context into the release notes PR (likely through changes to https://github.com/rust-lang/relnotes/), but manually extracting it should already be pretty easy due to the prominent title.

There are many future improvements, e.g.:

* Automatic nomination for the right team based on initial PR labels
* Automatic creation for any FCP (this feels right to me, but it is more work than teams typically do today).
* Random assignment to (subset of?) team members rather than nomination?
* Live-ish rendering of the proposed content, perhaps by replacing the code blocks with some other delimiter.

But my hope is that we can iterate on this in practice as we gain experience.

cc @rust-lang/release 

r? @ehuss

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -166,6 +166,33 @@ impl GithubClient {
         let (body, _req_dbg) = self.send_req(req).await?;
         Ok(serde_json::from_slice(&body)?)
     }
+
+    pub(crate) async fn new_issue(
+        &self,
+        repo: &IssueRepository,
+        title: &str,
+        body: &str,
+        labels: Vec<String>,
+    ) -> anyhow::Result<NewIssueResponse> {
+        #[derive(serde::Serialize)]
+        struct NewIssue<'a> {
+            title: &'a str,
+            body: &'a str,
+            labels: Vec<String>,
+        }
+        self.json(self.client.post(repo.url(&self)).json(&NewIssue {
+            title,
+            body,
+            labels,
+        }))
+        .await
+        .context("failed to create issue")
+    }
+}
+
+#[derive(Debug, serde::Deserialize)]
+pub struct NewIssueResponse {
+    pub number: u64,
 }
 
 impl User {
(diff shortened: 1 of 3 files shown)


Mark-Simulacrum:
We discussed in the last release team meeting using a GitHub project for tracking this, but I decided to not pursue that here. The "text" input in those is very limited (basically just one line) which doesn't feel great, and also lacks a clear place to discuss the changes being proposed, which also doesn't feel right. Long-term, I think having a clearly separate place for discussing how changes should be announced to the public also feels like a good thing, since it creates a dedicated point on every team's change process rather than making that a separate activity done weeks after decisions are made (and typically by different people).

src/github.rs
```diff
@@ -166,6 +166,33 @@ impl GithubClient {
...
+        body: &str,
+        labels: Vec<String>,
+    ) -> anyhow::Result<NewIssueResponse> {
+        #[derive(serde::Serialize)]
+        struct NewIssue<'a> {
+            title: &'a str,
+            body: &'a str,
+            labels: Vec<String>,
+        }
+        self.json(self.client.post(repo.url(&self)).json(&NewIssue {
```
ehuss, line 183: The URL for this doesn't look correct to me. It needs to go to the `/issues` endpoint.

Also, the `post` method needs to be on `self` to properly configure the request.

```suggestion
        let url = format!("{}/issues", repo.url(&self));
        self.json(self.post(&url).json(&NewIssue {
```

src/handlers/relnotes.rs
```diff
@@ -0,0 +1,116 @@
...
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+struct RelnotesState {
+    relnotes_issue: Option<u64>,
+}
+
+pub async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
+    let e = if let Event::Issue(e) = event {
+        e
+    } else {
```
ehuss, line 33: Just a style nit.

```suggestion
    let Event::Issue(e) = event else {
```

src/github.rs
Mark-Simulacrum: Yeah, good catch.

ehuss (approved):
Feel free to merge when you are ready.

PR #1836, opened 2024-08-28
add more parameters to the project goals message

nikomatsakis:
Also, write the code to automatically ping on a regular cadence, but don't yet schedule it.

I'm not sure how I *would* schedule it! Probably a good idea to land this first without an automatic schedule and run it manually for a bit. But I'd like to know where in the code I have to go.

r? @jackh726

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -103,6 +103,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "atty"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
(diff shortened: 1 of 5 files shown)


apiraino:
To have an idea on where to touch to add a scheduled cronjob, you can have a look at a [recent patch](https://github.com/rust-lang/triagebot/pull/1824/commits/3c61fa6841b25497c5b625392d87ac3c0211a722) where I removed a scheduled job (and do the opposite :slightly_smiling_face: ). The interesting code is all in `./src/jobs.rs`, in your handler you have to impl `Job` and add a `name()` and a `run()`.

I don't remember offhand if I ever understood the machinery to _manually_ trigger a job. Maybe creating a small binary in `./src/bin` to run the `Job`?

jackh726 (approved):

PR #1839, opened 2024-09-06
Extend automatic rendered link to the blog.rust-lang.org repo

Urgau:
This PR extends the automatic addition of the "Rendered" link to the [blog.rust-lang.org repo](https://github.com/rust-lang/blog.rust-lang.org), as is currently done for the [rfcs repo](https://github.com/rust-lang/rfcs).

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -44,9 +44,9 @@ pub mod project_goals;
 pub mod pull_requests_assignment_update;
 mod relabel;
 mod relnotes;
+mod rendered_link;
 mod review_requested;
 mod review_submitted;
-mod rfc_helper;
 pub mod rustc_commits;
 mod shortcut;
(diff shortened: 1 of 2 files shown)


apiraino:
Maybe it's me but I don't understand what this patch is doing (I feel it's doing something useful). Can you please provide an example? :slightly_smiling_face:

apiraino:
Ah OK, so the blog posts PRs will automatically have the markdown rendered link in the opening comment instead of the user having to click around and find it themself.

Cool, thanks.

ehuss (approved):
Thanks! I'm a little on the fence with how useful this is. There are also some drawbacks (like these links will often be broken after merging). However, I don't think it should cause too much harm, and we can easily revert this if people have problems with it.

PR #1843, opened 2024-09-13
Copy over labels to the relnote tracking issue

Urgau:
This PR extends the relnotes handling to copy over *some* labels to the relnote tracking issue.

This is done to avoid manually copying them from the nominated relnote issue/PR.

r? @Mark-Simulacrum

### DIFF
--- a/src/handlers/relnotes.rs
+++ b/src/handlers/relnotes.rs
@@ -116,7 +116,18 @@ cc {} -- origin issue/PR authors and assignees for starting to draft text
                     &e.issue.repository(),
                     &title,
                     &body,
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (approved):
I guess this seems okay, but I'm not convinced this is helpful - but I don't really use these labels often so maybe useful for others. Merging.

Urgau:
I'm not sure of it's usefulness either, but I have noticed that @jieyouxu started adding them to the recently created [relnotes tracking issues](https://github.com/rust-lang/rust/issues?q=is%3Aopen+is%3Aissue+label%3Arelnotes-tracking-issue), so I thought that copying them over automatically was "better" and faster. If it's not useful we can revert it.

Issue #1803, opened 2024-05-02
triagebot rfc_helper cannot deal with the file being renamed to the RFC number


RalfJung:
The RFC process says

> Now that your RFC has an open pull request, use the issue number of the PR to update your 0000- prefix to that number.

However, rfcbot will create the "Rendered" link to point to the file "0000-*", which will then be a dead link that one has to manually update. The bot should probably update this link automatically?

ehuss:
I'm going to reopen, since I don't think #1848 quite resolves this issue since it doesn't update for pushes.

PR #1848, opened 2024-10-15
Improve rendered link handler with proper edit and sha-based link

Urgau:
This PR refactors the `rendered_link` handler to be more robust, in particular regarding closing/merging of a PR with a SHA-based url in those cases.

Fixes https://github.com/rust-lang/triagebot/issues/1803
r? @ehuss

### DIFF
--- a/src/handlers/rendered_link.rs
+++ b/src/handlers/rendered_link.rs
@@ -1,3 +1,5 @@
+use anyhow::bail;
+
 use crate::{
     github::{Event, IssuesAction, IssuesEvent},
     handlers::Context,
@@ -9,6 +11,10 @@ pub async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
     } else {
         return Ok(());
     };
+
+    if !e.issue.is_pr() {
+        return Ok(());
+    }
 
     let repo = e.issue.repository();
     let prefix = match (&*repo.organization, &*repo.repository) {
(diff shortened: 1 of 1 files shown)


src/handlers/rendered_link.rs
```diff
@@ -25,32 +35,47 @@ pub async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
 }
 
 async fn add_rendered_link(ctx: &Context, e: &IssuesEvent, prefix: &str) -> anyhow::Result<()> {
-    if e.action == IssuesAction::Opened {
+    if e.action == IssuesAction::Opened
+        || e.action == IssuesAction::Closed
+        || e.action == IssuesAction::Reopened
+    {
+        let edit = EditIssueBody::new(&e.issue, "RENDERED_LINK");
```
ehuss, line 42: I'm a little concerned about the approach of using `EditIssueBody` for a few reasons:

* AFAIK, the triagebot comments persist in the git log.
* Some users manually add their own Rendered links, which I think will conflict with this (it seems it will add multiple rendered links).
* This may conflict with all RFCs that are currently open.

Would it be possible to keep the same logic of just manually looking for the link?

src/handlers/rendered_link.rs
```diff
@@ -25,32 +35,47 @@ pub async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
...
+            current_data.rendered_link = format!(
+                "https://github.com/{}/blob/{}/{}",
+                head.repo.full_name,
+                if e.action == IssuesAction::Closed {
+                    &head.sha
+                } else {
+                    &head.git_ref
+                },
+                file.filename
+            );
```
ehuss, line 67: I don't think this is going to be the correct link for what we want. I think for now the preference is:

If it is merged, link to `https://github.com/rust-lang/rfcs/blob/master/text/{filename}`. We don't want to link to the user's repo with a sha hash, because I don't think it will persist over time if deleted (the branch or the repo). And we probably want master to render any minor fixes made afterwards.

If it is not merged, link to the last known copy as `https://github.com/rust-lang/rfcs/blob/{sha}/text/{filename}`. I'm pretty sure those links should be permanent, as GitHub keeps a ref to the PR which should keep the blobs from being gc'd.

src/handlers/rendered_link.rs
```diff
@@ -25,32 +35,47 @@ pub async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
...
+
+            edit.apply(
+                &ctx.github,
+                format!("[Rendered]({})", &current_data.rendered_link),
+                current_data,
+            )
+            .await?;
+        } else if let Some(mut current_data) = current_data {
+            // No render link to show, but one previously, so remove it
```
ehuss, line 76: I'm curious, is this addressing something that has happened in the past, or is it just being defensive?

src/handlers/rendered_link.rs
Urgau: > * AFAIK, the triagebot comments persist in the git log.

I'm sure I understand this point. Are you saying that triagebot comments are somehow included in git log?

> * Some users manually add their own Rendered links, which I think will conflict with this (it seems it will add multiple rendered links).

I don't think it would as triagebot will add the comment and surround it with hidden html marker (which is the reason I wanted to use `EditIssueBody`, since those markers simplifies many things).

https://github.com/rust-lang/triagebot/blob/f563bc9578ae26d825f0ab6b1255068afebe6a2a/src/interactions.rs#L100-L106

> * This may conflict with all RFCs that are currently open.

Good point. I can add a defensive mechanism for older RFCs.

> Would it be possible to keep the same logic of just manually looking for the link?

The issue is not really looking for the link but updating it as well, which is heavily simplified by `EditIssueBody`.

src/handlers/rendered_link.rs
Urgau: Should we worry about missing some webhook that would make the "last known copy" (sha) no longer the "last known copy"?

I ask, because right now the link we produce is immunized against that.

src/handlers/rendered_link.rs
ehuss: > Are you saying that triagebot comments are somehow included in git log?

Yea. When a PR is merged, GitHub sets the contents of the PR description body as the commit message. AFAIK it does not strip HTML comments.

> I don't think it would as triagebot will add the comment and surround it with hidden html marker (which is the reason I wanted to use `EditIssueBody`, since those markers simplifies many things).

If someone posts a PR, and they manually add a rendered link, the body will look like:

```
Hey, here's my new PR!

[Rendered](https://github.com/octocat/rfcs/blob/4917eb9a8c2a3da10a5e46b080498b33655b1054/text/0001-foo.md)
```

and then triagebot would edit it to be:

```
Hey, here's my new PR!

[Rendered](https://github.com/octocat/rfcs/blob/4917eb9a8c2a3da10a5e46b080498b33655b1054/text/0001-foo.md)

<!-- TRIAGEBOT_START -->

<!-- TRIAGEBOT_RENDERED_LINK_START -->
[Rendered](https://github.com/octocat/rfcs/blob/branch/text/0001-foo.md)

<!-- TRIAGEBOT_RENDERED_LINK_DATA_START$${"rendered_link":"https://github.com/rust-lang/rfcs/blob/branch/text/0001-foo.md"}$$TRIAGEBOT_RENDERED_LINK_DATA_END -->

<!-- TRIAGEBOT_RENDERED_LINK_END -->
<!-- TRIAGEBOT_END -->
```

This will render in Markdown as two rendered links.

> The issue is not really looking for the link but updating it as well, which is heavily simplified by `EditIssueBody`.

I understand that `EditIssueBody` makes it easier to find and edit content, but it does not work well with PRs (and thus why we never use it there).

src/handlers/rendered_link.rs
Urgau: In the "If it is not merged" branch you used `{sha}` for the "last known copy", but since we only react to open, reopened, closed; using the SHA would not work as the PR is updated, we should be using `{git_ref}` or also react to push (and update the sha).

> I'm not too terribly concerned about that (and there's not much to be done about it anyways).

:+1:

src/handlers/rendered_link.rs
Urgau: Alright, changed to a more manual approach.

src/handlers/rendered_link.rs
Urgau: I've update the PR with the new logic. Might be easier to follow what I was saying.

src/handlers/rendered_link.rs
```diff
@@ -25,32 +31,69 @@ pub async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
...
+                // add rendered link to the end of the body
+                format!("{}\n\n{rendered_link}", e.issue.body)
+            } else if let Some(start_pos) = e.issue.body.find("[Rendered](") {
+                let Some(end_pos) = &e.issue.body[start_pos..].find(')') else {
+                    bail!("no `)` after `[Rendered]` found")
+                };
+
+                // replace the current rendered link with the new one
+                e.issue.body.replace(
+                    &e.issue.body[start_pos..=(start_pos + end_pos + 1)],
```
ehuss, line 87: When merging, this is causing a panic: "byte index 91 is out of bounds of"

I'm not sure what the `=` is for, but it doesn't seem correct to me.

```suggestion
                    &e.issue.body[start_pos..(start_pos + end_pos + 1)],
```

src/handlers/rendered_link.rs
```diff
@@ -25,32 +31,69 @@ pub async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
...
+                    &rendered_link,
+                )
+            } else {
+                bail!(
+                    "found `[Rendered]` but no it's associated link, can't replace it, bailing out"
```
ehuss, line 92: Something seems off here. Is this maybe missing a letter?

```suggestion
                    "found `[Rendered]` but not it's associated link, can't replace it, bailing out"
```

src/handlers/rendered_link.rs
Urgau: Indeed, not sure what happened, when I tested it locally it was fine, anyway fixed.

ehuss (approved):
Thanks! Let's give it a shot and see how it goes.

Issue #1833, opened 2024-08-19
Outdated Merge commit warning


ZuseZ4:
After seing a github notification about merge conflicts with my PR, I used githubs GUI to resolve them, which I guess I shouldn't do.
It introduced a merge, which promptly caused this bot message, telling me how to get rid of them:
https://github.com/rust-lang/rust/pull/129176#issuecomment-2295171771

I thus fixed it locally and force-pushed, but unlike previous outdated messages which automatically got minimized, this stayed.
I guess that's not intended and for this message there is just no trigger to minimize it yet?

PR #1849, opened 2024-10-15
Hide no-merge comments once merge commits removed

Urgau:
This PR implements the a GraphQL call to minimize a comment, modifies the return of `post_comment` to return the node-id (inside the more general `Comment`).

And then use those pre-requires to hide all the no-merge comments once the merge commits are removed.

Fixes https://github.com/rust-lang/triagebot/issues/1833
r? @ehuss

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -405,6 +405,17 @@ pub struct Comment {
     pub pr_review_state: Option<PullRequestReviewState>,
 }
 
+#[derive(Debug, serde::Deserialize, serde::Serialize, Eq, PartialEq)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum ReportedContentClassifiers {
+    Abuse,
+    Duplicate,
+    OffTopic,
+    Outdated,
+    Resolved,
+    Spam,
+}
+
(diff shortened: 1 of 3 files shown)


src/github.rs
```diff
@@ -608,10 +610,32 @@ impl Issue {
...
+        node_id: &str,
+        reason: &str,
+    ) -> anyhow::Result<()> {
+        client
+            .graphql_query(
+                "mutation($node_id: ID!, $reason: ReportedContentClassifiers!) {
+                    minimizeComment(input: {subjectId: $node_id, classifier: $reason}) {
+                        __typename
+                    }
+                }",
```
ehuss, line 632: This doesn't seem to work. It generates an error:

```
failed to hide previous merge commit comment

Caused by:
    error: Variable $reason of type ReportedContentClassifiers! was provided invalid value
```

The classifier is an [enum](https://docs.github.com/en/graphql/reference/enums#reportedcontentclassifiers) with specific values, but it looks like this is passing in an arbitrary string. I would suggest adding an enum for the classifiers.

Urgau: Good catch. I was wondering what values where being accepted, I guess I didn't look enough to see the enum.

Added the enum with only the permitted values.

PR #1850, opened 2024-10-30
Handle pushes and rename for rendered links

### LINKED ISSUE #1803: triagebot rfc_helper cannot deal with the file being renamed to the RFC number

Urgau:
Follow up to #1848 to handle new commits/pushes on PR with rendered link.

I also added the logic to remove a rendered link if we can't point to it anymore, as well as a small optimize to avoid unnecessarily updating the body if not necessary.

Fixes #1803 *(for real this time)*
r? @ehuss

### DIFF
--- a/src/handlers/rendered_link.rs
+++ b/src/handlers/rendered_link.rs
@@ -34,6 +34,7 @@ async fn add_rendered_link(ctx: &Context, e: &IssuesEvent, prefix: &str) -> anyh
     if e.action == IssuesAction::Opened
         || e.action == IssuesAction::Closed
         || e.action == IssuesAction::Reopened
+        || e.action == IssuesAction::Synchronize
     {
         let files = e.issue.files(&ctx.github).await?;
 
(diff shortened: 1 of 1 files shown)


src/handlers/rendered_link.rs
```diff
@@ -34,65 +36,83 @@ async fn add_rendered_link(ctx: &Context, e: &IssuesEvent, prefix: &str) -> anyh
     if e.action == IssuesAction::Opened
         || e.action == IssuesAction::Closed
         || e.action == IssuesAction::Reopened
+        || e.action == IssuesAction::Synchronize
```
Urgau, line 39: I originally though I would have to go through the `PushEvent` instead of the `Issue/PushRequestEvent`, but it seems to be as simple as checking the `synchronize` action, yay for simplicity.

ehuss (approved):
Thanks!

PR #1853, opened 2024-11-05
feat(notify-zulip): support `notify-zulip.<label>` getting mapped to multiple actions

ismailarilik:
This is a requested feature in the Rust project: https://github.com/rust-lang/rust/blob/master/triagebot.toml#L492 (Notice FIXME comments there)

With this PR, multiple `notify-zulip` tables are possible. For example:

```toml
[notify-zulip."beta-nominated"]
required_labels = ["T-rustdoc"]
zulip_stream = 266220
topic = "beta-nominated: #{number}"
message_on_add = "PR #{number}'s beta-nomination has been added."
message_on_remove = "PR #{number}'s beta-nomination has been removed."
message_on_close = "PR #{number} has been closed. Thanks for participating!"
message_on_reopen = "PR #{number} has been reopened. Pinging @*T-rustdoc*."

[notify-zulip."beta-nominated".compiler]
required_labels = ["T-compiler"]
zulip_stream = 266221
topic = "beta-nominated: #{number}"
message_on_add = "PR #{number}'s beta-nomination has been added."
message_on_remove = "PR #{number}'s beta-nomination has been removed."
message_on_close = "PR #{number} has been closed. Thanks for participating!"
message_on_reopen = "PR #{number} has been reopened. Pinging @*T-compiler*."

[notify-zulip."beta-nominated".miri]
required_labels = ["T-miri"]
zulip_stream = 266222
topic = "beta-nominated: #{number}"
message_on_add = "PR #{number}'s beta-nomination has been added."
message_on_remove = "PR #{number}'s beta-nomination has been removed."
message_on_close = "PR #{number} has been closed. Thanks for participating!"
message_on_reopen = "PR #{number} has been reopened. Pinging @*T-miri*."
```

According to the config above, for example, if an issue with the labels `beta-nominated`, `T-rustdoc` and `T-miri` opened, the streams `266220` and `266222` would be notified simultaneously.

It is backward-compatible, there is no need to change any current configuration. New `notify-zulip` items should have names like above.

**Note:** This is just a draft PR. I believe my approach is right but I am not confident on the implementation so I need your feedback. Another thing is that I forgot to update the docs. =)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -265,7 +265,51 @@ pub(crate) struct AutolabelLabelConfig {
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
 pub(crate) struct NotifyZulipConfig {
     #[serde(flatten)]
-    pub(crate) labels: HashMap<String, NotifyZulipLabelConfig>,
+    pub(crate) labels: HashMap<String, NotifyZulipTablesConfig>,
+}
+
+#[derive(PartialEq, Eq, Debug)]
+pub(crate) struct NotifyZulipTablesConfig {
+    pub(crate) subtables: HashMap<String, NotifyZulipLabelConfig>,
+}
+
+impl<'de> serde::Deserialize<'de> for NotifyZulipTablesConfig {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        use serde::de::Error;
+        use toml::Value;
+
+        // Deserialize into a toml::value::Table for dynamic inspection
+        let table = toml::Value::deserialize(deserializer)?
+            .as_table()
+            .cloned()
+            .ok_or_else(|| Error::custom("expected a TOML table"))?;
+
+        let mut subtables = HashMap::new();
+        let mut direct_fields = toml::value::Table::new();
+
+        for (k, v) in table {
+            if let Some(subtable) = v.as_table() {
+                // This is a subtable; deserialize as NotifyZulipLabelConfig
+                let sub = NotifyZulipLabelConfig::deserialize(Value::Table(subtable.clone()))
+                    .map_err(Error::custom)?;
+                subtables.insert(k, sub);
+            } else {
+                // This is a direct field; collect for the "" entry
+                direct_fields.insert(k, v);
+            }
+        }
+
+        if !direct_fields.is_empty() {
+            let direct = NotifyZulipLabelConfig::deserialize(Value::Table(direct_fields))
+                .map_err(Error::custom)?;
+            subtables.insert("".to_string(), direct);
+        }
+
+        Ok(NotifyZulipTablesConfig { subtables })
+    }
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
--- a/src/handlers/notify_zulip.rs
+++ b/src/handlers/notify_zulip.rs
@@ -1,5 +1,5 @@
 use crate::{
-    config::{NotifyZulipConfig, NotifyZulipLabelConfig},
+    config::{NotifyZulipConfig, NotifyZulipLabelConfig, NotifyZulipTablesConfig},
     github::{Issue, IssuesAction, IssuesEvent, Label},
     handlers::Context,
 };
@@ -12,6 +12,9 @@ pub(super) struct NotifyZulipInput {
     /// For example, if an `I-prioritize` issue is closed,
     /// this field will be `I-prioritize`.
     label: Label,
+    /// List of strings for tables such as [notify-zulip."beta-nominated"]
+    /// and/or [notify-zulip."beta-nominated".compiler]
+    include_config_names: Vec<String>,
 }
 
 pub(super) enum NotificationType {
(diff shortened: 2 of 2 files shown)


apiraino:
hey @ismailarilik thank you for this contribution. This was actually something we in T-compiler wanted (was on my TODO list...), see [Zulip discussion](https://rust-lang.zulipchat.com/#narrow/channel/238009-t-compiler.2Fmeetings/topic/.E2.9C.94.20Poll.3A.20suggested.20changes.20to.20the.20triage.20agenda/near/432715547).

One thing I can spot at a first look is that our discussion for more favorable to publish these notification in a substream (e.g. `T-compiler/backports` instead of `T-compiler`). Let's remember about this detail when this change is merged and deployed, a specific stream needs to be created (maybe I can even do that, need to check my permissions) :slightly_smiling_face: 

cc @rust-lang/rustdoc so they can express their preference either

aDotInTheVoid:
From a rustdoc side, my personal preference would be to keep those notifications in the main stream. It's low trafic enough I don't think it's worth (us) making a new stream for it, although I can imagine this being different for a larger/higher traffic team like compiler.

ismailarilik:
I was thinking that the docs were in that repo but it seems that they are in the Forge. I will update it after this PR is merged. BTW, here is the related page in Forge: https://forge.rust-lang.org/triagebot/zulip-notifications.html

ismailarilik:
How can I request a review? =)
I couldn't see a `CONTRIBUTING.md` so couldn't figure this out. Maybe I should just wait someone to pick this up and review. Sorry if that is the case.

ismailarilik:
Sorry for the noise but do you know how I can test this appropriately? I setup triagebot in my local by following README and tried this implementation; it seems that it is working. But of course, I didn't try to send any Zulip notification to a team. Even if I tried this, I couldn't totally ensure that it would work on production.

apiraino:
@ismailarilik I'll have a look at your patch later. Maintainers of this repo are listed [here](https://github.com/rust-lang/team/blob/e172b40586144aad58ceea7033cf3aacea82e3a1/teams/triagebot.toml#L6-L10) (not an obvious place), you can ping any of them. Please allow some time for a review.

> how I can test this appropriately

To test the Zulip part one needs a Zulip instance. The quickest way is probably to register a [new free instance](https://zulip.com/new/) for testing purposes.

r? @ehuss

src/config.rs
```diff
@@ -221,7 +221,15 @@ pub(crate) struct AutolabelLabelConfig {
...
 pub(crate) struct NotifyZulipConfig {
     #[serde(flatten)]
-    pub(crate) labels: HashMap<String, NotifyZulipLabelConfig>,
+    pub(crate) labels: HashMap<String, NotifyZulipNameConfig>,
+}
+
+#[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+pub(crate) struct NotifyZulipNameConfig {
+    #[serde(flatten)]
+    pub(crate) default: Option<NotifyZulipLabelConfig>,
```
apiraino, line 230: @ismailarilik can you put a comment explaining what this field (is this a label?) is and why is it needed?

apiraino:
I gave a first look and this looks good to me :)

src/config.rs
ismailarilik: Done.

GuillaumeGomez:
> From a rustdoc side, my personal preference would be to keep those notifications in the main stream. It's low trafic enough I don't think it's worth (us) making a new stream for it, although I can imagine this being different for a larger/higher traffic team like compiler.
> 
> Also relevant context: [rust-lang/rust#116957](https://github.com/rust-lang/rust/pull/116957)

Another rustdoc point of view: I agree. :)

apiraino:
@ismailarilik I am working on making it possible for the triagebot using another Zulip instance for local testing. See my work in progress [patch](https://github.com/apiraino/triagebot/commit/bcf7836607a3be4d0dc194d3661fa155d9c6990a), feel free to test it, if you like (I didn't yet).

apiraino:
I've tested this on a test Zulip instance (by manually applying [this patch](https://github.com/rust-lang/triagebot/pull/1872)) and it seems to work.

src/handlers/notify_zulip.rs
```diff
@@ -12,6 +12,8 @@ pub(super) struct NotifyZulipInput {
     /// For example, if an `I-prioritize` issue is closed,
     /// this field will be `I-prioritize`.
     label: Label,
+    is_default_valid: bool,
+    names: Vec<String>,
```
Urgau, line 16: It's not clear at first glance what both of them do, a comment for each of them would be appreciated.

src/handlers/notify_zulip.rs
```diff
@@ -12,6 +12,8 @@ pub(super) struct NotifyZulipInput {
     /// For example, if an `I-prioritize` issue is closed,
     /// this field will be `I-prioritize`.
     label: Label,
+    is_default_valid: bool,
```
Urgau, line 15:
```suggestion
    include_default: bool,
```

src/handlers/notify_zulip.rs
```diff
@@ -52,26 +54,71 @@ pub(super) async fn parse_input(
...
+                        is_default_valid = true;
+                    }
+                    IssuesAction::Unlabeled { .. }
+                        if !label_config.messages_on_remove.is_empty() =>
+                    {
+                        is_default_valid = true;
+                    }
+                    _ => (),
+                }
+            }
```
Urgau, line 76: Can you de-duplicate this code with the one below?

src/handlers/notify_zulip.rs
```diff
@@ -92,24 +139,69 @@ fn parse_close_reopen_input(
...
+                                is_default_valid = true;
+                            }
+                            IssuesAction::Reopened
+                                if !label_config.messages_on_reopen.is_empty() =>
+                            {
+                                is_default_valid = true;
+                            }
+                            _ => (),
+                        }
+                    }
```
Urgau, line 159: As well as de-duplicating this one with the one below.

src/handlers/notify_zulip.rs
Urgau, line 16: maybe `include_config_names`?
```suggestion
    include_config_names: Vec<String>,
```

ismailarilik: Done.

src/handlers/notify_zulip.rs
ismailarilik: Done.

src/handlers/notify_zulip.rs
ismailarilik: Done.

src/handlers/notify_zulip.rs
ismailarilik: I thought it before but there are too many differences in these codes so I think de-duplication doesn't worth it.

Urgau: Taking a step back, most of the duplication is due to "default" and named case being two separate cases the deserialization, maybe we should re-conciliate them at parsing time, as to avoid complicated logic down the line.

What I'm thinking is to pretend that when the user wrote `[notify-zulip."beta-nominated"]` they actually wrote `[notify-zulip."beta-nominated".""]` (notice the empty string).

I experimented with the deserializing of that with serde and a custom Deserializer at https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=aed3e7f8424ca4ea56da6216d06d3e8a.

I'm curious what you think of it and if it helps with the de-duplication in the code.

ismailarilik: Done.

Urgau:
Thanks for the continued effort.

I intend to merge it tomorrow (or Saturday).

cc @Kobzol (in case you want to take a look)

src/handlers/notify_zulip.rs
```diff
@@ -52,26 +55,40 @@ pub(super) async fn parse_input(
...
     label: Label,
-    config: &NotifyZulipLabelConfig,
+    config: &NotifyZulipTablesConfig,
 ) -> Option<NotifyZulipInput> {
-    if !has_all_required_labels(&event.issue, config) {
-        // Issue misses a required label, ignore this event
+    let mut include_config_names: Vec<String> = vec![];
+
+    for (name, label_config) in &config.subtables {
+        if has_all_required_labels(&event.issue, &label_config) {
```
Kobzol, line 63: Hmm, I wonder how this works (even though this is pre-existing behavior). When the required label was removed (`IssuesAction::Unlabeled`), then `event.issue` will not contain the required label anymore, so `has_all_required_labels` doesn't trigger, right? So how does the code then send the `message_on_remove` message? :thinking:

ismailarilik: As I understood, `IssuesAction::Labeled` and `IssuesAction::Unlabeled` is only for the title of the table, such as `beta-nomination`. So when we delete a required label, such as, `T-rustdoc`, no message would be sent. It makes sense to me because it is a *required* label so without it, no message should be sent.

Kobzol: Oh, I see, thanks for clarifying! A bit confusing that there are two sets of labels involved, but I guess that's just how it works.

PR #1854, opened 2024-11-06
Allow setting review assignment limit

apiraino:
Part of #1753 

With this patch we allow team members to set their own review capacity. 

This will work by sending a message to the triagebot bot on Zulip. Usage:
- `work set #`: will set the number of max PRs assigned to me

This setting will only work for a subset of Github users that volunteered (see [here](https://github.com/rust-lang/triagebot/pull/1790/files#diff-3e34a2cbc9338bb237b61758076b2df481617021529ca7e0a044abd61de431cdR244-R253)).

An update the documentation on the forge will follow up.

(I cannot test it, don't have yet a testing infrastructure with also a Zulip instance)

r? @jackh726 

thank you

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,32 @@ gh webhook forward --repo=ehuss/triagebot-test --events=* \
 
 Where the value in `--secret` is the secret value you place in `GITHUB_WEBHOOK_SECRET` in the `.env` file, and `--repo` is the repo you want to test against.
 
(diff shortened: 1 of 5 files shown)


Kobzol:
This was implemented in https://github.com/rust-lang/triagebot/pull/1919, which was based on this PR. Thank you!

PR #1856, opened 2024-11-08
Add merge-conflict notifications

ehuss:
This adds a new handler which will post comments on PRs when there is a merge conflict. I have appreciated this feature in homu, and I think it would be helpful for repositories that are not using homu.

A high-level overview:
- There is a new `[merge-conflict]` conflg table to enable the feature. It supports labels just like homu.
- There are two separate events that it listens to:
    - A push to a repo branch will trigger a scan through all open PRs for PRs that have new conflicts.
    - A push to a PR, which either adds a conflict or resolves an existing conflict. This also includes opening/reopening PRs.
- State is stored in the database so that only one conflict comment is posted.
    - After the conflict is resolved, the comment will be hidden.
- This is complicated by the fact that the `mergeable` status of a PR is updated both lazily and asynchronously. That is, GitHub doesn't bother computing it until you ask for it, which it then returns `null`. Then, GitHub has an async background task which computes the value. There is no notification (AFAIK) when that is done, so we just poll again after a delay.

### DIFF
--- a/src/handlers/merge_conflict.rs
+++ b/src/handlers/merge_conflict.rs
@@ -11,8 +11,7 @@
 //!
 //! In general, multiple pushes happening quickly should be rare. And when it
 //! does happen, hopefully the state in the database will prevent duplicate
-//! messages. However, the database check itself is racey without a lock (see
-//! `maybe_add_comment`).
+//! messages.
 
 use crate::{
     config::MergeConflictConfig,
@@ -262,11 +261,6 @@ async fn maybe_add_comment(
     issue: &Issue,
     possibly: Option<&str>,
 ) -> anyhow::Result<()> {
-    // Note: This database access is racey. Preferably there would be a
-    // row-lock held between the time it is loaded and the save call below.
-    // The `post_comment` call should normally be pretty fast, so races should
-    // be rare. If they happen too often, consider adding some locking
-    // mechanism here.
     let mut state: IssueData<'_, MergeConflictState> =
         IssueData::load(db, issue, MERGE_CONFLICT_KEY).await?;
     if state.data.last_warned_comment.is_some() {
--- a/src/handlers/merge_conflict.rs
+++ b/src/handlers/merge_conflict.rs
@@ -91,9 +91,9 @@ async fn handle_branch_push(
     let repo = push.repository.clone();
     let db = ctx.db.get().await;
     // Spawn since this can trigger a lot of work.
+    let gh = ctx.github.clone();
     tokio::task::spawn(async move {
         // See module note about locking.
-        let gh = GithubClient::new_from_env();
         if let Err(e) = scan_prs(&gh, db, &config, repo, &branch_name, &push_sha).await {
             log::error!("failed to scan PRs for merge conflicts: {e:?}");
         }
@@ -118,10 +118,10 @@ async fn handle_pr(
             let pr_number = issue.number;
             let db = ctx.db.get().await;
             let config = config.clone();
+            let gh = ctx.github.clone();
             tokio::task::spawn(async move {
                 // See module note about locking.
                 tokio::time::sleep(UNKNOWN_RESCAN_DELAY).await;
-                let gh = GithubClient::new_from_env();
                 if let Err(e) = rescan_pr(&gh, db, &config, repo, pr_number).await {
                     log::error!("failed to rescan PR for merge conflicts: {e:?}");
                 }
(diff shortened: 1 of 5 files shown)


Urgau:
Full support for having merge conflict warnings, love it :heart:.

On a more technical note, have you been able to test it?

ehuss:
Yes, here was my test plan:

- PR operations:
    - ✓ Normal PR open
    - ✓ PR open with a conflict
    - ✓ PR reopen with a conflict
    - ✓ Push to PR that generates a conflict
        - Verify it works synchronously, and with the unknown rescan delay.
    - ✓ Resolving a conflict in a PR hides the comment
        - ✓ And then getting a new conflict will post a new comment, which will also get hidden properly
- Push to branch from a PR
    - ✓ "possibly" should be filled out correctly
    - ✓ Doesn't warn on PRs for other branches
    - ✓ Doesn't warn on mergeable PRs
    - ✓ Doesn't warn on the PR that was just closed (is there a race between merging and closing the PR?)
    - ✓ Verify unknown scan works correctly
    - ✓ Verify graphql pagination works correctly
    - ✓ Doesn't post a comment to a PR that already has a comment
- ✓ Push to a branch without a PR, should use SHA in "possibly" message
- ✓ Labels get set correctly

src/handlers/merge_conflict.rs
```diff
@@ -0,0 +1,330 @@
...
+/// The amount of time to wait before scanning an unknown mergeable status.
+///
+/// GitHub has a background job which updates the mergeable status. We have to
+/// wait for it to be finished. Unfortunately there is no notification when it
+/// is done. It seems to usually run pretty quickly, but this timeout is set
+/// to be a little conservative just in case it takes a while to compute. I
+/// don't particularly want to loop to avoid hitting GitHub too hard, and this
+/// conflict notification is not that important to be perfect. If it is too
+/// unreliable, then we could add a loop that will try one or two more times.
+const UNKNOWN_RESCAN_DELAY: Duration = Duration::new(60, 0);
```
Mark-Simulacrum, line 46:
```suggestion
const UNKNOWN_RESCAN_DELAY: Duration = Duration::from_secs(60);
```

src/handlers/merge_conflict.rs
```diff
@@ -0,0 +1,330 @@
...
+    };
+    let branch_name = branch_name.to_string();
+    let push_sha = push.after.to_string();
+    let config = config.clone();
+    let repo = push.repository.clone();
+    let db = ctx.db.get().await;
+    // Spawn since this can trigger a lot of work.
+    tokio::task::spawn(async move {
+        // See module note about locking.
+        let gh = GithubClient::new_from_env();
```
Mark-Simulacrum, line 97: Is there a reason we need a new client vs. using the existing ctx.github client? Seems like that should be Cloneable?

src/handlers/merge_conflict.rs
```diff
@@ -0,0 +1,330 @@
...
+    db: &mut DbClient,
+    config: &MergeConflictConfig,
+    issue: &Issue,
+    possibly: Option<&str>,
+) -> anyhow::Result<()> {
+    // Note: This database access is racey. Preferably there would be a
+    // row-lock held between the time it is loaded and the save call below.
+    // The `post_comment` call should normally be pretty fast, so races should
+    // be rare. If they happen too often, consider adding some locking
+    // mechanism here.
```
Mark-Simulacrum, line 269: I think there *is* a lock held, actually on the full table? `load(...)` starts with `LOCK TABLE issue_data` and presumably that's kept until the transaction ends (which happens in save).

https://github.com/rust-lang/triagebot/blob/742b66bf475296212ebd913cfa909b1dbc747973/src/db/issue_data.rs#L41

Do we think we need more?

Mark-Simulacrum:
Generally onboard with landing mostly as-is, a few questions/comments.

src/handlers/merge_conflict.rs
ehuss: I don't remember what I was thinking at the time. I saw the LOCK TABLE call, but for some reason was thinking there was something else. I can't think of what it is now, and the lock is clearly held in the `IssueData`, so I removed the comment. Thanks for pointing that out!

src/handlers/merge_conflict.rs
ehuss: Yea, that sounds good! For some reason I was just assuming it wasn't cloneable (or maybe I was concerned about how the connection pooling would work?).

Mark-Simulacrum (approved):

PR #1858, opened 2024-11-18
Lock Major Change Proposal issue

Urgau:
This PR makes it so that every new major change proposal issue is locked by default.

This is done to affirm more strongly that discussion should happen on the generated Zulip thread instead.

This is particularly relevant for outsiders who *(I think)* just skip the part about no technical discussion on the issue, as it will systemically prevent them from doing so.

The locking of the issue doesn't affect "collaborators", aka members of the repo with at least "write" access (which is the case for all t-compiler/t-types members).

cc @apiraino
r? @ehuss

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -416,6 +416,18 @@ pub enum ReportedContentClassifiers {
     Spam,
 }
 
+#[derive(Debug, serde::Deserialize, serde::Serialize, Eq, PartialEq)]
+pub enum LockReason {
+    #[serde(rename = "off-topic")]
+    OffTopic,
+    #[serde(rename = "too heated")]
+    TooHeated,
+    #[serde(rename = "resolved")]
+    Resolved,
+    #[serde(rename = "spam")]
+    Spam,
+}
+
 #[derive(Debug, serde::Deserialize, Eq, PartialEq)]
 #[serde(rename_all = "snake_case")]
 pub enum PullRequestReviewState {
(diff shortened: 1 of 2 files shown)


Urgau:
> ... and our bots as well, I imagine. Correct?

Yes, they [also get "write" permissions](https://github.com/rust-lang/sync-team/blob/15ffe3d8ef58ca5723eb1c2f1df9a183c0eb0177/src/github/mod.rs#L525).

> [..] my remembering is we don't have very often people discussing on the issue itself and ignoring Zulip.

I agree it's not very recurring but when it happens we are put in a bad state where the comment will likely be missed and if it isn't missed we have to remind people to post it on Zulip which IMO greats a bad taste (especially for newcomers). It greats a disconnect between the two place, which this PR makes it much harder to happen for IMO a small cost.

src/handlers/major_change.rs
```diff
@@ -294,6 +294,10 @@ async fn handle(
             .post_comment(&ctx.github, &comment)
             .await
             .context("post major change comment")?;
+        issue
+            .lock(&ctx.github, None)
+            .await
+            .context("lock major change issue")?;
```
apiraino, line 300:
```suggestion
            .context("lock Major Change Proposal")?;
```

(I believe it's the official branding :upside_down_face: )

apiraino:
So, after thinking more about it, and after speaking with @Urgau again - I was not fully convinced about the implications of hard locking conversations.

Similar topic was raised by Oli [a while ago](https://rust-lang.zulipchat.com/#narrow/channel/233931-t-compiler.2Fmajor-changes/topic/Policy.3A.20no.20discussions.20on.20T-compiler.20trac.E2.80.A6.20compiler-team.23739/near/433137208). My concern at the time was *"if no discussions should happen on a tracking issue, where else?"*.

At the time the answer was unclear. Today - since a few weeks - we have sub-issues on Github so I think we can experiment with this and see the results.

src/github.rs
```diff
@@ -416,6 +416,18 @@ pub enum ReportedContentClassifiers {
     Spam,
 }
 
+#[derive(Debug, serde::Deserialize, serde::Serialize, Eq, PartialEq)]
+pub enum LockReason {
```
apiraino, line 420: Which of this reasons would apply to the use case "tracking issues are just not for discussions"?

Urgau: None of these apply/transmit the idea we want, so I wouldn't use using any of those reason. The reason is optional, I didn't use any of them for this PR.

apiraino:
Ok, so personally I am in favor of experimenting this, it's just for the Compiler MCPs so  a relatively small audience - which I think it's safe to assume is already mostly on Zulip, so probably my concern about "*where else would a discussion take place*" is unfounded.

I'll let T-compiler get a FIY about this.

apiraino:
Quick update: this was discussed in T-compiler triage meeting (on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202024-11-28/near/484942780)). The vibe is positive on testing this.

cc @ehuss when you have time for a quick look at this

thanks to both!

ehuss (approved):
Thanks!

Issue #1852, opened 2024-11-01
Teach rustbot about beta/stable PRs


cuviper:
I wonder if we could improve messages like this on backports: https://github.com/rust-lang/rust/pull/132468#issuecomment-2452371114

> ⚠️ **Warning** ⚠️
> 
> * Pull requests are usually filed against the master branch for this repo, but this one is against beta. Please double check that you specified the right target!

When the PR title starts with "[beta" or "[stable", we should instead warn if the PR is not targeting those specific branches.

(I left the strings open because some folks add a version in the brackets, like https://github.com/rust-lang/rust/pull/132337)

Urgau:
Looking at the recent backport PRs in rust-lang/rust I think #1860 successfully addressed this issue.

Is my assessment right @cuviper?

cuviper:
Yeah, it's not harassing me any more, thanks! :)

PR #1860, opened 2024-11-24
Add support for exceptions in non default branch warning

Urgau:
This PR adds support for specifying exceptions to the global default branch according to the PR title.

This is done to improve the usefulness of the non-default branch warning by making it warn if a PR trying to target the `stable` branch is actually targeting the `master` branch. 

```toml
[assign]
warn_non_default_branch.enable = true

[[assign.warn_non_default_branch.exceptions]]
title = "[beta"
branch = "beta"

[[assign.warn_non_default_branch.exceptions]]
title = "[stable"
branch = "stable"
```

cc https://github.com/rust-lang/triagebot/issues/1852 @cuviper 
r? @ehuss

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -91,10 +91,10 @@ pub(crate) struct PingTeamConfig {
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
 #[serde(deny_unknown_fields)]
 pub(crate) struct AssignConfig {
-    /// If `true`, then posts a warning comment if the PR is opened against a
+    /// If enabled, then posts a warning comment if the PR is opened against a
     /// different branch than the default (usually master or main).
     #[serde(default)]
-    pub(crate) warn_non_default_branch: bool,
+    pub(crate) warn_non_default_branch: WarnNonDefaultBranchConfig,
     /// A URL to include in the welcome message.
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -118,6 +118,45 @@ impl AssignConfig {
...
+#[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+#[serde(untagged)]
+pub(crate) enum WarnNonDefaultBranchConfig {
+    Simple(bool),
+    Extended {
+        enable: bool,
+        /// List of exceptions that have a different default branch
+        #[serde(default)]
+        exceptions: Vec<WarnNonDefaultBranchException>,
```
Urgau, line 130: We could also use a `HashMap<String, WarnNonDefaultBranchException>` where the string would be the branch name, which would permit this and be more consistent with the rest of the configs, but we would losz the "exceptions" field. I'm not sure which one is better.
```toml
[assign.warn_non_default_branch]
enable = true

[assign.warn_non_default_branch."beta"]
title = "[beta"

[assign.warn_non_default_branch."stable"]
title = "[stable"
```

ehuss (approved):
Thanks!

PR #1861, opened 2024-11-28
fix: handle cargo milestone update for PR from GitHub merge queue

weihanglo:
Cargo has migrated to GitHub merge queue:
https://github.com/rust-lang/cargo/pull/14718

The regex doesn't work for merge commit made by GitHub merge queue.
This PR adds extra regex pattern to capture that.

This regex has been proved working locally.
See <https://github.com/ehuss/cargo-new-release/pull/2>

cc @ehuss


src/handlers/milestone_prs.rs
```diff
@@ -145,24 +145,34 @@ async fn milestone_cargo(
...
+    let merge_re =
+        Regex::new(r"(?:Auto merge of|Merge pull request) #([0-9]+)|\(#([0-9]+)\)$").unwrap();
+
+    let pr_nums = commits
+        .iter()
+        .filter(|commit|
+            // Assumptions:
+            // * A merge commit always has two parent commits.
+            // * Cargo's PR never got merged as fast-forward / rebase / squash merge.
+            commit.parents.len() == 2)
```
weihanglo, line 159: While the regex below has been proved working, this filter hasn't. This may not be necessary. I added because for preventive purpose.

src/handlers/milestone_prs.rs
```diff
@@ -145,24 +145,34 @@ async fn milestone_cargo(
...
+        .filter_map(|commit| {
+            log::info!(
+                "getting PR number for cargo commit {} (len={})",
+                commit.sha,
+                commit.commit.message.len()
+            );
+
+            merge_re.captures(&commit.commit.message).map(|cap| {
```
ehuss, line 167: I'm wondering, would it be a good idea to only look at the first line of the commit message? I'm wondering what would happen if there is a stray mention of another issue/PR in the body somewhere, like:

```
fix(fix): Migrate cargo script manifests across editions (#14864)

This is a multi-line description
that happens to mention an issue (#1234)
somewhere in its text.
```

ehuss: Looking at this closer, this regex doesn't seem to work with a multi-line commit message. For example:

```
Improve error handling when PathSource is relative (#14854)

When editing dependencies with cargo, if a relative PathSource is
supplied cargo panics with "both paths are absolute". This is the
opposite of what's actually wrong leading to confusion.

Instead, use the same error formatting we use in other diff_paths calls
and return that error instead of panicking.
```

This is not matching here, since the `$` requires the `(#14854)` to be at the end of the message (since regex does not have mutliline enabled by default).

weihanglo: Hah! I overlooked that difference. Thanks for investigating I'll update to check the first line only then.

weihanglo: Addressed in 7229b2a8a9b7b4b1dbd5b2c2c442d6535f101ad5!

ehuss (approved):
Thanks! I pushed a small change to remove a panic if the commit message is empty. Although unlikely (or impossible?) to happen, I didn't feel comfortable crashing triagebot if it does.

PR #1863, opened 2024-12-04
Fix MCP concern syntax hint

apiraino:
Fixing a typo when suggesting the syntax to raise concerns on compiler MCPs (template was implemented in #1747)

### DIFF
--- a/src/handlers/major_change.rs
+++ b/src/handlers/major_change.rs
@@ -270,7 +270,7 @@ async fn handle(
         \n \
         ``` \
         \n \
-        @rustbot concern reason-for-concern \
+        @rfcbot concern reason-for-concern \
         \n \
         <description of the concern> \
         \n \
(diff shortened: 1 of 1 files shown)


ehuss:
Does rfcbot just ignore these commands because it is not enabled on the repo?

What happens if you do `@rustbot concern` on it?

ehuss:
I'm confused. The link to that comment uses `@rustbot`. Nothing in that PR seems to mention rfcbot.
Does rfcbot also ignore it?
Do you want to change this just to be more consistent with the use of rfcbot in other repositories?

apiraino:
@rfcbot concern this-is-a-test

apiraino:
Ok, no user-visible effects. But the logic in the triagebot to gather concerns needs the `@rfcbot` invocation,  just as [I had implemented](https://github.com/rust-lang/triagebot/blob/4b28086e6b7e7566534a8e936ad9b9acf6105a6c/src/github.rs#L2117) it since the beginning

ehuss (approved):
Ok, I see, thanks!

Issue #1864, opened 2024-12-05
Find some way to work around github not running CI on GHA-created PRs


RalfJung:
PRs created in GHA do not trigger CI. That is a problem when combined with merge queues, where a PR will only be considered ready-to-land when CI passes. The best work-around I know of so far is to close and re-open the PR, which *will* trigger CI.

It might make sense to have a bot do the close-and-reopen automatically, just so that CI has generally already run by the time a human looks at the PR.

RalfJung:
Thanks!

Do we need to configure something to enable this in Miri, or should it "Just Work (TM)"?

Mark-Simulacrum:
You'll need to add the empty table to config.toml, the newly added forge docs should have an example.

RalfJung:
Unfortunately this does not seem to work yet: https://github.com/rust-lang/miri/pull/4085.

FWIW another possible modus operandi here might be for the PR description itself to contain `@rustbot reopen` as a signal to please close and reopen the PR.

RalfJung:
In https://github.com/rust-lang/miri/pull/4088 it worked, amazing :) Thanks a lot!

PR #1866, opened 2024-12-08
Add github-action PR open/closer

Mark-Simulacrum:
This isn't tested, but my hope is that it works ~first try or after small tweaks. I intentionally made it relatively safe (I hope) in terms of not falling into retry loops calling itself, but open to either dropping that or discussing alternatives.

cc @RalfJung 

Closes https://github.com/rust-lang/triagebot/issues/1864

Sample PR: https://github.com/rust-lang/miri/pull/4081

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -189,6 +189,31 @@ impl GithubClient {
         .await
         .context("failed to create issue")
     }
+
+    pub(crate) async fn set_pr_state(
+        &self,
+        repo: &IssueRepository,
+        number: u64,
+        state: PrState,
+    ) -> anyhow::Result<()> {
+        #[derive(serde::Serialize)]
+        struct Update {
+            state: PrState,
+        }
+        let url = format!("{}/pulls/{number}", repo.url(&self));
+        self.send_req(self.patch(&url).json(&Update { state }))
+            .await
+            .context("failed to update pr state")?;
+        Ok(())
+    }
+}
+
+#[derive(Debug, serde::Serialize)]
+pub(crate) enum PrState {
+    #[serde(rename = "open")]
+    Open,
+    #[serde(rename = "closed")]
+    Closed,
 }
 
 #[derive(Debug, serde::Deserialize)]
--- a/src/handlers/bot_pull_requests.rs
+++ b/src/handlers/bot_pull_requests.rs
@@ -1,0 +1,40 @@
+use crate::github::{IssuesAction, PrState};
+use crate::{github::Event, handlers::Context};
+
+pub(crate) async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
+    let Event::Issue(event) = event else {
+        return Ok(());
+    };
+    // Note that this filters out reopened too, which is what we'd expect when we set the state
+    // back to opened after closing.
+    if event.action != IssuesAction::Opened {
+        return Ok(());
+    }
+    if !event.issue.is_pr() {
+        return Ok(());
+    }
+
+    // avoid acting on our own open events, otherwise we'll infinitely loop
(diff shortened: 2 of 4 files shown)


src/handlers/bot_pull_requests.rs
```diff
@@ -0,0 +1,66 @@
...
+    };
+    if event.action != IssuesAction::Opened {
+        return Ok(());
+    }
+    if !event.issue.is_pr() {
+        return Ok(());
+    }
+
+    // avoid acting on our own open events, otherwise we'll infinitely loop
+    if event.sender.login == ctx.username {
```
Kobzol, line 20: I think that when the PR is reopened, it will trigger the "reopened" event, and not the "opened" event (https://docs.github.com/en/rest/using-the-rest-api/issue-event-types?apiVersion=2022-11-28#reopened). But it doesn't hurt to be extra safe here.

src/handlers/bot_pull_requests.rs
```diff
@@ -0,0 +1,66 @@
...
+        return Ok(());
+    }
+
+    // If it's not the github-actions bot, we don't expect this handler to be needed. Skip the
+    // event.
+    if event.sender.login != "github-actions" {
```
Kobzol, line 26: I'm not sure if this will work, an user with such login doesn't seem to exist? https://github.com/github-actions

I tried to run this on https://github.com/rust-lang/miri/pull/4070:
image

It looks like the login is `app/github-actions`. Ideally we should also check if it's a bot.

src/handlers/bot_pull_requests.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    if DISABLE.load(std::sync::atomic::Ordering::Relaxed) {
+        tracing::warn!("skipping bot_pull_requests handler due to previous disable",);
+        return Ok(());
+    }
+
+    // Sanity check that our logic above doesn't cause us to act on PRs in a loop, by
+    // tracking a window of PRs we've acted on. We can probably drop this if we don't see problems
+    // in the first few days/weeks of deployment.
+    {
```
Kobzol, line 38: I think/hope that this logic should be unnecessary due to the difference between the opened/reopened events.

Kobzol (commented):
Left a few comments. Thanks a lot for implementing this!

src/handlers/bot_pull_requests.rs
Mark-Simulacrum: Yeah, I missed that those are different - I think I'll go ahead and delete the extra logic since it's not trivial then.

src/handlers/bot_pull_requests.rs
Mark-Simulacrum: Hm, I'm not sure if we get is_bot = true in the `sender` field, I'd have to go look at events we've seen. I've updated to app/github-actions, and we can revisit if this doesn't work in practice IMO.

src/github.rs
```diff
@@ -189,6 +189,31 @@ impl GithubClient {
...
+        &self,
+        repo: &IssueRepository,
+        number: u64,
+        status: PrStatus,
+    ) -> anyhow::Result<()> {
+        #[derive(serde::Serialize)]
+        struct Update {
+            status: PrStatus,
+        }
+        let url = format!("{}/pulls/{number}", repo.url(&self));
```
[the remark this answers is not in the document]
Mark-Simulacrum: Fixed.

Kobzol:
I tested this locally on my bors-kindergarten repo. The PATCH endpoint with just "state=closed" and "state=open" worked, and it has generated a "reopened" event. So looks good to me.

ehuss (approved):
Thanks!

PR #1867, opened 2024-12-12
Fix sender login

Mark-Simulacrum:
This is pulled from a sample event which had { login: "github-actions[bot]", id: 41898282 }

cc https://github.com/rust-lang/triagebot/issues/1864#issuecomment-2533735743


ehuss (approved):
Thanks! Confirmed in API that is the username.

PR #1868, opened 2024-12-17
Add proper config for rendered link handler

Urgau:
This PR adds proper config for the rendered link handler. This is done to allow more repositories (as well as the current ones) to configure by them selves the handler.

```toml
[rendered-link]
trigger-files = ["posts/"]
```

r? @ehuss


ehuss:
Thanks, can you open PRs to rfcs and blog.rust-lang.org with the new config before we merge this?

(triagebot will probably complain about unknown fields, but we can ignore that.)

Urgau:
Sure, I have opened https://github.com/rust-lang/blog.rust-lang.org/pull/1452 and https://github.com/rust-lang/rfcs/pull/3758; and as supposed triagebot complained (gave an error) for both PRs.

ehuss (approved):
Thanks!

Issue #1832, opened 2024-08-16
`r?` requests should ignore zero-width spaces


tgross35:
There are some cases where it looks like Rustbot doesn't pick up review requests:

- https://github.com/rust-lang/rust/pull/128862
- https://github.com/rust-lang/rust/pull/128868
- https://github.com/rust-lang/rust/pull/129003

@ehuss noticed [on zulip](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/Is.20rustbot.20up.3F) that these comments contain a zero-width space. I am not sure where this would be coming from - maybe GitHub adds them when you autocomplete an `@` name - but rustbot should probably work around this.

apiraino:
What the GH comment contains
```
$ hexdump -C test.txt 
00000000  72 e2 80 8b 3f 20 40 73  63 6f 74 74 6d 63 6d 0a  |r...? @scottmcm.|
00000010
```
What I expect to the text to be
```
$ hexdump -C test1.txt 
00000000  72 3f 20 40 73 63 6f 74  74 6d 63 6d 0a           |r? @scottmcm.|
```

So yes that [U+200B](https://codepoints.net/U+200B?lang=en) is added somehow. Tried [adding a comment](https://github.com/rust-lang/rust/pull/124081#issuecomment-2294102286) to a PR letting GH autocomplete my username and it looks fine.

Seems unlikely, but I wonder if these weird whitespaces are added by the user client when opening the PR.

tgross35:
I have no clue, one of those is mine and I didn't do anything out of the ordinary. Spurious GH bug?

apiraino:
ok @tgross35 , the mistery is solved. Our issue github template had a bug.

Fixed by [rust#131151](https://github.com/rust-lang/rust/pull/131151)

I saw this [Zulip comment](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/feedback.20on.20github.20issues.20improvements/near/475231381).

jieyouxu:
> Our issue github template had a bug.

Note that this is not a bug, this was actually a feature: it was introduced in https://github.com/rust-lang/rust/pull/123492#discussion_r1595820514 because apparently rustbot gets confused if people didnt' remove the html comment.

Urgau:
I think this issue can be closed as the root cause as been fixed by #1869 and https://github.com/rust-lang/rust/pull/135189.

PR #1869, opened 2024-12-18
Ignore inline HTML and HTML blocks

Urgau:
This PR adds inline HTML (`<a>`) and HTML blocks to the ignore logic when parsing commands.

This is particularly useful for rust-lang/rust [pull request template](https://github.com/rust-lang/rust/blob/master/.github/pull_request_template.md?plain=1) which currently hacks around triagebot finding `r?` in the HTML comment.

This PR also updates `pulldown-cmark` to `0.12`.

Related to https://github.com/rust-lang/triagebot/issues/1832
r? @ehuss

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1738,15 +1738,22 @@ dependencies = [
 
 [[package]]
 name = "pulldown-cmark"
-version = "0.7.2"
+version = "0.12.2"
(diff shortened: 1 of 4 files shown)


ehuss (approved):
Thanks!

PR #1872, opened 2024-12-27
Read Zulip config from env vars

apiraino:
Sometimes we want to test the triagebot against a test Zulip instance (f.e. #1853) but currently the Zulip config is hardcoded. This patch allows settings env vars to use a custom Zulip instance. If these env vars are not set, the triagebot will work as usual and connect to the production Zulip instance.

I've tested this with a binary that I didn't include in the repository, it simply uses the public api is `zulip.rs`:
<details><summary>code</summary>
<p>

```
use triagebot::github::GithubClient;
use triagebot::zulip;

#[tokio::main(flavor = "current_thread")]
async fn main() -> anyhow::Result<()> {
    let gh = GithubClient::new_from_env();

    let zulip_topic_name = "New topic";
    let message = "This is a message";

    let zulip_req = crate::zulip::MessageApiRequest {
        recipient: crate::zulip::Recipient::Stream {
            id: 123456, // #ChannelName
            topic: &zulip_topic_name,
        },
        content: &message,
    };

    // .send() reads env vars ZULIP_URL, ZULIP_API_TOKEN, ZULIP_BOT_EMAIL
    let response = zulip_req.send(&gh.raw()).await?;
    println!("RESP: {:?}", response.text().await);

    Ok(())
}
```

</p>
</details> 

maybe r? @ehuss  (or anyone who has time for a review)

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -9,7 +9,14 @@ use anyhow::{format_err, Context as _};
 use std::env;
 use std::fmt::Write as _;
 use std::str::FromStr;
+use std::sync::LazyLock;
 use tracing as log;
+
+static ZULIP_URL: LazyLock<String> =
+    LazyLock::new(|| env::var("ZULIP_URL").unwrap_or("https://rust-lang.zulipchat.com".into()));
+static ZULIP_BOT_EMAIL: LazyLock<String> = LazyLock::new(|| {
+    env::var("ZULIP_BOT_EMAIL").unwrap_or("triage-rust-lang-bot@zulipchat.com".into())
+});
 
 #[derive(Debug, serde::Deserialize)]
 pub struct Request {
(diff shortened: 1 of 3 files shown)


src/zulip.rs
```diff
@@ -71,8 +75,6 @@ struct Response {
     content: String,
 }
 
-pub const BOT_EMAIL: &str = "triage-rust-lang-bot@zulipchat.com";
```
apiraino, line 74: I wanted to also set this global but then it needs to be cloned multiple times, so unsure what's the best solution

ehuss: I'm not sure I see the concern about needing to be cloned. If you do it in a LazyLock like you did the URL, then you can do `&*ZULIP_BOT_EMAIL` to get a reference. The `basic_auth` method does not need an owned value.

An alternative would be to place all these environment values inside a structure stored in the `Context`, and just pass that around as an argument (or pass in the whole Context).

apiraino: thank you for a second opinion! I've amended the code with just adding another LazyLock to keep changes as contained here as possible.

ehuss (approved):
Thanks!

apiraino:
thanks @ehuss !

PR #1873, opened 2024-12-29
fix(transfer): dont create labels if missing

weihanglo:
This is quite annoying if the target repo doesn't need those labels, or have a completely different label management approach.

While issues need to relabel anyway after transfer, it requires more steps to delete a label. I don't think it is worth.

Alternatively, we could have an argument `@rustbot transfer cargo createMissingLabels=1` in the future, if people really want missing labels created.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -1055,7 +1055,7 @@ impl Issue {
             .graphql_query(
                 "mutation ($issueId: ID!, $repoId: ID!) {
                   transferIssue(
-                    input: {createLabelsIfMissing: true, issueId: $issueId, repositoryId: $repoId}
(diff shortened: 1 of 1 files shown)


weihanglo:
Another alternative: configuring via the `[transfer]` table.

```toml
[transfer]
create-labels-if-missing = true
```

apiraino:
This was introduced in #1780 and I don't see a specific discussion about labels, therefore I think it's ok to experiment with *not* migrating labels. I guess the reasoning is that labelling is very specific to the git repository and the workflow of maintainers working there, so there is very little overlap between codebases about labels.

ehuss (approved):
Thanks! I don't remember what happened here. That was not the behavior I intended. Perhaps I was testing things and forgot to change it back?

PR #1879, opened 2025-01-07
Add work queue tracking in triagebot DB

apiraino:
Rebase of #1786

This implements the new way of assigning pull requests reviews (designed in #1753). When this PR is merged, PRs in `rust-lang/rust` will be assigned based on user preferences. These can be queried from Zulip (see [docs](https://forge.rust-lang.org/triagebot/pr-assignment-tracking.html)).

If no preferences are set, PRs will be just assigned (like always).

There are multiple points where PRs are assigned (via comment or using the GH UI), I've tried to cover all cases, took some time to test them all. The amount of changes reflect that it is not so easy to catch all spots where this happens. I wish I could make a smaller patch but I think this is an _all-or-nothing_ situation where all cases should be handled at once.

Also modified the zulip triagebot command `work show` to return the # work queue capacity set by the user. This is helpful in case a PR assignment is rejected. A PR to update the [documentation](https://forge.rust-lang.org/triagebot/pr-assignment-tracking.html) will follow-up after merge of this.

If a PR assignment is rejected we will return a message with some context. Examples:

Failed PR assignment to a user:
screenshot-20240403-104629

Failed self-assignment:
screenshot-20240403-104717

r? @jackh726

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -728,7 +728,6 @@ impl fmt::Display for FindReviewerError {
                     f,
                     "Could not assign reviewer from: `{}`.\n\
                      User(s) `{}` are either the PR author, already assigned, or on vacation. \
-                     If it's a team, we could not find any candidates.\n\
                      Please use `r?` to specify someone else to assign.",
                     initial.join(","),
                     filtered.join(","),
@@ -820,14 +819,11 @@ SELECT username
 FROM review_prefs r
 JOIN users on users.user_id=r.user_id
 AND username = ANY('{{ {} }}')
-AND ARRAY_LENGTH(r.assigned_prs, 1) < max_assigned_prs",
+AND CARDINALITY(r.assigned_prs) < LEAST(COALESCE(r.max_assigned_prs,1000000))",
         usernames
     );
     let result = db.query(&q, &[]).await.context("Select DB error")?;
-    let mut candidates: HashSet<String> = HashSet::new();
-    for row in result {
-        candidates.insert(row.get("username"));
-    }
+    let candidates: HashSet<String> = result.iter().map(|row| row.get("username")).collect();
     Ok(candidates)
 }
 
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -7,6 +7,9 @@
 //! * `@rustbot release-assignment`: Removes the commenter's assignment.
 //! * `r? @user`: Assigns to the given user (PRs only).
 //!
+//! Note: this module does not handle review assignments issued from the
+//! GitHub "Assignees" dropdown menu
+//!
 //! This is capable of assigning to any user, even if they do not have write
 //! access to the repo. It does this by fake-assigning the bot and adding a
 //! "claimed by" section to the top-level comment.
@@ -20,7 +23,7 @@
 use crate::{
     config::{AssignConfig, WarnNonDefaultBranchException},
     github::{self, Event, FileDiff, Issue, IssuesAction, Selection},
-    handlers::{Context, GithubClient, IssuesEvent},
+    handlers::{pr_tracking::has_user_capacity, Context, GithubClient, IssuesEvent},
     interactions::EditIssueBody,
 };
 use anyhow::{bail, Context as _};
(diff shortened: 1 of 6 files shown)


src/handlers/assign.rs
```diff
@@ -708,14 +781,57 @@ fn find_reviewer_from_names(
...
+        .collect::<Vec<&str>>()
+        .join(",");
+
+    let q = format!(
+        "
+SELECT username
+FROM review_prefs r
+JOIN users on users.user_id=r.user_id
+AND username = ANY('{{ {} }}')
+AND CARDINALITY(r.assigned_prs) < LEAST(COALESCE(max_assigned_prs,1000000))",
```
apiraino, line 823: A note about this SQL dance, it is a safeguard against *null* fields to be able to do a number comparison. [CARDINALITY](https://www.postgresql.org/docs/16/arrays.html#ARRAYS-ACCESSING) returns the number or array items even when empty, `LEAST(COALESCE(...))` returns a number when `max_assigned_prs` is *null*.

jackh726: Does this need to be `r.max_assigned_prs`? (If not, should it for clarity?)

src/handlers/assign.rs
```diff
@@ -708,14 +781,57 @@ fn find_reviewer_from_names(
...
+        usernames
+    );
+    let result = db.query(&q, &[]).await.context("Select DB error")?;
+    let mut candidates: HashSet<String> = HashSet::new();
```
jackh726, line 827: Nit: you could initialize with the known capacity?

src/handlers/pr_tracking.rs
```diff
@@ -66,18 +70,60 @@ pub(super) async fn handle_input<'a>(
     if matches!(event.action, IssuesAction::Unassigned { .. }) {
         delete_pr_from_workqueue(&db_client, assignee.id, event.issue.number)
             .await
-            .context("Failed to remove PR from workqueue")?;
+            .context("Failed to remove PR from workq ueue")?;
```
jackh726, line 73: Nit: typo

src/handlers/assign.rs
```diff
@@ -665,13 +727,23 @@ impl fmt::Display for FindReviewerError {
                 write!(
                     f,
                     "Could not assign reviewer from: `{}`.\n\
-                     User(s) `{}` are either the PR author, already assigned, or on vacation, \
-                     and there are no other candidates.\n\
-                     Use `r?` to specify someone else to assign.",
+                     User(s) `{}` are either the PR author, already assigned, or on vacation. \
+                     If it's a team, we could not find any candidates.\n\
```
jackh726, line 731: Not needed, but we could just be smart and use a message tailored to if it was a team or not.

src/handlers/pr_tracking.rs
```diff
@@ -66,18 +70,60 @@ pub(super) async fn handle_input<'a>(
...
+pub async fn has_user_capacity(
+    db: &crate::db::PooledClient,
+    assignee: &str,
+) -> anyhow::Result<ReviewPrefs, FindReviewerError> {
+    let q = "
+SELECT username, r.*
+FROM review_prefs r
+JOIN users ON users.user_id = r.user_id
+WHERE username = $1
+AND CARDINALITY(r.assigned_prs) < LEAST(COALESCE(max_assigned_prs,1000000));";
```
jackh726, line 117: As above, r.max_assigned_prs?

jackh726: So, I can somewhat see an argument for having the check be done the same as when fetching users, but it's a bit weird that we don't just fetch the number of assigned prs and max assigned prs and check it outside the db.

For consistency, I think let's keep it this way, but put a comment.

jackh726 (changes requested):
Just double checking, we don't have to do anything fancy with the db before merging, right

src/handlers/assign.rs
apiraino: Needs not since it's not ambigous but good for clarity :+1:

src/handlers/assign.rs
apiraino: Now that I read this message again, I think mentioning a team there doesn't make a lot of sense. The "team not found" reason should be already covered by the enum item `FindReviewerError::TeamNotFound`.

Think I'll just remove that sentence.

jackh726 (approved):
We talked a bit in DMs - there are certainly things we could do to minimize the risk (bubble up potential db failures instead of `expect`ing, land the db changes separately, fallbacking to existing non-filtering if db failture), but these aren't needed.

Let's land this.

PR #1884, opened 2025-01-18
Update release notes template

BoxyUwU:
I wanted us to have a more exhaustive list of what the acceptable categories were. I also think it would be helpful to have a list of prompts as to what extra categories might be required so that its harder to forget. Also, link to forge docs about the relnotes process as its not very clear to people outside the team

### DIFF
--- a/src/handlers/relnotes.rs
+++ b/src/handlers/relnotes.rs
@@ -75,36 +75,34 @@ pub async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
                 "
 This issue tracks the release notes text for #{}.
 
-### Steps
+cc {} -- original issue/PR authors and assignees for drafting text
 
-- [ ] Proposed text is drafted by PR author (or team) making the noteworthy change.
-- [ ] Issue is nominated for release team review of clarity for wider audience.
-- [ ] Release team includes text in release notes/blog posts.
+See the forge.rust-lang.org chapter about [release notes](https://forge.rust-lang.org/release/release-notes.html#preparing-release-notes) for an overview of how the release team makes use of these tracking issues.
 
 ### Release notes text
 
-The responsible team for the underlying change should edit this section to replace the automatically generated link with a succinct description of what changed, drawing upon text proposed by the author (either in discussion or through direct editing).
+This section should be edited to specify the correct category(s) for the change, with succinct description(s) of what changed. Some things worth considering:
+- Does this need an additional compat notes section?
+- Was this a libs stabilization that should have additional headers to list new APIs under `Stabilized APIs` and `Const Stabilized APIs`?
+
 
 ````markdown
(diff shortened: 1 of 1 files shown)


src/handlers/relnotes.rs
```diff
@@ -75,36 +75,34 @@ pub async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
...
->
-> *More than one section can be included if needed.*
 
 ### Release blog section
 
-If the change is notable enough for inclusion in the blog post, the responsible team should add content to this section.
-*Otherwise leave it empty.*
+If this change is notable enough for inclusion in the blog post then this section should be edited to contain a draft for the blog post. *Otherwise leave it empty.*
+
+**Note:** If a blog post section is required the `release-blog-post` label should be added to this issue as otherwise it may be missed by the release team.
```
BoxyUwU, line 101: This label doesn't exist yet but I will add it if this PR is merged, I think it would be helpful to be able to track which issues had drafts written so they aren't potentially missed when the blog post is written.

BoxyUwU (commented):
.

src/handlers/relnotes.rs
Mark-Simulacrum: Maybe we should give the triagebot command + make sure it's allow listed for anyone to set (not just team members)?

Also probably good to edit the release notes process on forge to match.

BoxyUwU: Ah right, I forget not everyone has triage perms

src/handlers/relnotes.rs
```diff
@@ -75,36 +75,34 @@ pub async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
...
+
 
 ````markdown
-# Category (e.g. Language, Compiler, Libraries, Compatibility notes, ...)
+# Language/Compiler/Libraries/Stabilized APIs/Const Stabilized APIs/Rustdoc/Compatibility Notes/Internal Changes/Other
 - [{}]({})
 ````
 
 > [!TIP]
 > Use the [previous releases](https://doc.rust-lang.org/nightly/releases.html) categories to help choose which one(s) to use.
```
Urgau, line 95:
```suggestion
> Use the [previous releases](https://doc.rust-lang.org/nightly/releases.html) for inspiration on how to write the release notes text.
```

src/handlers/relnotes.rs
```diff
@@ -75,36 +75,36 @@ pub async fn handle(ctx: &Context, event: &Event) -> anyhow::Result<()> {
...
```
BoxyUwU, line 95:
```suggestion
> Use the [previous releases](https://doc.rust-lang.org/nightly/releases.html) for inspiration on how to write the release notes text and which categories to pick.
```

Mark-Simulacrum (approved):

PR #1891, opened 2025-01-24
Improve query for getting users for review

Kobzol:
The previous query had a problem where if some user was missing in the `review_prefs` table, it would not be returned from the query. This is problematic, because for several reasons, the `review_prefs` table might be incomplete, and in general, it should be possible to get users even if they are not in the `review_prefs` table.

There is one problem with the new query though, because there is no index on the `users.username` column, this will result in a sequential scan on the `users` table. It's probably still fast enough, but I wanted to mention it.

The `LEAST` clause was not required, I think. The added `COALESCE` is needed to treat the number of PRs for people missing in `review_prefs` as zero, otherwise they would be again ignored.

You can experiment with this in https://onecompiler.com/postgresql/43722tb3w.

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -820,13 +820,21 @@ async fn filter_by_capacity(
         .collect::<Vec<&str>>()
         .join(",");
 
+    // We need to select everyone from the users database here, even if
+    // they might be missing in the review_prefs table.
+    // Therefore, we use a LEFT JOIN to get ALL users plus their associated data
+    // from the review_prefs, if there are any.
+    //
+    // Then we filter users whose assigned PRs are large enough.
+    // The final WHERE clause filters only users from the input list.
     let q = format!(
         "
-SELECT username
-FROM review_prefs r
-JOIN users on users.user_id=r.user_id
-AND username = ANY('{{ {} }}')
-AND CARDINALITY(r.assigned_prs) < LEAST(COALESCE(r.max_assigned_prs,1000000))",
+SELECT u.username AS username
+FROM users AS u
+LEFT JOIN review_prefs AS r ON u.user_id = r.user_id
+    AND COALESCE(CARDINALITY(r.assigned_prs), 0) < COALESCE(r.max_assigned_prs, 100000)
+WHERE u.username = ANY('{{ {} }}')
+",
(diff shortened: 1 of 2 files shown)


apiraino:
> The `LEAST` clause was not required, I think.

So in my [comment](https://github.com/rust-lang/triagebot/pull/1879#discussion_r1907879153) I was wrong about `LEAST`, right? :thinking:

Kobzol:
`COALESCE` should return the first non-null argument, so it should be enough to do `COALESCE(NULL, 100000)`. I don't think that you need `LEAST`, if the first argument to `COALESCE` is non-NULL, it should be returned.

src/handlers/assign.rs
```diff
@@ -820,13 +820,21 @@ async fn filter_by_capacity(
...
-FROM review_prefs r
-JOIN users on users.user_id=r.user_id
-AND username = ANY('{{ {} }}')
-AND CARDINALITY(r.assigned_prs) < LEAST(COALESCE(r.max_assigned_prs,1000000))",
+SELECT u.username AS username
+FROM users AS u
+LEFT JOIN review_prefs AS r ON u.user_id = r.user_id
+    AND COALESCE(CARDINALITY(r.assigned_prs), 0) < COALESCE(r.max_assigned_prs, 100000)
+WHERE u.username = ANY('{{ {} }}')
+",
```
apiraino, line 837: Just an observation: before your change, this function didn't return any record if no "joined" user was found. Changing the query as you suggest also changes the API, now it always return something so the caller **must** check if the returned candidate actually has capacity.

There is a TODO at the top of [has_user_capacity()](https://github.com/rust-lang/triagebot/blob/de42aaf47905b847b162f9c5555499c6bb8ace9b/src/handlers/pr_tracking.rs#L108) (which does more or less the same work) and that was implemented in #1893. I'll probably split that patch to make it clearer.

Kobzol: Right, so my query assumes that the only way the capacity could be affected is through the review_prefs column. Which I assume should be true? Apart from the on vacation list, that has to be checked elsewhere (until we get rid of it).

In other words, if you don't have an entry in review_prefs, then you have capacity.

PR #1893, opened 2025-01-27
Disable new PR assignment, just log

apiraino:
Following discussion on Zulip, I am going to disable the new PR assignment and just log its behaviour.

r? @Kobzol

cc @jackh726


jackh726 (approved):
@apiraino I'm assuming we didn't hit an expect or something because that would have been logged, right?

apiraino:
> @apiraino I'm assuming we didn't hit an expect or something because that would have been logged, right?

Yes, the only important `expect` I can see here is if the DB query fails being run (which would be a bigger issue)

PR #1895, opened 2025-01-29
Better tracking when candidates for PR assignment are filtered out

apiraino:
This PR contains:
- adds more logging when no candidates for a PR assignment are found
- split and logs the exact reason why a user was filtered out from a list of possible candidates

Besides these changes, nothing in the PR assignment changes. Overall, this should help us getting more insights from the logs. I've tried splitting the changes in commits to make the review easier.

Thanks!

r? @Kobzol or @jackh726

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -107,9 +107,14 @@ Please contact us on [#t-infra](https://rust-lang.zulipchat.com/#narrow/stream/2
 
 cc: @jackh726 @apiraino";
 
-fn on_vacation_msg(user: &str) -> String {
-    ON_VACATION_WARNING.replace("{username}", user)
-}
+const REVIEWER_IS_PR_AUTHOR: &str = "Pull request author cannot be assigned as reviewer.
+
+Please choose another assignee.";
+
+const REVIEWER_ALREADY_ASSIGNED: &str =
+    "Requested reviewer is already assigned to this pull request.
+
+Please choose another assignee.";
 
 #[derive(Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 struct AssignData {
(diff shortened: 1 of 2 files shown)


src/handlers/assign.rs
```diff
@@ -508,7 +510,10 @@ pub(super) async fn handle_command(
                 {
                     // This is a comment, so there must already be a reviewer assigned. No need to assign anyone else.
                     issue
-                        .post_comment(&ctx.github, &on_vacation_msg(&username))
```
apiraino, line 511: the deletion of `on_vacation_msg` is in 82c549a7b397079b73449dc4e4e181da2e6b174b (sorry)

src/handlers/assign.rs
```diff
@@ -804,14 +809,18 @@ async fn find_reviewer_from_names(
...
-        log::info!("Filtered list of PR assignee is empty");
+        log::info!("[#{}] Filtered list of PR assignee is empty", issue.number);
         // return Err(FindReviewerError::AllReviewersFiltered {
         //     initial: names.to_vec(),
         //     filtered: names.to_vec(),
         // });
     }
 
-    log::info!("Filtered list of candidates: {:?}", filtered_candidates);
+    log::info!(
```
Kobzol, line 856: We should really migrate to tracing and use spans to get have proper log context :)

src/handlers/assign.rs
Kobzol: I find the functions better than the consts, because with the function you cannot forget to pass the required placeholders, with the static string you can easily just use it without replacing the placeholder.

Kobzol (approved):
The `filtered_debug`  variable doesn't seem to be used? But anyway, this PR should hopefully only improve logging.

Issue #1, opened 2019-02-10
command testing


Mark-Simulacrum:

Mark-Simulacrum:
label: -T-libs

Mark-Simulacrum:
label: +bug -duplicate question

Mark-Simulacrum:
label: foobar

Mark-Simulacrum:
label: foobar

Mark-Simulacrum:
@rust-highfive modify labels to +bug -question -wontfix

Mark-Simulacrum:
@rust-highfive modify labels to -bug, +question, and +wontfix.

Mark-Simulacrum:
@rust-highfive modify labels -bug, +question, and +wontfix.

Mark-Simulacrum:
@rust-highfive modify labels: -bug, +question, and +wontfix.

PR #1897, opened 2025-02-04
Automatically fixup linked issues in subtree repository

Urgau:
This PR adds a new handler to "relink" linked GitHub issues into their unambiguous form so that when pulling commits from a subtree into the main repository we don't accidentally closes issues in the wrong repository.

Example: `Fixes #123` (in rust-lang/clippy) would be automatically changed to `Fixes rust-lang/clippy#123`

r? @ehuss

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -46,6 +46,7 @@ pub(crate) struct Config {
     pub(crate) merge_conflicts: Option<MergeConflictConfig>,
     pub(crate) bot_pull_requests: Option<BotPullRequests>,
     pub(crate) rendered_link: Option<RenderedLinkConfig>,
+    pub(crate) relink: Option<RelinkConfig>,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -414,6 +415,11 @@ pub(crate) struct RenderedLinkConfig {
     pub(crate) trigger_files: Vec<String>,
 }
 
+#[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub(crate) struct RelinkConfig {}
+
 fn get_cached_config(repo: &str) -> Option<Result<Arc<Config>, ConfigurationError>> {
     let cache = CONFIG_CACHE.read().unwrap();
     cache.get(repo).and_then(|(config, fetch_time)| {
@@ -535,6 +541,8 @@ mod tests {
 
             [shortcut]
 
+            [relink]
+
             [rendered-link]
             trigger-files = ["posts/"]
         "#;
@@ -598,7 +606,8 @@ mod tests {
                 bot_pull_requests: None,
                 rendered_link: Some(RenderedLinkConfig {
                     trigger_files: vec!["posts/".to_string()]
-                })
+                }),
+                relink: Some(RelinkConfig {}),
             }
         );
     }
@@ -662,6 +671,7 @@ mod tests {
                 merge_conflicts: None,
                 bot_pull_requests: None,
                 rendered_link: None,
+                relink: None
             }
         );
     }
--- a/src/config.rs
+++ b/src/config.rs
@@ -46,7 +46,7 @@ pub(crate) struct Config {
     pub(crate) merge_conflicts: Option<MergeConflictConfig>,
     pub(crate) bot_pull_requests: Option<BotPullRequests>,
     pub(crate) rendered_link: Option<RenderedLinkConfig>,
-    pub(crate) relink: Option<RelinkConfig>,
+    pub(crate) canonicalize_issue_links: Option<CanonicalizeIssueLinksConfig>,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -418,7 +418,7 @@ pub(crate) struct RenderedLinkConfig {
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
 #[serde(rename_all = "kebab-case")]
 #[serde(deny_unknown_fields)]
-pub(crate) struct RelinkConfig {}
+pub(crate) struct CanonicalizeIssueLinksConfig {}
 
 fn get_cached_config(repo: &str) -> Option<Result<Arc<Config>, ConfigurationError>> {
     let cache = CONFIG_CACHE.read().unwrap();
@@ -541,7 +541,7 @@ mod tests {
 
             [shortcut]
 
-            [relink]
+            [canonicalize-issue-links]
 
             [rendered-link]
             trigger-files = ["posts/"]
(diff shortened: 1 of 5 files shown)


src/handlers/relink.rs
```diff
@@ -0,0 +1,101 @@
...
+    config::RelinkConfig,
+    github::{Event, IssuesAction, IssuesEvent},
+    handlers::Context,
+};
+
+lazy_static! {
+    // Taken from https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue?quot#linking-a-pull-request-to-an-issue-using-a-keyword
+    static ref LINKED_RE: Regex =
+        Regex::new("(?i)(close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved) +(#[0-9]+)").unwrap();
+}
```
apiraino, line 21: Nit:

since `lazy_static` [has been deprecated](https://github.com/rust-lang-nursery/lazy-static.rs/issues/214), when writing new code I personally prefer using the new [LazyLock](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) API (where appropriate) so that we start slowing phasing out that crate.

wdy?

Urgau: Since we use `lazy_static` I didn't wanted to changed what we use, but since you asked I've changed all the places to use `LazyLock`.

apiraino: Well :sweat_smile:

I didn't exactly ask to replace the crate everywhere, that's would be a change out of scope in this patch. I was just thinking to use it *here* and in your patch (then maybe one day in another patch remove the `lazy_static` crate). I didn't mean to get you do unnecessary work - sorry if I wasn't clear :slightly_smiling_face:

Urgau: > I didn't exactly ask to replace the crate everywhere

I didn't meant to imply that, sorry.

I extracted the removal in https://github.com/rust-lang/triagebot/pull/1898.

jackh726:
I think we should gate this behind a manual command to do the relinking, in case there are issues we don't wreck everyone's PR descriptions.

Urgau:
It's already gated behind the `[relink]` config, if there are any issue (which is very unlikely IMO) the relink can be disabled by removing the section from the configuration file.

Putting the relink behind a manual command renders (IMO) this handler/pass useless, as I don't think anyone is going to use it and the situation is going to stay the same without any improvements.

If we are worried about edge-cases we might miss, we can do a slow deployment, starting with lower traffic repos, and going higher as we have more confidence in it.

jackh726:
Hmm, I suppose it's fine to land this without a gate if we start with a slow repo and allow plenty of testing time

ehuss:
This doesn't seem like it will help if the link is in a commit description, correct? If so, I'm not sure if it will help much if the common case is someone entering it in a commit.

Urgau:
No indeed, if the "fixes" is in the commit it-self this handler won't help. Maybe we should give a warning when we cannot just fix it ourselves.

src/handlers.rs
```diff
@@ -102,6 +103,16 @@ pub async fn handle(ctx: &Context, event: &Event) -> Vec<HandlerError> {
         );
     }
 
+    if let Some(relink_config) = config.as_ref().ok().and_then(|c| c.relink.as_ref()) {
```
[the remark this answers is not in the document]
Kobzol: I was just a bit confused, as you recently created `no_mentions` as an issue handler, but created this functionality as a separate thing :laughing: So I was wondering why not create an issue handler for this too.

src/config.rs
```diff
@@ -48,6 +48,7 @@ pub(crate) struct Config {
     pub(crate) merge_conflicts: Option<MergeConflictConfig>,
     pub(crate) bot_pull_requests: Option<BotPullRequests>,
     pub(crate) rendered_link: Option<RenderedLinkConfig>,
+    pub(crate) relink: Option<RelinkConfig>,
```
Kobzol, line 51: I don't want to bikeshed too much, but this name is quite generic. Maybe something like `relink_issue_links`, `canonicalize_issue_links` or `normalize_issue_links`?

src/handlers.rs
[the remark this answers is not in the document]
Urgau: No particular reason, so I just did it. It's now a "proper" issue handler.

src/config.rs
Urgau: I agree, `relink` is a bit too generic, I think canonicalize is right word so I went with your suggestion.

src/handlers/canonicalize_issue_links.rs
```diff
@@ -0,0 +1,112 @@
...
+});
+
+pub(super) struct CanonicalizeIssueLinksInput {}
+
+pub(super) async fn parse_input(
+    _ctx: &Context,
+    event: &IssuesEvent,
+    config: Option<&CanonicalizeIssueLinksConfig>,
+) -> Result<Option<CanonicalizeIssueLinksInput>, String> {
+    if event.issue.is_pr() {
```
Kobzol, line 31: Shouldn't this condition be reversed?

Urgau: My bad, yes it should have been.

src/handlers/canonicalize_issue_links.rs
```diff
@@ -0,0 +1,112 @@
...
+
+use crate::{
+    config::CanonicalizeIssueLinksConfig,
+    github::{IssuesAction, IssuesEvent},
+    handlers::Context,
+};
+
+// Taken from https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue?quot#linking-a-pull-request-to-an-issue-using-a-keyword
+static LINKED_RE: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new("(?i)(close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved) +(#[0-9]+)")
```
Kobzol, line 20:
```suggestion
    Regex::new("(?i)(close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved)(:? +)(#[0-9]+)")
```
It is common to write something like `Fixes: #1`, so we should count with the colon, I think. Also, it would be nice to capture everything in between and then put it back when replacing the issue number, to avoid making further changes, even if they only affect whitespace.

image

Note: after this change, the replace string should become `let replace_by = format!("${{1}}${{2}}{full_repo_name}${{3}}");`

Kobzol (commented):
I tested this on my test repo and it seems to work fine (modulo one nit).

Since this is gated behind an option, I'm fine with landing it after the nit is fixed. And then we can try it on some "smaller" repos, e.g. Cargo or Rust Analyzer.

src/handlers/canonicalize_issue_links.rs
Urgau: Good call about the colon, I forgot it was accepted, fixed.

As suggested I also re-added the exact spaces.

And to improve the readability I used named-group instead of unnamed ones.

Kobzol (approved):
Looks great! Feel free to merge with/without squashing.

Issue #1865, opened 2024-12-06
Add support for prepared message templates


RalfJung:
There are some messages I end up typing over and over since each contributor needs to learn about our tooling. For example:
```
@rustbot author

Please write `@rustbot ready` when the PR is ready for another round of review.
```
Or
```
This looks good, thanks! Please squash the PR into a single commit.
Use `git rebase --interactive --keep-base ...` to do that so that there's no unnecessary diff.

@rustbot author
```

It would be nice if I could configure some repo-specific help messages so that I don't have to type the same thing over and over again. (I won't always have the time to be as patient as above on explaining the squashing procedure.)

So for instance:
```
@rustbot author-help
@rustbot author-squash
```
could act like `@rustbot author` but also print a message configured in `triagebot.toml`. (I am not sure if it is a good idea to tie this to `@rustbot author`, but both of the most prominent examples in my interaction happen to be of that sort.)

apiraino:
Would the GitHub [Saved Replies](https://docs.github.com/en/get-started/writing-on-github/working-with-saved-replies/using-saved-replies) solve your need? (I use them, too)

RalfJung:
I didn't know about that feature. Will give it a try, thanks!

Urgau:
I'm curious @RalfJung, did you have a chance try to out GitHub Saved Replies?

Does it still make sense to support prepared message in triagebot?

RalfJung:
Yeah I've tried those now and they work great. :)

That said, it could still be useful for triagebot to support something like this so that messages can be shared across maintainers.

PR #1899, opened 2025-02-06
Add reminder to use `ready` shortcut when using `author` shortcut

Urgau:
This PR adds a small reminder to use `@bot ready` when using `@bot author`, similar to what was proposed in in https://github.com/rust-lang/triagebot/issues/1865#issue-2723928123.

This is done to help newcomers, which might not know or have forgot what the command is.

> `@rustbot author`

> Reminder, use `@{bot} ready` to indicate that the PR is ready for review.

~~To avoid spam the bot adds the reminder to the comment using the command instead of creating a new comment.~~

To avoid spam the bot only post the reminder once per PR.

cc @RalfJung
r? @ehuss

### DIFF
--- a/src/handlers/shortcut.rs
+++ b/src/handlers/shortcut.rs
@@ -4,11 +4,22 @@
 
 use crate::{
     config::ShortcutConfig,
+    db::issue_data::IssueData,
     github::{Event, Label},
     handlers::Context,
     interactions::ErrorComment,
 };
 use parser::command::shortcut::ShortcutCommand;
+
+/// Key for the state in the database
+const AUTHOR_REMINDER_KEY: &str = "author-reminder";
+
+/// State stored in the database for a PR.
+#[derive(Debug, Default, serde::Deserialize, serde::Serialize)]
+struct AuthorReminderState {
+    /// ID of the reminder comment.
+    reminder_comment: Option<String>,
+}
 
 pub(super) async fn handle_command(
     ctx: &Context,
@@ -51,6 +62,27 @@ pub(super) async fn handle_command(
                 }],
             )
             .await?;
+    }
+
+    // We add a small reminder for the author to use `@bot ready` when ready
+    if matches!(input, ShortcutCommand::Author) {
+        // Get the state of the author reminder for this PR
+        let mut db = ctx.db.get().await;
+        let mut state: IssueData<'_, AuthorReminderState> =
+            IssueData::load(&mut db, &issue, AUTHOR_REMINDER_KEY).await?;
+
+        if state.data.reminder_comment.is_none() {
+            let comment_body = format!(
+                "Reminder, once the PR becomes ready for a review, use `@{bot} ready`.",
+                bot = &ctx.username,
+            );
+            let comment = issue
+                .post_comment(&ctx.github, comment_body.as_str())
+                .await?;
+
+            state.data.reminder_comment = Some(comment.node_id);
+            state.save().await?;
+        }
     }
 
     Ok(())


RalfJung:
> To avoid spam the bot adds to the comment using the comment instead of creating a new comment.

I assume one of these "comment" should be "command" :D

Urgau:
> Would rather post a new comment _once_ (and then no more).

We do already post [such comment](https://www.github.com/rust-lang/rust/pull/132732#issuecomment-2462696240), but from my experience it doesn't seem to work very much, in particular for new-comers.

I think that is mainly due to due to the time between the comment and the usage of `@rustbot author` by reviewers.

jackh726:
Well...then maybe we should just post the comment every time. (We could e.g. even filter if the user is not in the team repo.)

jackh726:
Also, this doesn't handle the case were the label is manually changed, right? I imagine that happens plenty often.

Urgau:
> Edit: just looked at the link, that's not quite what I was saying, it could still be valuable to have another comment create in response to the author command

I fear that posting a comment instead of editing the comment will quickly be annoying/spam.

I personally wouldn't want the bot to add a comment every-time I (or someone else) uses the `author` command. I would probably find it too distracting and annoying, as it would mean that I would be getting a new notification, that I would then need to triage (mark the email as read or mark it as read on GitHub UI).

@RalfJung, as the issue author, would you happen to have an opinion/preference on the new comment vs edit comment?

> Also, this doesn't handle the case were the label is manually changed, right? I imagine that happens plenty often.

Yes it wouldn't, and that's a conscience choice from me. I think that changing labels and using a command are two distinct way of doing an action and that we shouldn't see them having the same intent.

I use the command with newcomers but often switch the labels manually with regular contributors (as it's faster and they know the process).

RalfJung:
> We do already post [such comment](https://www.github.com/rust-lang/rust/pull/132732#issuecomment-2462696240), but from my experience it doesn't seem to work very much, in particular for new-comers.

We also post that only on their first PR.

> I fear that posting a comment instead of editing the comment will quickly be annoying/spam.

Yeah that's why my original proposal involved the reviewer using a slightly different command to express their intent of posting a hint for the PR author.

How easy would it be to only post a comment the *first* time someone uses `@rustbot author` in a PR?

Urgau:
> How easy would it be to only post a comment the _first_ time someone uses `@rustbot author` in a PR?

Not too difficult, we have the infra to have data per issue/pr inside triagebot.

-------

So if I understand, the consensus seems to be to post the reminder once in a PR in a new comment at the first use of `@rustbot author`, is that right?

I could get behind that, I don't think it will be as effective but it also won't be annoying and spam; and as Jack alluded to earlier we could even omit the reminder when the PR author is in the team database.

ehuss:
A minor concern I have is that I think editing comments could be flaky or dangerous. For example, any handlers that run after this one will not see the edited comment. If they do other strange things like editing, then they will conflict with one another. Another concern is that editing the comment will trigger another webhook, and triagebot will process the comment again. Although triagebot is pretty good about not re-running commands on edit, I don't have 100% confidence in it.

src/handlers/shortcut.rs
```diff
@@ -53,5 +64,26 @@ pub(super) async fn handle_command(
             .await?;
     }
 
+    // We add a small reminder for the author to the `@bot ready` when ready
```
Kobzol, line 67:
```suggestion
    // We add a small reminder for the author to use `@bot ready` when ready
```

src/handlers/shortcut.rs
```diff
@@ -53,5 +64,26 @@ pub(super) async fn handle_command(
...
+    if matches!(input, ShortcutCommand::Author) {
+        // Get the state of the author reminder for this PR
+        let mut db = ctx.db.get().await;
+        let mut state: IssueData<'_, AuthorReminderState> =
+            IssueData::load(&mut db, &issue, AUTHOR_REMINDER_KEY).await?;
+
+        if state.data.reminder_comment.is_none() {
+            let comment_body = format!(
+                "Reminder, use `@{bot} ready` to indicate that the PR is ready for review.",
```
Kobzol, line 76: This makes it sound like they should execute the command right away, IMO. I would change the text to something like "Reminder, once the PR will become ready for a review, use `@{bot} ready`".

Kobzol (commented):
Left two nits, otherwise the implementation looks good to me. Sending the comment once per PR is a good compromise, vs editing a comment, IMO.

src/handlers/shortcut.rs
Urgau: Sure, changed to your suggested text.

Kobzol:
I think that this is a nice improvement and the code looks fine. I will try to test this on a test repo, and unless someone stops me, approve it in a few days.

Kobzol (approved):
Tried it on my [test repo](https://github.com/Kobzol/bors-kindergarten/pull/20), looks good. Thank you!

Kobzol:
https://github.com/rust-lang/rust/pull/112049#issuecomment-2768606830 seems to work :tada:

Issue #1840, opened 2024-09-09
`@rustbot ready` should imply `@rustbot label -S-inactive`


jyn514:
e.g. https://github.com/rust-lang/rust/pull/119286#issuecomment-2338087876

PR #1900, opened 2025-02-09
Remove `S-inactive` label when using shortcuts

Urgau:
This PR removes the `S-inactive` label when any of the shortcuts commands are used.

Fixes https://github.com/rust-lang/triagebot/issues/1840
r? @ehuss


ehuss (approved):
Thanks!

Issue #1831, opened 2024-08-13
Modified submodules are only warned about on PR creation


kadiwa4:
rustbot gives a useful warning when someone modifies a submodule in their PR.

Accidental submodule modifications often happen while rebasing PRs onto a newer base commit ([example 1](https://github.com/rust-lang/rust/pull/123007#discussion_r1559783692), [example 2](https://github.com/rust-lang/rust/pull/128962/commits/70e0f69632591add54bc1e4d625e7bbb9fa02095)). However, rustbot does not usually warn about that because the accidental modification warning is only sent in the greeting message and not if a submodule modification is introduced later on.

PR #1901, opened 2025-02-09
Also show warnings when PRs are updated

Urgau:
This PR tweaks the assign handler to also show warnings when PRs are updated, instead of just doing it when a PR is opened.

Fixes https://github.com/rust-lang/triagebot/issues/1831
r? @ehuss

### DIFF
--- a/src/handlers/check_commits.rs
+++ b/src/handlers/check_commits.rs
@@ -3,18 +3,35 @@ use anyhow::bail;
 use super::Context;
 use crate::{
     config::Config,
-    github::{Event, IssuesAction},
+    db::issue_data::IssueData,
+    github::{Event, IssuesAction, IssuesEvent, ReportedContentClassifiers},
 };
 
 mod modified_submodule;
 mod non_default_branch;
+
+/// Key for the state in the database
+const CHECK_COMMITS_WARNINGS_KEY: &str = "check-commits-warnings";
+
+/// State stored in the database
+#[derive(Debug, Default, serde::Deserialize, serde::Serialize)]
+struct CheckCommitsWarningsState {
+    /// List of the last warnings in the most recent comment.
+    last_warnings: Vec<String>,
+    /// ID of the most recent warning comment.
+    last_warned_comment: Option<String>,
+}
 
 pub(super) async fn handle(ctx: &Context, event: &Event, config: &Config) -> anyhow::Result<()> {
     let Event::Issue(event) = event else {
         return Ok(());
     };
 
-    if !matches!(event.action, IssuesAction::Opened) || !event.issue.is_pr() {
+    if !matches!(
+        event.action,
+        IssuesAction::Opened | IssuesAction::Synchronize
+    ) || !event.issue.is_pr()
+    {
         return Ok(());
     }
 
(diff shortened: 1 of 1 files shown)


Urgau:
That's a good observation, it would indeed post unnecessary comments.

As suggested I have used `IssueData` with some state to keep track of the last warnings to avoid posting unnecessary comments. I haven't tested it, but it's very similar to the other *avoid comment* logic.

Kobzol:
Hmm, it's quite weird in the first place that a handler called `assign` is handling warnings about modified submodules :sweat_smile: Since triagebot doesn't try to send the welcome/assignment and warning messages in a single commit, I think that we should split the logic here.

Keep welcoming and initial assignment in the `assign.rs` handler (which will still only run on Open), and then create a separate handler for checking the commits of a PR (that would run on Open and Synchronize), that would check:
- submodules
- merge commits
- mentions

As I proposed [here](https://github.com/rust-lang/triagebot/pull/1917#issuecomment-2757351488). What do you think?

Kobzol:
I think that we could do a single "check_commits" handler that would have several sub-responsibilities (checking submodules, mentions and merge commits), that could also make it more discoverable to figure out where we check commits (I certainly wouldn't look for submodule changes in the assign handler, lol). It's not so much about the used API (diff vs commit list), but rather about related functionality, which could live at a single place. It would also help with removing some of the duplication of PR status checks (e.g. draft and PR title). It's weird if checking merge commits doesn't trigger when you have a WIP PR title, but no_mentions triggers in that situation, that's quite inconsistent..

On the other hand, that would be a bit of a larger change. We can just keep the status quo, as always.

In terms of performance, rather than doing lazy loading, I would create some "per-event context" that is filled when the various handlers are executed. In other words, when handling a single event and threading it through many handlers, the first time something calls ".commits()", they will be loaded from GH and cached. The second time something calls ".commits()", they would be loaded from memory. If we only keep the cache for a specific GH event and don't persist it further, there should be no invalidation problems (it would even help keep the behavior of the individual handlers more consistent, as in theory the GH state can change in between the execution of two handlers on the same webhook event).

Urgau:
Sure, seems fine to parent handler which would have many sub-responsibility, it would even share the logic for printing the warnings, hidden them, checking the draft status, ... I like that.

In terms of how to proceed, what about?
1. Creation of the `check_commits` handler and moving of "submodules" and "non-default-branch"
2. Rebase this PR to be in the `check_commits` handler (their by introducing hidding and synchronize)
3. Move the "merge commits" checks into `check_commits` handler
4. Add the "mentions" checks within the `check_commits` handler (tbd in https://github.com/rust-lang/triagebot/pull/1917)
5. Add caching/per-event context

Kobzol:
I didn't want to block this functionality on that refactoring, but I'm willing to review it quickly, and help with the implementation if needed, so hopefully it shouldn't take long. I'm also working now on extending the test suite to give us the ability to actually test the issue handlers, but that does not need to block this refactor.

That plan sounds good!

Urgau:
Rebased on top of https://github.com/rust-lang/triagebot/pull/1922, the logic is now in the `check_commits` handler.

Kobzol (approved):
For `warn_non_default_branch`, we should also be checking the PR edited event, but I don't think that's super important, as it should be very rare to "repoint" a PR to a different base branch after it has been created. The important event is `Synchronize` for the submodules, because that happens quite often, so it would be nice to warn about it.

PR #1876, opened 2025-01-05
Revert "Lock Major Change Proposal issue"

Urgau:
This PR reverts commit https://github.com/rust-lang/triagebot/commit/d5649f42835bad5e9438a584dddda0a570f6e75f from https://github.com/rust-lang/triagebot/pull/1858 as [it was found](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Lock.20discussions.20on.20MCPs/near/491957647) that if the PR is not member of the org he cannot edit the MCP description.

I think it's more important to let author edit the MCP description than preventing people from potential posting comment on the issue.

r? @ehuss

PR #1902, opened 2025-02-11
Emphasis Zulip thread more in MCP opening comment

Urgau:
Instead of locking the MCP GitHub issue, like was done in #1876 (until a subsequent revert), I think we should put more emphasis on the Zulip thread in MCP opening comment.

Currently it's kind of hidden between a wall of text, I propose that we use [GitHub blockquote annotations](https://github.com/orgs/community/discussions/16925#discussion-4085374) to increase the emphasis on it. I also included some minor stylistic changes to help break the wall of text.

I choose "Important" as I feel like it's the only one that fits our goal.

cc @apiraino
r? @ehuss

### DIFF
--- a/src/handlers/major_change.rs
+++ b/src/handlers/major_change.rs
@@ -259,8 +259,9 @@ async fn handle(
     if new_proposal {
         let topic_url = zulip_req.url();
         let comment = format!(
-            "This issue is not meant to be used for technical discussion. \
-        There is a Zulip [stream] for that. Use this issue to leave \
+            "> [!IMPORTANT] \n\
+        > This issue is *not meant to be used for technical discussion*. \
+        There is a **Zulip [stream]** for that. Use this issue to leave \
         procedural comments, such as volunteering to review, indicating that you \
         second the proposal (or third, etc), or raising a concern that you would \
         like to be addressed. \


Urgau:
> [!IMPORTANT]
> This issue is *not meant to be used for technical discussion*. There is a **Zulip [stream]** for that. Use this issue to leave procedural comments, such as volunteering to review, indicating that you second the proposal (or third, etc), or raising a concern that you would like to be addressed. 

Concerns or objections to the proposal should be discussed on Zulip and formally registered here by adding a comment.

 ``` 
 @rfcbot concern reason-for-concern 
 <description of the concern> 
 ``` 
 Concerns can be lifted with: 
 ``` 
 @rfcbot resolve reason-for-concern 
 ```

See documentation at [https://forge.rust-lang.org](https://forge.rust-lang.org/compiler/mcp.html#what-kinds-of-comments-should-go-on-the-tracking-issue-in-compiler-team-repo) 

cc `@rust-lang/compiler`

[stream]: https://rust-lang.zulipchat.com/#narrow/stream/233931-xxx/topic/Give.20integer.20literals.20a.20sign.20instead.20of.20r.E2.80.A6.20compiler-team.23835

ehuss:
@apiraino Are you OK with this?

apiraino:
¯\\_(ツ)_/¯

My opinion is that people *not* reading comments will keep *not* reading comments. Any sort of emphasis we add after a while is not noticed anymore. Maybe these comments come because some people reply to a notification? Who knows.

In any case, I don't think this is not such a pressing or extremely annoying problem worth our time.

Let's merge and move on.

ehuss (approved):
Thanks!

apiraino:
@Urgau the MCP template is now a little bit hostile :sweat_smile: the same warning is now in CAPS and in blockquote.
https://togithub.com/rust-lang/compiler-team/issues/836#issuecomment-2659236468

Do you think we can remove the warning from the MCP template and just use yours (whichI think is more visible)?

apiraino:
alternative we can add also a

`<blink>This issue is not meant to be used for technical discussion</blink>`

:D

Urgau:
> Do you think we can remove the warning from the MCP template and just use yours (whichI think is more visible)?

Yeah, sure, good idea.

It will even clear some vertical space.

Done in https://github.com/rust-lang/compiler-team/pull/837

PR #1905, opened 2025-02-17
Handle open/close PR events in PR tracking

Kobzol:
This should improve the robustness of PR tracking, as before it was not taking (re)open/close/merge PR events into account. This PR also extends the test infrastructure. We still don't test the input/output GitHub API calls, but even just testing the DB is better than nothing.

r? @apiraino :)

### DIFF
--- a/src/handlers/pr_tracking.rs
+++ b/src/handlers/pr_tracking.rs
@@ -238,7 +238,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn remove_pr_from_workqueue_on_assign() {
+    async fn remove_pr_from_workqueue_on_unassign() {
         run_test(|ctx| async move {
             let user = user("Martin", 2);
             set_assigned_prs(&ctx, &user, &[10]).await;
@@ -305,6 +305,7 @@ mod tests {
         .await;
     }
 
+    // Make sure that we only consider pull requests, not issues.
     #[tokio::test]
     async fn ignore_issue_assignments() {
         run_test(|ctx| async move {
--- a/src/handlers/pr_tracking.rs
+++ b/src/handlers/pr_tracking.rs
@@ -6,6 +6,8 @@
 //! - Adds the PR to the workqueue of one team member (after the PR has been assigned)
 //! - Removes the PR from the workqueue of one team member (after the PR has been unassigned or closed)
 
+use super::assign::{FindReviewerError, REVIEWER_HAS_NO_CAPACITY, SELF_ASSIGN_HAS_NO_CAPACITY};
+use crate::github::User;
 use crate::{
     config::ReviewPrefsConfig,
     db::notifications::record_username,
@@ -17,9 +19,12 @@ use anyhow::Context as _;
 use tokio_postgres::Client as DbClient;
 use tracing as log;
 
-use super::assign::{FindReviewerError, REVIEWER_HAS_NO_CAPACITY, SELF_ASSIGN_HAS_NO_CAPACITY};
-
-pub(super) struct ReviewPrefsInput {}
+pub(super) enum ReviewPrefsInput {
+    Assigned { assignee: User },
+    Unassigned { assignee: User },
+    Reopened,
+    Closed,
+}
 
 pub(super) async fn parse_input(
     _ctx: &Context,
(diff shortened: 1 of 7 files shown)


src/handlers/pr_tracking.rs
```diff
@@ -164,3 +206,187 @@ WHERE r.user_id = $1;";
...
+
+            check_assigned_prs(&ctx, &user, &[10]).await;
+
+            Ok(ctx)
+        })
+        .await;
+    }
+
+    #[tokio::test]
+    async fn remove_pr_from_workqueue_on_assign() {
```
jackh726, line 241: `on_unassign`?

src/handlers/pr_tracking.rs
```diff
@@ -164,3 +206,187 @@ WHERE r.user_id = $1;";
...
+
+            check_assigned_prs(&ctx, &user, &[10, 42]).await;
+
+            Ok(ctx)
+        })
+        .await;
+    }
+
+    #[tokio::test]
+    async fn ignore_issue_assignments() {
```
jackh726, line 309: This test confuses me...what's its purpose and how is it different than add_pr_to_workqueue_on_assign?

Kobzol: I spent like 1 hour "fixing" PR tracking, because I thought that it also considers issues, not only PRs. Only after that I noticed `parse_input` :facepalm: So I thought that I would add an explicit test that PR tracking only considers PRs and not *issues*, because in code they are represented in an almost identical way.

Kobzol: Added a comment.

src/handlers/pr_tracking.rs
Kobzol: Thanks! Fixed.

src/handlers/pr_tracking.rs
jackh726: Oh I see, `issue()` instead of `pull_request()`, I didn't even notice.

src/handlers/pr_tracking.rs
```diff
@@ -137,15 +174,20 @@ AND CARDINALITY(r.assigned_prs) < LEAST(COALESCE(r.max_assigned_prs,1000000));";
...
 async fn upsert_pr_into_workqueue(
     db: &DbClient,
-    user_id: u64,
+    user: &User,
     pr: u64,
 ) -> anyhow::Result<u64, anyhow::Error> {
+    // Ensure the user has entry in the `users` table
+    record_username(db, user.id, &user.login)
+        .await
+        .context("failed to record username")?;
```
apiraino, line 183: Here there is now a "hidden" a second SQL query where I expect to only have one. `upsert_pr_into_workqueue` says to just "Add a PR to the workqueue of a team member", we are sort of breaking this promise.

Maybe call `record_username()` just before (at the cost of calling everywhere)?

src/handlers/pr_tracking.rs
```diff
@@ -164,3 +164,145 @@ WHERE r.user_id = $1;";
...
+    async fn set_assigned_prs(ctx: &TestContext, user: &User, prs: &[i32]) {
+        let client = ctx.db_client().await;
+        let client = client.client();
+
+        record_username(client, user.id, &user.login).await.unwrap();
+        for &pr in prs {
+            upsert_pr_into_workqueue(&ctx.db_client().await.client(), user.id, pr as u64)
+                .await
+                .unwrap();
+        }
```
apiraino: given my other comment, here you would end with multiple redundant SQL INSERT being run.

(ok this is just a test, but still)

apiraino (changes requested):
I checked the code part and skimmed through the tests.

Besides a nit , I think it looks good.

thanks @Kobzol !

src/handlers/pr_tracking.rs
Kobzol: I think that it's better if the function works in a self-contained way, rather than not working if you don't call a specific other function before it. We don't call it in a loop outside tests, it's always called incrementally, and the code that calls it always called `record_username` before it anyways, so there wouldn't be any change.

But mainly, I want to nuke this whole code in a follow-up PR, so.. :laughing:

PR #1906, opened 2025-02-18
Store assigned PRs in memory

Kobzol:
This PR stops using `assigned_prs` column in the `review_prefs` table, and remembers PR assignments in-memory instead.


src/db.rs
```diff
@@ -344,4 +344,5 @@ CREATE UNIQUE INDEX IF NOT EXISTS review_prefs_user_id ON review_prefs(user_id);
     "
 ALTER TABLE review_prefs ADD COLUMN IF NOT EXISTS max_assigned_prs INTEGER DEFAULT NULL;
 ",
+    "ALTER TABLE review_prefs DROP COLUMN assigned_prs;",
```
Kobzol, line 347: We can also just skip this migration and keep the field in the DB, but I don't think that it's needed anymore.

Kobzol:
Rebased.

src/handlers/assign.rs
```diff
@@ -841,24 +840,24 @@ async fn find_reviewer_from_names(
...
-        //     filtered: names.to_vec(),
-        // });
-    }
-
-    log::info!(
-        "[#{}] Filtered list of candidates: {:?}",
-        issue.number,
-        filtered_candidates
-    );
+    // let filtered_candidates = filter_by_capacity(db, &candidates)
```
jackh726, line 843: Similar here.

src/handlers/pull_requests_assignment_update.rs
```diff
@@ -17,70 +13,23 @@ impl Job for PullRequestAssignmentUpdate {
...
-        // delete all PR assignments before populating
-        init_table(&db).await?;
-
-        // aggregate by user first
-        let aggregated = prs.into_iter().fold(HashMap::new(), |mut acc, (user, pr)| {
-            let (_, prs) = acc.entry(user.id).or_insert_with(|| (user, Vec::new()));
-            prs.push(pr);
-            acc
-        });
+        let rust_repo = gh.repository("kobzol/bors-kindergarten").await?;
```
jackh726, line 20: This definitely isn't right...

jackh726 (changes requested):

src/db.rs
Kobzol: Ok, fair enough, I'll keep the column.

src/handlers.rs
```diff
@@ -348,4 +349,7 @@ pub struct Context {
     pub db: crate::db::ClientPool,
     pub username: String,
     pub octocrab: Octocrab,
+    /// Represents the workqueue (assigned open PRs) of individual reviewers.
+    /// tokio's RwLock is used to avoid deadlocks, since we run on a single-threaded tokio runtime.
+    pub workqueue: Arc<tokio::sync::RwLock<ReviewerWorkqueue>>,
```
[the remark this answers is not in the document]
Kobzol: The problem is that this state has to be available across handlers. It is populated in `pull_requests_assignment_update.rs`, updated in `pr_tracking.rs` and it will be read in the future in `assign.rs`. Thus it's prett ymuch global state, hence its storage in `Context`.

src/handlers/pull_requests_assignment_update.rs
Kobzol: Oops :) I accidentally committed this during the rebase, good catch!

src/handlers/assign.rs
```diff
@@ -560,22 +559,22 @@ pub(super) async fn handle_command(
...
-                        );
-                        // issue
-                        //     .post_comment(
-                        //         &ctx.github,
-                        //         &REVIEWER_HAS_NO_CAPACITY.replace("{username}", &name),
-                        //     )
-                        //     .await?;
-                        // return Ok(());
-                    }
+                    // let work_queue = has_user_capacity(&db_client, &name).await;
```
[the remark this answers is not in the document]
Kobzol: These functions that I commented were using the `assigned_prs` column in the DB, which is no longer used after this PR. So these functions simply do not work anymore (in fact I removed them in this PR).

I'm planning to further work on assigning PR capacity limit, and then refactor the way PRs are assigned, and add back logging (and then we can hopefully enable the new assignment logic again).

I did not want to do all that in this PR, hence I commented this code.

Kobzol:
Let me know if you want further changes :)

jackh726 (approved):

jackh726:
@Kobzol I'm holding off merging, since I think we should make sure to merge when we can quickly revert if there are problems.

PR #1907, opened 2025-02-19
Fix overflowing integer when retrieving comments for FCP

apiraino:
This is a patch I've been using locally since long, should be the final nail in the coffin to fix the overflowing GH ID issue we started experiencing a while ago.

I'd like to upstream it after some testing.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -648,7 +648,7 @@ impl Issue {
         self.state == IssueState::Open
     }
 
-    pub async fn get_comment(&self, client: &GithubClient, id: i32) -> anyhow::Result<Comment> {
+    pub async fn get_comment(&self, client: &GithubClient, id: u64) -> anyhow::Result<Comment> {
         let comment_url = format!("{}/issues/comments/{}", self.repository().url(client), id);
         let comment = client.json(client.get(&comment_url)).await?;
         Ok(comment)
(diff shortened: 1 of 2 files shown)


Mark-Simulacrum (approved):
Might want to switch to u64 to preempt another round of problems.

apiraino:
Note for myself to investigate before merge: using u64 causes a runtime error when retrieving issues

```
2025-02-26T16:29:55.830555Z DEBUG triagebot::github: send_req with RequestBuilder { method: GET, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("api.github.com")), port: None, path: "/search/issues", query: Some("q=state:open+no:assignee+label:T-compiler+label:P-critical+repo:rust-lang/rust&sort=created&order=asc&per_page=100&page=1"), fragment: None }, headers: {"user-agent": "rust-lang-triagebot", "authorization": Sensitive} }
...
Error: out of range integral type conversion attempted
```

The error happens [here](https://github.com/rust-lang/triagebot/blob/d17ef9ca40d0578830db112882baffa20f560994/src/github.rs#L1346-L1350)

Urgau:
@apiraino I don't know how changes to `rfcbot` structs could be related to a failure de-serializing `IssueSearchResult` in GitHub, are you sure they are related, may have been spurious failures.

Kobzol (approved):
I'm not sure how does it help if RFCbot still serializes these fields as `i32` (even if it converts them to `u32` right before writing them, but that doesn't help if the value is over 2.something billion already on the rfcbot side), but it also hopefully shouldn't hurt, so let's see :)

Seems like in https://rfcbot.rs/api/all, some `fk_initiating_comment` values are e.g. `2669499833`, so clearly they go over `i32`. Not sure how that's even possible, probably the values are actually represented as negative numbers in rfcbot or something? xD Anyway.

PR #1908, opened 2025-02-20
Extend user loading from the DB

Kobzol:
I split these changes from a larger PR that I'm working on that adds the option to set max review PR capacity.

It's a bit fishy that we user the GitHub `User` type, which is returned from webhooks, to also represent the user type in the DB, but so far they are pretty much the same, so unless we need to separate them, I'd keep it that way to keep the code simpler.

### DIFF
--- a/src/db/users.rs
+++ b/src/db/users.rs
@@ -1,7 +1,9 @@
+use crate::github::User;
 use anyhow::Context;
 use tokio_postgres::Client as DbClient;
 
-/// Add a new user (if not existing)
+/// Add a new user.
+/// If an user already exists, updates their username.
 pub async fn record_username(db: &DbClient, user_id: u64, username: &str) -> anyhow::Result<()> {
     db.execute(
         r"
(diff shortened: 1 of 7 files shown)


src/db/users.rs
```diff
@@ -0,0 +1,60 @@
...
+        )
+        .await
+        .context("cannot load user from DB")?;
+    Ok(row.map(|row| {
+        let username: &str = row.get(0);
+        User {
+            id: user_id,
+            login: username.to_string(),
+        }
+    }))
```
apiraino, line 38: I think if you `SELECT *` then you can just `Ok(row.into())` but it's a matter of preference ...

apiraino (approved):

src/db/users.rs
Kobzol: We don't currently implement `From<tokio_postgres::Row> for User` anywhere :( Which goes back to my comment that we kind of misuse the GitHub user to also mean the database user.

Introducing a separate type would mean doing some conversions, but I guess that if we want to keep the `users` table, it's the Right Thing™ to do, in the long term. What do you think? :)

PR #1910, opened 2025-02-24
Temporarily disable new contributor welcomes

ehuss:
This query is taking too long, and is causing the webhook timeouts which prevents all work from completing.


apiraino:
I'm fine with experimenting with this to get concrete data about how long these GraphQL queries are taking. To establish if they are the (or part of the) problem.

This said, the welcome message *per se* is imo not essential but it served the purpose to set expectations about the review. That I will miss a little bit.

Mark-Simulacrum (approved):
In favor of avoiding problems while we discuss further.

Issue #1881, opened 2025-01-14
rustbot assigned two reviewers


RalfJung:
In https://github.com/rust-lang/rust/pull/135423 I wrote "r? compiler" in a [review summary comment](https://github.com/rust-lang/rust/pull/135423#pullrequestreview-2549051490) and now the PR has two reviewers assigned...

RalfJung:
> and then:

I wrote `r? compiler` again to resolve the reviewer tie, so that is expected.

Seems like me submitting the review somehow triggered two events, and furthermore there's a race condition in triagebot where if two events occur at the same time, we can end up with two reviewers.

Nadrieril:
Same thing happened on https://github.com/rust-lang/rust/pull/135102 a few days ago 

![Image](https://github.com/user-attachments/assets/2963d531-4621-40cf-8fe0-c53173385cb0)

apiraino:
> there's a race condition in triagebot where if two events occur at the same time, we can end up with two reviewers.

Just to be clear, multiple reviewers are supported by the GH API so the triagebot reflects that (it's just that we don't use multiple reviewers, as far as I know).

Why sometimes we receive two events when someone requests an assignment is the interesting question 🤔

RalfJung:
> Just to be clear, multiple reviewers are supported by the GH API so the triagebot reflects that (it's just that we don't use multiple reviewers, as far as I know).

Yeah I know, but triagebot is supposed to maintain the general property of there being only one reviewer for our PRs.

apiraino:
Another case (from [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/New.20PR.20assignment!/near/501734193))

https://github.com/rust-lang/rust/pull/137549#event-16442291855

ehuss:
For https://github.com/rust-lang/rust/pull/137549#event-16442291855, the logs show that GitHub sent two events in quick succession:

2025-02-25T01:05:15.525531Z -- pull_request_review edited, changes=Some(Changes { title: None, body: None }) (body=None makes no sense)
2025-02-25T01:05:15.743609Z -- pull_request_review submitted (happened after "edited"!?)

I'm curious what the raw JSON of the first event looked like, since it doesn't make sense to send an "edited" event with no changes.

[...]

PR #1911, opened 2025-02-26
Ignore comment edited with no changes

ehuss:
This changes it so that when GitHub sends an issue comment "edited" event with no changes, triagebot will ignore it.

For unknown reasons, sometimes when someone posts a PR review, GitHub sends an "edited" event with `"changes":{}`, immediately followed by a "created" event with the actual content, which doesn't make much sense. I'm pretty sure that any "edited" event without changes can be safely ignored. For example, under normal circumstances GitHub will correctly report empty bodies with the empty string (not null), so those should still continue to be handled correctly.

Fixes https://github.com/rust-lang/triagebot/issues/1881

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -246,10 +246,25 @@ macro_rules! command_handlers {
                     log::debug!("skipping event, issue was {:?}", e.action);
                     return;
                 }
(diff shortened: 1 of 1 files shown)


apiraino (approved):
I'm down to experiment with this

PR #1916, opened 2025-03-07
Allow anyone to relabel `CI-spurious-*`

thaliaarchi:
As suggested by @Noratrieb in reference to me attempting to set this label in https://github.com/rust-lang/rust/pull/136780#issuecomment-2707364425, allow unauthenticated users to relabel `CI-spurious-*`.


thaliaarchi:
This should have been filed against rust-lang/rust's triagebot.toml.

PR #1917, opened 2025-03-07
Add no GitHub mentions in commits handler

Urgau:
This PR implements a no-mentions in commits handler. This is so that people are not spammed by GitHub when someone pushes a commit with their GitHub handle in it.

This is coming from https://github.com/rust-lang/rust/issues/137990. I should note that I also have been annoyed by that in the past.

The implementation is heavily inspired by the `[no-merges]` one.

The warning is gated behind the `[no-mentions]` config and produces such warning:

> There are mentions (`@mention`) in your commits. We have a no mention policy so these commits will need to be removed for this pull request to be merged.
>
> The following commits have mentions is them:
>  - commit1
>  - commit2

r? @ehuss

### DIFF
--- a/src/handlers/check_commits/no_mentions.rs
+++ b/src/handlers/check_commits/no_mentions.rs
@@ -35,22 +35,44 @@ fn mentions_in_commits_warn(commits: Vec<&str>) -> String {
 }
 
 #[test]
-fn test_warning_printing() {
-    let commits_to_warn = vec![
-        "4d6ze57403udfrzefrfe6574",
-        "f54efz57405u46z6ef465z4f6ze57",
-        "404u57403uzf5fe5f4f5e57405u4zf",
-    ];
+fn test_mentions_in_commits() {
+    fn dummy_commit_from_body(sha: &str, body: &str) -> GithubCommit {
+        use chrono::{DateTime, FixedOffset};
 
-    let msg = mentions_in_commits_warn(commits_to_warn);
+        GithubCommit {
+            sha: sha.to_string(),
+            commit: crate::github::GithubCommitCommitField {
+                author: crate::github::GitUser {
+                    date: DateTime::<FixedOffset>::MIN_UTC.into(),
+                },
+                message: body.to_string(),
+                tree: crate::github::GitCommitTree {
+                    sha: "60ff73dfdd81aa1e6737eb3dacdfd4a141f6e14d".to_string(),
+                },
+            },
+            parents: vec![],
+        }
+    }
+
+    let mut commits = vec![dummy_commit_from_body(
+        "d1992a392617dfb10518c3e56446b6c9efae38b0",
+        "This is simple without mentions!",
+    )];
+
+    assert_eq!(mentions_in_commits(&NoMentionsConfig {}, &commits), None);
+
+    commits.push(dummy_commit_from_body(
+        "d7daa17bc97df9377640b0d33cbd0bbeed703c3a",
+        "This is a body with a @mention!",
+    ));
 
     assert_eq!(
-        msg,
-        r#"There are username mentions (such as `@user`) in the commit messages of the following commits.
+        mentions_in_commits(&NoMentionsConfig {}, &commits),
+        Some(
+            r#"There are username mentions (such as `@user`) in the commit messages of the following commits.
   *Please remove the mentions to avoid spamming these users.*
-    - 4d6ze57403udfrzefrfe6574
-    - f54efz57405u46z6ef465z4f6ze57
-    - 404u57403uzf5fe5f4f5e57405u4zf
-"#
+    - d7daa17bc97df9377640b0d33cbd0bbeed703c3a
+"#.to_string()
+        )
     );
 }
--- a/src/handlers/check_commits/no_mentions.rs
+++ b/src/handlers/check_commits/no_mentions.rs
@@ -1,0 +1,56 @@
+//! Purpose: When opening a PR, or pushing new changes, check for github mentions
+//! in commits and notify the user of our no-mentions in commits policy.
+
+use std::fmt::Write;
+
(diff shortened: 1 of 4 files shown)


ehuss:
If this is intended to go in rust-lang/rust, then I think this would need approval from the primary teams that would be affected there (mainly compiler and libs I'm thinking), but I imagine would transitively also be required for all subtree teams, or otherwise it seems like it would be quite noisy. It's also not clear to me that "need to be removed" is not true, as I don't see anything that would prevent it.

I'd also say that this doesn't seem to cover the PR description which I think is the more common culprit. However, that will conflict with `r?` commands or general comments people leave to copy other people.

I have a slight worry that this will be too overbearing.

Urgau:
Sure, this would need buying from subtree first.

> I'd also say that this doesn't seem to cover the PR description which I think is the more common culprit. However, that will conflict with r? commands or general comments people leave to copy other people.

I think this is in part already handled by mentions being escaped in homu: [rust-lang/homu#100](https://github.com/rust-lang/homu/pull/100)

Non-homu handled repositories would still have this issue, but I think PR description are not as bad as commit messages. I have rarely see people link merge commit (where the PR description end up), while it's much more common for individual commits IMO to be copied. 

> It's also not clear to me that "need to be removed" is not true, as I don't see anything that would prevent it.

It's the same generic message as the one regarding merge commits not being allowed, which is also not enforced, particularly for subtrees.

> I have a slight worry that this will be too overbearing.

As proposed it's limited it to commit message, which hopefully will make it near invisible to most people.

Kobzol:
The implementation looks good to me. It omits some functionality of `no_merges` though, notably ignoring drafts and `exclude_titles`. I'm also a bit worried about duplicating the commit loading between these two handlers. After a push to a PR, two handlers will not attempt to load all commits of a PR, which seems a bit wasteful. But since they don't run concurrently, it's probably fine? I wonder if we should perhaps merge these two handlers, into something like "check_commits" handler. They share a lot of code anyway, and in this way we could scan the commits just once and then check merge commits and mentions together.

Same as with `no_merges`, I think that it's fine that this warns, but doesn't otherwise prevent the PR from being merged.
I'm not sure if the subtrees are such a problem? On sync PRs, this might send a notice if the subtree uses a mention in its commits, but that doesn't seem like such an issue.

src/handlers/no_mentions.rs
```diff
@@ -0,0 +1,158 @@
...
+const NO_MENTIONS_KEY: &str = "no_mentions";
+
+pub(super) struct NoMentionsInput {
+    /// Hashes of commits that have mentions in the pull request.
+    mentions_commits: HashSet<String>,
+}
+
+#[derive(Debug, Default, Deserialize, Serialize)]
+struct NoMentionsState {
+    /// Hashes of mention commits that have already been mentioned by triagebot in a comment.
```
Kobzol: Too many `mention`s :joy:

src/handlers/check_commits/no_mentions.rs
```diff
@@ -0,0 +1,56 @@
...
+    if mentions_commits.is_empty() {
+        None
+    } else {
+        Some(mentions_in_commits_warn(mentions_commits))
+    }
+}
+
+fn mentions_in_commits_warn(commits: Vec<&str>) -> String {
+    let mut warning =
+        format!("There are mentions (`@user`) in the following commits, please remove them.\n");
```
Kobzol, line 29:
```suggestion
        format!("There are username mentions (such as `@user`) in the commit messages of the following commits. Please remove the mentions to avoid spamming these users.\n");
```

src/handlers/check_commits/no_mentions.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    for commit in commits {
+        let _ = writeln!(warning, "    - {commit}");
+    }
+
+    warning
+}
+
+#[test]
+fn test_warning_printing() {
```
Kobzol, line 39: It shouldn't be that hard to test the actual functionality more end-to-end. Could you please test `mentions_in_commits` instead of just `mentions_in_commits`? Should be enough to construct fake `GitHubCommit`s, we can add helper functions for doing that into `src/tests`, it will surely be useful in other places.

I want to add tests to new functionality or modifications if possible, and here I think it's a low-hanging fruit.

Kobzol (commented):
Looks good, but left a few comments.

When multiple warnings are included together, it looks a bit [confusing](https://github.com/Kobzol/bors-kindergarten/pull/26#issuecomment-2785792618), because the submodules handler doesn't actually list the actual commits with submodules. Maybe we could reword it from "These commits modify submodules" to "Some commits modify submodules".

src/handlers/check_commits/no_mentions.rs
Urgau: I was trying to avoid a too long line compared to the other warnings, but going to the line between the two sentences makes it okay. Changed.

src/handlers/check_commits/no_mentions.rs
Urgau: Sure, added an end-to-end test, not sure about the quality of it, but works for our use-case.

Kobzol: Thanks, much better than nothing!

Kobzol:
Looks good! However, while testing, I found out that the detection of "same warnings" in `check_commits` is broken :sweat_smile:
[Here](https://github.com/rust-lang/triagebot/blob/596f0c5f60cd0b87175994393325b14f9615334e/src/handlers/check_commits.rs#L101) we overwrite the `warnings` variable by adding an asterisk to them, so the bot always things that the warnings are different, even though they are in fact the same.

Could you fix that in this PR please? Since it touches these files anyway.

Urgau:
That's embarrassing. Fixed as a separate commit.

PR #1918, opened 2025-03-10
Make pull request assignment loading more robust

Kobzol:
This PR reimplements the loading of open pull requests and their assignments to GitHub users. In particular, it changes the implementation from a GraphQL query to REST API calls, which is 5-10x faster based on my measurements, and should avoid the timeouts that I saw in production logs. It also performs this fetch periodically, every 30 minutes, so that we can periodically synchronize the latest state and thus paper over any potentially missed webhooks.

The initial state is also loaded eagerly when triagebot starts, so that we always have a consistent state of the world. If triagebot started without it, it could (in the future, once the workqueue is taken into account) for some time assign PRs based on the assumption that no reviewer has any assigned PRs, which is not great.

The last two commits remove the GQL query, which is now unused, and also the one-off job mechanism, since the only job that was using it is now a cron job instead.


src/handlers/pull_requests_assignment_update.rs
```diff
@@ -13,23 +15,27 @@ impl Job for PullRequestAssignmentUpdate {
...
+/// Loads the workqueue (mapping of open PRs assigned to users) from GitHub
+pub async fn load_workqueue(client: &Octocrab) -> anyhow::Result<ReviewerWorkqueue> {
+    tracing::debug!("Loading workqueue for rust-lang/rust");
+    let prs = retrieve_pull_request_assignments("rust-lang", "rust", &client).await?;
+
+    // Aggregate PRs by user
+    let aggregated: HashMap<UserId, HashSet<PullRequestNumber>> =
+        prs.into_iter().fold(HashMap::new(), |mut acc, (user, pr)| {
+            let prs = acc.entry(user.id).or_default();
+            prs.insert(pr as PullRequestNumber);
```
fbstj, line 36:
```suggestion
            prs.insert(pr);
```
isn't this like this from the return type of `retrieve_pull_request_assignments`?

src/main.rs
```diff
@@ -253,6 +256,23 @@ async fn run_server(addr: SocketAddr) -> anyhow::Result<()> {
         .personal_token(github::default_token_from_env())
         .build()
         .expect("Failed to build octograb.");
+
+    // Load the initial workqueue state from GitHub
+    // In case this fails, we do not want to block triagebot, instead
+    // we use an empty workqueue and let it be updated later through
+    // webhooks and the `PullRequestAssignmentUpdate` cron job.
+    let workqueue = match tokio::time::timeout(Duration::from_secs(60), load_workqueue(&gh)).await {
```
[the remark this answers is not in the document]
Kobzol: The loading should take ~15 seconds, 60s is just an upper bound. We will webhooks in that period, but that already happens when triagebot is being (re)started.

The reason why we eagerly load this data is that we need the workqueue to be up-to-date to provide the correct assignments (in the near future, once we actually start using the workqueue). If the workqueue started empty, then until it gets loaded, all PRs opened in the meantime would be assigned as if no reviewer had any PR assigned, which is not ideal.

After this PR, the PR assignment loading should be much faster than before, so I think that waiting ~15s should be fine.

src/handlers/pull_requests_assignment_update.rs
Kobzol: Yes, before that function returned `u64`, but now it's redundant. Thanks!

src/main.rs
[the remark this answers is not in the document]
Urgau: I'm still a bit unease about loosing ~15 seconds of events while we can avoid it. I don't how much does triagebot needs to start-up and how much downtime we currently have when deploying.

But it's only 15 seconds and consistency is a pretty good argument, so I won't block on it. (We can always change it later, anyway).

src/github.rs
```diff
@@ -3049,86 +3053,55 @@ async fn project_items_by_status(
...
+    let stream = client
+        .pulls(owner, repository)
+        .list()
+        .state(State::Open)
+        .direction(Direction::Ascending)
+        .sort(Sort::Created)
+        .per_page(100)
+        .send()
+        .await?
+        .into_stream(client);
```
Urgau, line 3077: :heart: much simpler!

Cargo.toml
```diff
@@ -33,7 +33,7 @@ postgres-native-tls = "0.5.0"
 native-tls = "0.2"
 x509-cert = { version = "0.2.5", features = ["pem"] }
 serde_path_to_error = "0.1.2"
-octocrab = "0.30.1"
+octocrab = { version = "0.30.1", features = ["stream"] }
```
Urgau, line 36: We should about updating the `octocrab` sometime in the future, our version seems quite old.

src/github.rs
```diff
@@ -3049,86 +3053,55 @@ async fn project_items_by_status(
...
+            .iter()
+            .any(|label| label.name == "rollup")
+        {
+            continue;
+        }
+        for user in pr.assignees.unwrap_or_default() {
+            assignments.push((
+                User {
+                    login: user.login,
+                    id: (*user.id).into(),
```
Urgau, line 3096: This is changing `database_id` (only on GraphQl?) to `id (only on REST?). Are we sure this won't cause issue with existing GitHub calls?

src/main.rs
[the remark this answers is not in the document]
Kobzol: An alternative could be to start the update job in the background, and then if a webhook requiring this data comes before the data is loaded, `await` until the workqueue is loaded. But that would bring considerable implementation complexity.

I don't think that triagebot restarts so often for this to be a big issue in practice, we'll see.

src/github.rs
Kobzol: GitHub IDs are a bit confusing. My understanding is that GQL's `database_id` corresponds to the `id` used in most REST API calls.

Urgau (approved):
The code looks good to me.

If you have confirmed that it works in practice, let's merge it.

Kobzol:
Well, it's hard to really confirm something with triagebot :laughing: But it seems to be loading fine for me locally. I added a small log to give us information when the loading starts and ends.

PR #1920, opened 2025-03-31
Allow assigning to users on vacation

Noratrieb:
People on vacation will usually not (they better >:() respond during their vacation, but they probably will be able to respond after. There is no reason why you wouldn't be able to assign to someone on vacation manually if this person is the best reviewer and you have enough patience for them to return, so don't prevent it. Do print a warning though to set expectations for both the author and triage.

If someone wants to be completely unassignable because they have left, they'll already be unassignable because they have left.

docs: https://github.com/rust-lang/rust-forge/pull/821

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -978,9 +978,7 @@ fn candidate_reviewers_from_names<'a>(
         }
 
         // Assume it is a user.
-        if filter(&group_or_user) {
-            candidates.insert(group_or_user);
-        }
+        candidates.insert(group_or_user);
     }
     if candidates.is_empty() {
         let initial = names.iter().cloned().collect();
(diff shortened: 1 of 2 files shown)


src/handlers/assign.rs
```diff
@@ -978,9 +1008,7 @@ fn candidate_reviewers_from_names<'a>(
         }
 
         // Assume it is a user.
-        if filter(&group_or_user) {
-            candidates.insert(group_or_user);
-        }
+        candidates.insert(group_or_user);
```
Kobzol, line 1011: Hmm, but this tests more things than just vacation, right? :thinking:

Kobzol: Ok, fair enough.

PR #1922, opened 2025-04-02
Extract non-default-branch and modifies-submodules warnings out of the `assign` handler

Urgau:
This PR is the first step of https://github.com/rust-lang/triagebot/pull/1901#issuecomment-2770008801

It extracts non-default-branch and modifies-submodules warnings out of the `assign` handler and creates a `check_commits` handler.

r? @Kobzol

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -27,6 +27,7 @@ impl fmt::Display for HandlerError {
 mod assign;
 mod autolabel;
 mod bot_pull_requests;
+mod check_commits;
 mod close;
 pub mod docs_update;
 mod github_releases;
@@ -69,6 +70,16 @@ pub async fn handle(ctx: &Context, event: &Event) -> Vec<HandlerError> {
 
     if let Some(body) = event.comment_body() {
         handle_command(ctx, event, &config, body, &mut errors).await;
+    }
+
+    if let Ok(config) = &config {
+        if let Err(e) = check_commits::handle(ctx, event, &config).await {
+            log::error!(
+                "failed to process event {:?} with `check_commits` handler: {:?}",
+                event,
+                e
+            );
+        }
     }
 
     if let Err(e) = project_goals::handle(ctx, event).await {
--- a/src/handlers/check_commits.rs
+++ b/src/handlers/check_commits.rs
@@ -1,0 +1,53 @@
+use anyhow::bail;
+
+use super::Context;
+use crate::{
+    config::Config,
+    github::{Event, IssuesAction},
+};
+
+mod modified_submodule;
+mod non_default_branch;
+
+pub(super) async fn handle(ctx: &Context, event: &Event, config: &Config) -> anyhow::Result<()> {
+    let Event::Issue(event) = event else {
+        return Ok(());
+    };
+
+    if !matches!(event.action, IssuesAction::Opened) || !event.issue.is_pr() {
+        return Ok(());
+    }
+
+    let Some(diff) = event.issue.diff(&ctx.github).await? else {
+        bail!(
+            "expected issue {} to be a PR, but the diff could not be determined",
+            event.issue.number
+        )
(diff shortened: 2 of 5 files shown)


src/handlers/check_commits/non_default_branch.rs
```diff
@@ -0,0 +1,35 @@
+use crate::{config::WarnNonDefaultBranchException, github::IssuesEvent};
+
+const NON_DEFAULT_BRANCH: &str =
```
Kobzol, line 3: I know that it's pre-existing, but could you please turn these strings from constants into functions (with placeholders as arguments), to make sure that we never forget to pass the required placeholders? I eventually want to do this with all such string constants in the codebase.

src/handlers/check_commits.rs
```diff
@@ -0,0 +1,53 @@
...
+        bail!(
+            "expected issue {} to be a PR, but the diff could not be determined",
+            event.issue.number
+        )
+    };
+
+    let mut warnings = Vec::new();
+
+    if let Some(assign_config) = &config.assign {
+        // For legacy reasons the non-default-branch and modifies-submodule warnings
```
Kobzol, line 31: We might want to eventually introduce a `[check_commits]` config option, but that doesn't need to happen in this PR.

Kobzol (commented):
Thank you! Looks good, left a few nits.

src/handlers/check_commits.rs
Urgau: Agree, I think it make sense to have a `[check_commits]` config for new warnings.

src/handlers/check_commits/non_default_branch.rs
Urgau: Seems reasonable, done.

Also took the time to rework the exception handling since it was becoming too complex for my liking.

src/handlers.rs
```diff
@@ -71,6 +72,16 @@ pub async fn handle(ctx: &Context, event: &Event) -> Vec<HandlerError> {
         handle_command(ctx, event, &config, body, &mut errors).await;
     }
 
+    if let Ok(config) = &config {
```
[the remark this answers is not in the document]
Kobzol: Oh, I didn't know that. Ok, let's keep it like this then.

Kobzol (approved):
Thanks! I tried it on my test repo and it seems to work fine. I will try to implement some end-to-end tests for this later, as it looks rather simple.

src/handlers/check_commits/non_default_branch.rs
```diff
@@ -0,0 +1,44 @@
...
+
+fn not_default_branch_warn(default_: &str, target: &str) -> String {
+    format!(
+        "Pull requests are usually filed against the {default_} branch for this repo, \
+         but this one is against {target}. \
+         Please double check that you specified the right target!"
+    )
+}
+
+fn not_default_exception_branch_warn(default_: &str, target: &str) -> String {
```
Kobzol, line 38: Just out of curiosity: why the underscore? `default` is not a keyword, right?

src/handlers/check_commits.rs
```diff
@@ -0,0 +1,53 @@
+use anyhow::bail;
```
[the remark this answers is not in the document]
Urgau: (that's the [2018 style](https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-modrs) ;-))

src/handlers/check_commits/non_default_branch.rs
Urgau: I believe `default` is a contextual keyword, for used for specialization (in `trait` impls), but it's not the case here, I think I was confused by my editor highlighting.

Fixed.

Kobzol (approved):
Great, thank you!

PR #1923, opened 2025-04-04
Require `test` job to succeed before starting `deploy` job

Kobzol:
I could sweat that I added the `needs` attribute there before, but it wasn't there.. :laughing:


Urgau (approved):

Kobzol:
Good, now the deploy job waits for `test` to finish. The Docker image build in the `deploy` job runs tests anyway, but not all of them, and doesn't check linting/formatting.

PR #1926, opened 2025-04-10
Improve canonicalize issue links handler

### LINKED ISSUE #1: command testing

Urgau:
This PR improves the canonicalize issue links handler by:
1. Taking into account ignore blocks (eg. `<!-- Fixes #1 -->`, shouldn't be canonicalized)
2. Taking every issue links into account, not just the action ones (not limited anymore to `Fixes` and the others)

To do the first I had to implement our own ~~custom~~ `Regex::replace_all` function which takes into account the ignore blocks from our parser.

### DIFF
--- a/parser/src/ignore_block.rs
+++ b/parser/src/ignore_block.rs
@@ -1,5 +1,6 @@
 use pulldown_cmark::{Event, Parser, Tag, TagEnd};
-use std::ops::Range;
+use regex::{Captures, Regex, Replacer};
+use std::{borrow::Cow, cell::LazyCell, ops::Range};
 
 #[derive(Debug)]
 pub struct IgnoreBlocks {
@@ -61,6 +62,27 @@ impl IgnoreBlocks {
         }
         None
     }
+}
+
+pub fn replace_all_outside_ignore_blocks<'h, R: Replacer>(
+    re: &Regex,
+    haystack: &'h str,
+    mut replacement: R,
+) -> Cow<'h, str> {
+    let ignore_blocks = LazyCell::new(|| IgnoreBlocks::new(haystack));
+    re.replace_all(haystack, |c: &Captures| {
+        let m = c.get(0).unwrap();
+        // This is the "custom" part, we check if the capture range does not
+        // overlap with a ignore range, if it does we use the match as-is,
+        // otherwise we apply the replacement.
+        if ignore_blocks.overlaps_ignore(m.range()).is_some() {
+            m.as_str().to_string()
+        } else {
+            let mut out = String::new();
+            replacement.replace_append(c, &mut out);
+            out
+        }
+    })
 }
 
 #[cfg(test)]
--- a/src/handlers/canonicalize_issue_links.rs
+++ b/src/handlers/canonicalize_issue_links.rs
@@ -15,11 +15,8 @@ use crate::{
     handlers::Context,
 };
 
-// Taken from https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue?quot#linking-a-pull-request-to-an-issue-using-a-keyword
-static LINKED_RE: LazyLock<Regex> = LazyLock::new(|| {
-    Regex::new("(?i)(?P<action>close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved)(?P<spaces>:? +)(?P<issue>#[0-9]+)")
-        .unwrap()
-});
+static LINKED_RE: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"(?P<start> |^)(?P<issue>#[0-9]+)\b").unwrap());
 
 pub(super) struct CanonicalizeIssueLinksInput {}
 
@@ -65,7 +62,7 @@ pub(super) async fn handle_input(
 }
 
 fn fix_linked_issues<'a>(body: &'a str, full_repo_name: &str) -> Cow<'a, str> {
-    let replace_by = format!("${{action}}${{spaces}}{full_repo_name}${{issue}}");
+    let replace_by = format!("${{start}}{full_repo_name}${{issue}}");
     parser::replace_all_outside_ignore_blocks(&LINKED_RE, body, replace_by)
 }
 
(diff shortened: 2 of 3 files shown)


src/handlers/canonicalize_issue_links.rs
```diff
@@ -15,11 +15,8 @@ use crate::{
     handlers::Context,
 };
 
-// Taken from https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue?quot#linking-a-pull-request-to-an-issue-using-a-keyword
-static LINKED_RE: LazyLock<Regex> = LazyLock::new(|| {
-    Regex::new("(?i)(?P<action>close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved)(?P<spaces>:? +)(?P<issue>#[0-9]+)")
-        .unwrap()
-});
+static LINKED_RE: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(" (?P<issue>#[0-9]+)(?P<end>[^a-zA-Z0-9])").unwrap());
```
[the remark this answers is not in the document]
Urgau: Funny thing, the end was also a problem. Fixed.

parser/src/ignore_block.rs
```diff
@@ -63,6 +64,40 @@ impl IgnoreBlocks {
...
+
+    let ignore_blocks = IgnoreBlocks::new(haystack);
+    let mut new = String::with_capacity(haystack.len());
+    for caps in it {
+        let m = caps.get(0).unwrap();
+        new.push_str(&haystack[last_match..m.start()]);
+        // This is the "custom" part, we check if the capture range does not
+        // overlaps with a ignore range, if it does we use the match as-is,
+        // otherwise we apply the replacer.
+        if ignore_blocks.overlaps_ignore(m.range()).is_some() {
```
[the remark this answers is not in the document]
Urgau: Isn't the `ignore` list created sequential from the markdown event stream? meaning it's already sorted.

parser/src/ignore_block.rs
```diff
@@ -63,6 +64,40 @@ impl IgnoreBlocks {
     }
 }
 
+pub fn replace_all_and_ignores<'h, R: Replacer>(
```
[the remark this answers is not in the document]
Urgau: Much better, thks.

parser/src/ignore_block.rs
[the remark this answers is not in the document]
Kobzol: Yeah, it should be. It's still iterated linearly though. But let's not deal with that, usually the number of ignore blocks will be so small that iterating linearly is almost surely faster, now that I think of it.

parser/src/ignore_block.rs
```diff
@@ -63,6 +64,27 @@ impl IgnoreBlocks {
...
+    re: &Regex,
+    haystack: &'h str,
+    mut replacement: R,
+) -> Cow<'h, str> {
+    let ignore_blocks = LazyCell::new(|| IgnoreBlocks::new(haystack));
```
Kobzol, line 72: Cool usage of `LazyCell` :)

parser/src/ignore_block.rs
```diff
@@ -63,6 +64,27 @@ impl IgnoreBlocks {
     }
 }
 
+pub fn replace_all_and_ignores<'h, R: Replacer>(
```
Kobzol, line 67: Could you please rename the function? I don't know what `and_ignores` means, tbh :sweat_smile: And also add a short doc comment please.

(Sorry, wanted to post this comment in the first review, but then I changed the comment to "please rewrite the function" xD).

src/handlers/canonicalize_issue_links.rs
```diff
@@ -15,10 +15,8 @@ use crate::{
     handlers::Context,
 };
 
-// Taken from https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue?quot#linking-a-pull-request-to-an-issue-using-a-keyword
 static LINKED_RE: LazyLock<Regex> = LazyLock::new(|| {
-    Regex::new("(?i)(?P<action>close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved)(?P<spaces>:? +)(?P<issue>#[0-9]+)")
-        .unwrap()
+    Regex::new("(?P<start> |^)(?P<issue>#[0-9]+)(?P<end>[^a-zA-Z0-9]|$)").unwrap()
```
Kobzol, line 19: Maybe we can also just use `( |$)` for the `end` part? Otherwise things like `#1234č` will still get canonicalized, which is not linked by GitHub. Well, it's a very big edge-case anyway :) So up to you.

parser/src/ignore_block.rs
Urgau: If you have a better name I'm all hears.

The only I can come up with that is descriptive enough is `replace_all_with_ignore_blocks_taken_into_account`.

Kobzol: `replace_all_outside_ignores` or `replace_all_outside_ignore_blocks`?

src/handlers/canonicalize_issue_links.rs
Urgau: Except that `#1.` is valid so we would need to include it, idk if it's worth the trouble.

Kobzol: Ah, good point. Then `\b`?

parser/src/ignore_block.rs
Urgau: When with `replace_all_outside_ignore_blocks`.

src/handlers/canonicalize_issue_links.rs
Urgau: Sure, let's use that.

src/handlers/canonicalize_issue_links.rs
```diff
@@ -15,11 +15,8 @@ use crate::{
     handlers::Context,
 };
 
-// Taken from https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue?quot#linking-a-pull-request-to-an-issue-using-a-keyword
-static LINKED_RE: LazyLock<Regex> = LazyLock::new(|| {
-    Regex::new("(?i)(?P<action>close|closes|closed|fix|fixes|fixed|resolve|resolves|resolved)(?P<spaces>:? +)(?P<issue>#[0-9]+)")
-        .unwrap()
-});
+static LINKED_RE: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"(?P<start> |^)(?P<issue>#[0-9]+)(?P<end>\b)").unwrap());
```
Kobzol, line 19: `\b` is just an assertion, it doesn't match anything (https://regex101.com/r/W8p1ED/1), so we don't even need to capture and replace it, just have it at the end of the regex.

Urgau: Even better, thks.

Kobzol (approved):
Thanks! Dealing with regexes is always a pain...

PR #1930, opened 2025-04-11
Revert "Use GitHub Actions cache to make deploys faster"

Kobzol:
Reverts rust-lang/triagebot#1925

I forgot that on triagebot (unlike bors), we deploy directly from the merge queue, and sadly there the cache doesn't work (https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Help.20with.20rust-analyzer's.20CI/with/510973141) :( So this only made things slower, because we were uploading the cache, but never consuming it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,15 +39,8 @@ jobs:
       - name: Checkout the source code
         uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


Urgau:
I also noticed that it was much slower. Sad that we can't cache it.

PR #1931, opened 2025-04-11
refactor: reduce allocations in `no_mentions`

snprajwal:
This patch adds the below optimisations:

- Allocate a single formatted string for the list of commits instead of a mutable vector which may possibly be resized at runtime.
- Use a single `format!` to build the full warning string.
- Eliminate the `mentions_in_commits_warn` function which moved the vector in the parameter and allocated a new string.

r? @Urgau

### DIFF
--- a/src/handlers/check_commits/no_mentions.rs
+++ b/src/handlers/check_commits/no_mentions.rs
@@ -1,37 +1,27 @@
 //! Purpose: When opening a PR, or pushing new changes, check for github mentions
 //! in commits and notify the user of our no-mentions in commits policy.
 
-use std::fmt::Write;
-
 use crate::{config::NoMentionsConfig, github::GithubCommit};
 
 pub(super) fn mentions_in_commits(
     _conf: &NoMentionsConfig,
     commits: &[GithubCommit],
 ) -> Option<String> {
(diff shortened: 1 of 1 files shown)


src/handlers/check_commits/no_mentions.rs
```diff
@@ -1,37 +1,29 @@
...
+            String::from(
+                r"There are username mentions (such as `@user`) in the commit messages of the following commits.
+  *Please remove the mentions to avoid spamming these users.*
+",
+            ) + &mentions_commits,
+        )
     }
-
-    warning
 }
```
Urgau, line 27: @Kobzol, you had some opinions in https://github.com/rust-lang/triagebot/pull/1922#discussion_r2026800801, how about what's being proposed here?

Kobzol: Performance is irrelevant here, this is not code is not hot. That being said, the iterator change looks pretty nice, and a separate function probably wasn't needed.

The second change is a bit weird though, it seems perplexing and unidiomatic to create a string and then  immediately add another one to it using `+`. We should just use `format!` there.

Kobzol (approved):

Urgau (approved):

Issue #1924, opened 2025-04-09
Warn when commits contain issue links


Kobzol:
Although issue links (`#12345`) in commit messages are not as bad as user mentions (`@user`), it is still better if they are not present, to avoid spamming the mentioned issues.

PR #1932, opened 2025-04-12
Add warning against un-canonicalized issue links in commits

Urgau:
This PR adds a `check_commits` handler for un-canonicalized issue links in commits.

I choose to re-use the newly added `[canonicalize-issue-links]` config from https://github.com/rust-lang/triagebot/pull/1897.

Fixes https://github.com/rust-lang/triagebot/issues/1924
r? @Kobzol


src/handlers/check_commits/canonicalize_issue_links.rs
```diff
@@ -0,0 +1,56 @@
...
+        .filter(|c| super::super::canonicalize_issue_links::LINKED_RE.is_match(&c.commit.message))
+        .map(|c| format!("    - {}\n", c.sha))
+        .collect::<String>();
+
+    if uncanonicalized_issue_links_commits.is_empty() {
+        None
+    } else {
+        Some(format!(
+            r"There are un-canonicalized issue links (such as `#123`) in the commit messages of the following commits.
+  *Please canonicalize them by prefixing them with `{full_repo_name}#` (such as `{full_repo_name}#123`).*
```
Urgau, line 19: Contrary to https://github.com/rust-lang/triagebot/issues/1924, I'm proposing to canonicalize them, since I can see valid reasons to link issues in individual commits.

Urgau: Hum, but your reasoning about avoiding those spamming references to commits in issues is compelling.

src/handlers/check_commits.rs
```diff
@@ -7,6 +7,9 @@ use crate::{
     github::{Event, IssuesAction, IssuesEvent, ReportedContentClassifiers},
 };
 
+#[cfg(test)]
+use crate::github::GithubCommit;
```
Kobzol, line 11: You can also do the import locally in the `dummy_commit_from_body` function and spell out the full path to the return type to avoid the cfg. I would perhaps move these test helpers functions to the dedicated `tests` module, but we don't need to do that in this PR.

src/handlers/check_commits/canonicalize_issue_links.rs
```diff
@@ -0,0 +1,56 @@
+use crate::{config::CanonicalizeIssueLinksConfig, github::GithubCommit};
+
+pub(super) fn issue_links_in_commits(
+    _conf: &CanonicalizeIssueLinksConfig,
```
Kobzol, line 4: Is the config here as a type system proof that the functionality was enabled?

src/handlers/check_commits/canonicalize_issue_links.rs
Kobzol: Canonicalized is better than just raw issue links, but they can still spam, yeah.

src/handlers/check_commits.rs
Urgau: I can't just import it in `dummy_commit_from_body` since it's used as the return type of the function.

src/handlers/check_commits/canonicalize_issue_links.rs
Urgau: Yes, it is. It's a for future-proofing, in case we add configuration options.

src/handlers/check_commits/canonicalize_issue_links.rs
Urgau: Issue links being spammy in commits feels IMO being better reason anyway to avoid them in commits than being un-canonicalized. It also seems easier to justify to contributor to do the work to remove them.

Urgau: Switch to warn against every issue links and ask to remove them.

Kobzol: I think that if you want to protect against issue links in PR descriptions, you will most likely also want to do it in commit messages. So I would ideally like to unify both and rename them, but I guess that we don't have a good story for atomic renames of triagebot options, without doing some deprecation period or aliases or something.

So let's add a new feature. Maybe `[detect-commit-issue-links]`? But what you proposed is also fine, so up to you.

src/handlers/check_commits.rs
Kobzol: Yeah as I saw, you could fully spell out the absolute path in the return type, but it doesn't matter.

src/handlers/check_commits/canonicalize_issue_links.rs
Urgau: > but I guess that we don't have a good story for atomic renames of triagebot options

No need to atomic update, but can just add an alias, so that's what I did, the new handler `issue-links` aliases the old one `canonicalize-issue-links`.

Kobzol (approved):
Thanks!

PR #1935, opened 2025-04-15
Refactor assignment logic slightly

Kobzol:
Before moving forward with using the PR workqueue for determining who to assign, I want to first refactor the assignment logic to make it easier to follow and understand, and also write more tests for it.

This PR just slightly refactors a few things and removes/adds comments.

### DIFF
--- a/parser/src/command/assign.rs
+++ b/parser/src/command/assign.rs
@@ -15,13 +15,13 @@ use std::fmt;
 #[derive(PartialEq, Eq, Debug)]
 pub enum AssignCommand {
     /// Corresponds to `@bot claim`.
-    Own,
+    Claim,
     /// Corresponds to `@bot release-assignment`.
-    Release,
+    ReleaseAssignment,
     /// Corresponds to `@bot assign @user`.
-    User { username: String },
+    AssignUser { username: String },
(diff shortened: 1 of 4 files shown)


parser/src/command/assign.rs
```diff
@@ -14,9 +14,13 @@ use std::fmt;
...
 #[derive(PartialEq, Eq, Debug)]
 pub enum AssignCommand {
+    /// Corresponds to `@bot claim`.
     Own,
+    /// Corresponds to `@bot release-assignment`.
     Release,
+    /// Corresponds to `@bot assign @user`.
     User { username: String },
+    /// Corresponds to `r? [@]user`.
     ReviewName { name: String },
```
Urgau, line 24: :heart: 

We should also think about renaming them to their actual command name. `Own` is particularly inscrutable without the command.

Urgau (approved):
Looks good to me.

parser/src/command/assign.rs
Kobzol: Done that!

Kobzol:
Based on Zulip discussion, I removed commits from this PR that changed the assignment logic, and only kept pure refactoring commits.

Urgau (approved):

PR #1936, opened 2025-04-15
Refactor `no_merges` handler into the `check_commits` super-handler

Urgau:
This PR refactors the `no_merges` handler into the `check_commits` super-handler.

In the process I introduced to the `check_commits` super-handler:
 - handling of labels (add & remove)
 - handling of draft/rollup PRs

Step 3-ish of the plan in https://github.com/rust-lang/triagebot/pull/1901#issuecomment-2770008801

`no_merges` is now ridiculously simple and short compared to before. Our efforts are paying of big!


Urgau:
The warning will look like this.

-----

 - The following commits have merge commits (commits with multiple parents) in your changes. We have a [no merge policy](https://rustc-dev-guide.rust-lang.org/git.html#no-merge-policy) so these commits will need to be removed for this pull request to be merged.

    - 440437ed90c75925a6463df797e4f72754dcc2f8

    You can start a rebase with the following commands:
    ```shell-session
    $ # rebase
    $ git pull --rebase https://github.com/{repository_name}.git {default_branch}
    $ git push --force-with-lease
    ```

src/handlers/check_commits/no_merges.rs
```diff
@@ -0,0 +1,105 @@
...
+    // Don't trigger if the PR has any of the excluded title segments.
+    if config
+        .exclude_titles
+        .iter()
+        .any(|s| event.issue.title.contains(s))
+    {
+        return None;
+    }
+
+    let mut merge_commits = HashSet::new();
```
Kobzol, line 25: Let's uset `BTreeSet` to have a deterministic order. Having a random order could break the check for "do we have the same warning as the last time?" even if the commits don't change.

Kobzol (commented):
Nice! Do you think you could write a simple test for this please? :innocent:

src/handlers/check_commits/no_merges.rs
Urgau: Done and added the some end-to-end tests.

Kobzol (approved):
Awesome, thanks!

Issue #1806, opened 2024-05-03
Feature request: support manual r? without automatic reviewer assignment


RalfJung:
In Miri we generally don't use automatic reviewer assignment, instead one of us just grabs a PR when something still needs review. It would still be nice to be able to use `r?` though. So it would be nice if triagebot could let us use `r?` without automatically assigning someone to every new PR.

apiraino:
@RalfJung ok so basically disable automatic assignment when a new PR with `A-miri` label is created, correct? There are also a number of labels in the same area (`A-mir*`). Should they be ignored as well?

RalfJung:
This is for PRs in https://github.com/rust-lang/miri/, so no there's no label involved. I am just looking for a way to say in triagebot.toml that we want to support `r?` but not have anyone automatically assigned.

PR #1937, opened 2025-04-16
Allow using r? even without the owners table

Kobzol:
The highfive bot hasn't been used in quite some time.

Fixes: https://github.com/rust-lang/triagebot/issues/1806

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -498,11 +498,6 @@ pub(super) async fn handle_command(
                 return Ok(());
             }
             AssignCommand::RequestReview { name } => {
-                if config.owners.is_empty() {
-                    // To avoid conflicts with the highfive bot while transitioning,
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

Kobzol:
I was wondering why I can't use `r?` on rust-analyzer :laughing: Luckily it was a simple fix.

ehuss:
Can you also update the documentation? I think [this line](https://github.com/rust-lang/rust-forge/blob/8d6339636d4ee0c095d67d6bbedf7f3056b04c34/src/triagebot/pr-assignment.md?plain=1#L75) should probably be replaced with something that says the existence of the `[assign]` table enables `r?` to work.

Kobzol:
https://github.com/rust-lang/rust-forge/pull/835

PR #1938, opened 2025-04-16
Unify `@rustbot assign` and `r?` behavior on PRs

Kobzol:
I understand that "assign" has a different semantic than "request review", and that "assign a team" might sound weird. But if we want to respect all the review rules, we shouldn't just allow users to assign anyone (currently the code does a check for a vacation, but in the future that check will be more complicated, and should be unified with `r?`), and IMO it's confusing that these two commands behave differently.

I think that we can do the following things:
- Completely unify `assign` and `r?` on PRs
- Ban `assign` on PRs
- Unify `assign` with `r?`, but do not allow `assign` to work with teams (shouldn't be that hard to do)

Let me know which one do you prefer (or if you have a different proposal :) ).

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -474,22 +474,8 @@ pub(super) async fn handle_command(
             // welcome message).
             return Ok(());
         }
-
         let requested_name = match cmd {
             AssignCommand::Claim => event.user().login.clone(),
-            AssignCommand::AssignUser { username } => {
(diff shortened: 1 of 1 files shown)


Kobzol:
Rebased.

Kobzol:
https://github.com/rust-lang/triagebot/pull/1939 was merged sooner, and it already unified the logic (with the caveat that `assign <team>` doesn't apply team labels), so closing this one.

PR #1939, opened 2025-04-16
Refactor and unify assignment logic

Kobzol:
Follow-up to https://github.com/rust-lang/triagebot/pull/1935.

This PR refactors the reviewer candidate selection logic with the following goals:
- Unify the reviewer selection logic inside `find_reviewer_from_names` and `candidate_reviewers_from_names`, so that various special cases are not scattered amongst other functions in the assign handler.
- Make the candidate selection logic easier to understand (and extend, in preparation of using review preferences for determining reviewer candidates).
- Improve error messages, by providing a specific error message when exactly one reviewer is requested. Before, we would just show a generic message like "the user is on vacation or something or something", even though we knew the exact cause of why they couldn't be assigned.
- Refactor tests to make them easier to be extended in the future. This is again in preparation of review preferences, where the tests will start needing access to the database.

Best reviewed commit-by-commit.

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -282,9 +282,6 @@ async fn determine_assignee(
     let db_client = ctx.db.get().await;
     let teams = crate::team_data::teams(&ctx.github).await?;
     if let Some(name) = find_assign_command(ctx, event) {
-        if is_self_assign(&name, &event.issue.user.login) {
-            return Ok((Some(name.to_string()), true));
-        }
         // User included `r?` in the opening PR body.
         match find_reviewer_from_names(&db_client, &teams, config, &event.issue, &[name]).await {
             Ok(assignee) => return Ok((Some(assignee), true)),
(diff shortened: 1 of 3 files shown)


Urgau (commented):
r=me on the first four, will look at the other ones later

src/handlers/assign.rs
```diff
@@ -739,6 +736,13 @@ async fn find_reviewer_from_names(
     issue: &Issue,
     names: &[String],
 ) -> Result<String, FindReviewerError> {
+    // Fast path for self-assign, which is always allowed.
+    if let [name] = names {
+        if is_self_assign(&name, &issue.user.login) {
+            return Ok(name.clone());
+        }
+    }
```
Urgau, line 667: I'm a bit surprised by this.

Intuitively, I would expect only team member (to be more precise, only people who have write access) to be able to be assigned on a pull request. Issue, no problem.

Urgau (approved):
Look good to me. Only a small remark/question.

Kobzol:
Rebased.

src/handlers/assign.rs
Kobzol: That being said, we have the `ReviewerIsPrAuthor` error, so it's not actually possible for *other people* to assign the PR author :laughing: That's inconsistent, yeah.

Kobzol: I guess we can do that in a separate PR, asked about it on Zulip.

Kobzol:
Let's see what happens.

PR #1941, opened 2025-04-17
Fix new-line at the start in `no_merges` handler

Urgau:
See https://github.com/rust-lang/rust/pull/139921#issuecomment-2810885267 for the example of what happens with a new-line at the start.


Kobzol (approved):
Thanks!

PR #1942, opened 2025-04-17
Suggest update the branch when behind too many commits

xizheyin:
As discussion in [#t-compiler/rustc-dev-guide > A reminder in the doc to try to work with &#96;rust-lang/rust&#96;](https://rust-lang.zulipchat.com/#narrow/channel/196385-t-compiler.2Frustc-dev-guide/topic/A.20reminder.20in.20the.20doc.20to.20try.20to.20work.20with.20.60rust-lang.2Frust.60), I add this feature. But I haven't test it (That may takes some time). Do I need to test it locally?

cc @Kobzol

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -49,6 +49,7 @@ pub(crate) struct Config {
     #[serde(alias = "canonicalize-issue-links")]
     pub(crate) issue_links: Option<IssueLinksConfig>,
     pub(crate) no_mentions: Option<NoMentionsConfig>,
+    pub(crate) behind_upstream: Option<BehindUpstreamConfig>,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -427,6 +428,16 @@ pub(crate) struct IssueLinksConfig {}
 #[serde(deny_unknown_fields)]
 pub(crate) struct NoMentionsConfig {}
 
+/// Configuration for PR behind commits checks
+#[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub(crate) struct BehindUpstreamConfig {
+    /// The threshold of days for parent commit age to trigger a warning.
+    /// Default is 7 days if not specified.
+    pub(crate) days_threshold: Option<usize>,
+}
+
 fn get_cached_config(repo: &str) -> Option<Result<Arc<Config>, ConfigurationError>> {
     let cache = CONFIG_CACHE.read().unwrap();
     cache.get(repo).and_then(|(config, fetch_time)| {
@@ -554,6 +565,9 @@ mod tests {
             trigger-files = ["posts/"]
 
             [no-mentions]
+
+            [behind-upstream]
+            days-threshold = 14
         "#;
         let config = toml::from_str::<Config>(&config).unwrap();
         let mut ping_teams = HashMap::new();
@@ -618,6 +632,9 @@ mod tests {
                 }),
                 issue_links: Some(IssueLinksConfig {}),
                 no_mentions: Some(NoMentionsConfig {}),
+                behind_upstream: Some(BehindUpstreamConfig {
+                    days_threshold: Some(14),
+                }),
             }
         );
     }
@@ -637,6 +654,9 @@ mod tests {
             branch = "stable"
 
             [canonicalize-issue-links]
+
+            [behind-upstream]
+            days-threshold = 7
         "#;
         let config = toml::from_str::<Config>(&config).unwrap();
         assert_eq!(
@@ -685,6 +705,9 @@ mod tests {
                 rendered_link: None,
                 issue_links: Some(IssueLinksConfig {}),
                 no_mentions: None,
+                behind_upstream: Some(BehindUpstreamConfig {
+                    days_threshold: Some(7),
+                }),
             }
         );
     }
--- a/src/github.rs
+++ b/src/github.rs
@@ -1458,7 +1458,7 @@ impl Repository {
 
     /// Returns a list of commits between the SHA ranges of start (exclusive)
     /// and end (inclusive).
-    pub async fn commits_in_range(
+    pub async fn github_commits_in_range(
         &self,
         client: &GithubClient,
         start: &str,
@@ -1497,6 +1497,18 @@ impl Repository {
             }
             page += 1;
         }
+    }
+
+    pub async fn github_commit(
+        &self,
+        client: &GithubClient,
+        sha: &str,
+    ) -> anyhow::Result<GithubCommit> {
+        let url = format!("{}/commits/{}", self.url(client), sha);
+        client
+            .json(client.get(&url))
+            .await
+            .with_context(|| format!("{} failed to get github commit {sha}", self.full_name))
     }
 
     /// Retrieves a git commit for the given SHA.
--- a/src/handlers/check_commits/behind_upstream.rs
+++ b/src/handlers/check_commits/behind_upstream.rs
@@ -1,0 +1,102 @@
+use crate::github::{GithubClient, GithubCommit, IssuesEvent, Repository};
+use tracing as log;
+
+/// Default threshold for parent commit age in days to trigger a warning
+pub const DEFAULT_DAYS_THRESHOLD: usize = 7;
+
+/// Check if the PR is based on an old parent commit
+pub async fn behind_upstream(
+    age_threshold: usize,
+    event: &IssuesEvent,
+    client: &GithubClient,
+    commits: &Vec<GithubCommit>,
+) -> Option<String> {
+    log::debug!("Checking if PR #{} is behind upstream", event.issue.number);
+
+    let Some(head_commit) = commits.first() else {
+        return None;
+    };
+
+    // First try the parent commit age check as it's more accurate
+    match is_parent_commit_too_old(head_commit, &event.repository, client, age_threshold).await {
+        Ok(Some(days_old)) => {
+            log::info!(
+                "PR #{} has a parent commit that is {} days old",
+                event.issue.number,
+                days_old
+            );
+
+            return Some(format!(
+                r"This PR is based on an upstream commit that is {} days old.
+
+    *It's recommended to update your branch according to the [rustc_dev_guide](https://rustc-dev-guide.rust-lang.org/contributing.html#keeping-your-branch-up-to-date).*",
+                days_old
+            ));
+        }
+        Ok(None) => {
+            // Parent commit is not too old, log and do nothing
+            log::debug!("PR #{} parent commit is not too old", event.issue.number);
+        }
+        Err(e) => {
+            // Error checking parent commit age, log and do nothing
+            log::error!(
+                "Error checking parent commit age for PR #{}: {}",
+                event.issue.number,
+                e
+            );
+        }
+    }
+
+    None
+}
+
+/// Checks if the PR's parent commit is too old.
+///
+/// This determines if a PR needs updating by examining the first parent of the PR's head commit,
+/// which typically represents the base branch commit that the PR is based on.
+///
+/// If this parent commit is older than the specified threshold, it suggests the PR
+/// should be updated/rebased to a more recent version of the base branch.
+///
+/// Returns:
+/// - Ok(Some(days_old)) - If parent commit is older than the threshold
+/// - Ok(None)
+///     - If there is no parent commit
+///     - If parent is within threshold
+/// - Err(...) - If an error occurred during processing
+pub async fn is_parent_commit_too_old(
+    commit: &GithubCommit,
+    repo: &Repository,
+    client: &GithubClient,
+    max_days_old: usize,
+) -> anyhow::Result<Option<usize>> {
+    // Get the first parent (it should be from the base branch)
+    let Some(parent_sha) = commit.parents.get(0).map(|c| c.sha.clone()) else {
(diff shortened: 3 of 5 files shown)


Kobzol (commented):
Hi, thanks for the PR! We should definitely test this, but I can do it on my test repo.

Since this check essentially checks commits, and should be ideally done after each push (unless we find the API call to be too expensive), could you please move it to the `check_commits` handler?

It would be nice to introduce some `[check_commits]` config option that would contain the individual commit checks as attributes, and migrate the existing commit checks to it (CC @Urgau, WDYT?), but that doesn't need to happen in this PR.

I wonder how this interacts with merge commits, as found e.g. in rust-lang/rust. We should check if `behind_by` only includes merge commits or all commits.

xizheyin:
Yeah, thanks for guide, I'll move it to check_commits.

I'm calling github's [compare api](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#compare-two-commits) here, which should take into account all of the commits, is there anything special about merge commits? Does it need to be considered separately?

Kobzol:
Well, the history isn't linear, so "number of commits between A and B" is not something that can be answered unambigously.

xizheyin:
In my latest change, I eliminated Auto-merge vs Rollup-merge commits, which should be more fair.

For the part of getting the comparing commits, I didn't take paging because I think a page can query up to 250 leading commits, which should be much higher than our threshold, and to save on queries, I only request it once.
@Kobzol

Urgau:
> It would be nice to introduce some `[check_commits]` config option that would contain the individual commit checks as attributes, and migrate the existing commit checks to it

Some of them have configs for them-selves, like [`[no_merges]`](https://github.com/rust-lang/triagebot/blob/ab55185984e3190d52aa503393ab7791f1bd09f4/src/config.rs#L166-L177), won't it be a bit awkward to configure them if they were also under a single `[check_commits]` config? We could prefix the individual config (like `no_merges_labels` and so on), but that since weird and unnecessary.

Kobzol:
Ideally I would like to merge all of these checks under a single config option, to have it be more consistent, but it may not be worth the churn. We could also support both configs and incrementally migrate the repos to the new form, but again, it's probably not worth it.

xizheyin:
I tried it yesterday and it seems like there was a slight problem with the macro expansion, I think this change may require a separate PR. I would like to do it if possible.

Kobzol:
Just to clarify on my earlier comments, we should only take "root" merge commits into account, so essentially merged PRs. A single PR can contain hundreds of commits (for example sync PRs from subtrees), so if we had a limit of 100, it could trigger after a single PR, which is not great.

So really we want to bave a notion of "you are N PRs behind". That being said, each merged PR is a master commit with two parents, the first one is the previous merge and the right one is the tip of the merged PR. I would kind of expect that maybe `behind_by` only "traverses" the first parents? That's what I wanted to check.

xizheyin:
Thank you, I got it. We need to count merge PRs and not just commits, I'll revise it.

Kobzol:
Brainstorming:

I did some experiments and it seems that the `behind_by` attribute is mostly useless to us, as it seems to count all commits. What we could do instead is to run `gh api "repos/rust-lang/rust/commits?author=bors@rust-lang.org"` to go through all recent merge commits, and try to see if the PR's base SHA is in them and on which position. But that sounds a bit expensive.

I checked if we can get something from the PR details (which we can either fetch, but they should be actually sent proactively in the PR status change webhook). It contains an attribute `base.sha`, which tells us the SHA of the `master` commit on which the PR is based. But this SHA is actually reset (to latest master) on each push to the PR, so that doesn't really help us.

So, another idea: we can take the SHA of the PR, and ask for the details of that commit (`gh api repos/kobzol/rust/commits/340272e54ed731ed827f991d75afc9db8aa21c50`), which has a `parents` field. So we could look up the first parent of the PR's SHA, and ask GH again, and see what is the date of that parent commit, and if it's too old, then we can warn.

But when I thought of this solution, I realized that we might not even need the parent commit at all, if we only take a look at the date? When PR HEAD commit date is older than N days, we know that it hasn't been rebased onto the latest master at the user's PC. If they do rebase, the author date will remain the same, but the commit date will be updated. Hmm, no, that wouldn't necessarily work, because even if you rebase, you might still rebase onto an *old local master*, so even though the commit date will be new, it doesn't mean that you have rebased onto a fresh master commit.

So perhaps the approach with checking the HEAD commit and its parent, and then warn if the parent is older than N days might be the way to go.

Urgau:
> So perhaps the approach with checking the HEAD commit and its parent, and then warn if the parent is older than N days might be the way to go.

Seems reasonable to me.

xizheyin:
I updated the code. The current implementation uses a two-step approach to determine if a PR needs updating:

1. Time-based Check (Preferred Method): Calls `is_parent_commit_too_old` to check if the PR's parent commit is too old (default: older than 14 days). If the parent commit is too old, immediately generates a warning message suggesting to update the branch. This is considered a more accurate method and is used first

2. Important Commits Check (Fallback Method): If the time check passes or errors, proceeds with the commit-based check. It gets the difference information between the PR and target branch(by github `Compare` API). We only count `Auto-merge` and `Rollup-merge` commits (important integration commits). If the number of missing important commits exceeds the threshold (default 100), generates a warning message. This method focuses on whether the PR is missing important integration changes, ignoring regular commits.

Since there may be a collection of PRs in the rollup merge, the number we count will only be a little less than the actual PRs, so as to avoid disturbing the users. This is not a bad alternative.

This approach considers both the timeliness of the PR (time perspective) and whether the PR includes the latest important merge commits (functional perspective), providing a comprehensive assessment of whether the PR needs updating.

This is the implementation as I understand it, I don't know if I'm wrong.

Kobzol:
Checking of the commit date looks good. I would just use that, the fallback would be problematic, because some subtree synces can contain fake/empty bors merge commits, and a single PR can contain tens of these. Detecting this based on number of commits simply won't work well in rust-lang/rust, IMO.

xizheyin:
Ok, I'll remove the logic related to checking for commits.

src/handlers/check_commits/behind_master.rs
```diff
@@ -0,0 +1,58 @@
...
+
+            return Some(format!(
+                "This PR is based on a commit that is {} days old. \
+It's recommended to update your branch according to the \
+[Rustc Dev Guide](https://rustc-dev-guide.rust-lang.org/contributing.html#keeping-your-branch-up-to-date).",
+                days_old
+            ));
+        }
+        Ok(None) => {
+            // Parent commit is not too old, continue with the commit count check
```
Kobzol, line 41: The comments need updating.

src/handlers/check_commits/behind_master.rs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        Ok(Some(days_old)) => {
+            log::info!(
+                "PR #{} has a parent commit that is {} days old",
+                event.issue.number,
+                days_old
+            );
+
+            return Some(format!(
+                "This PR is based on a commit that is {} days old. \
```
Kobzol, line 34:
```suggestion
                r"This PR is based on a commit that is {} days old.
```
You can use a multi-line string to avoid the backslashes.

Kobzol, line 34:
```suggestion
                "This PR is based on an upstream commit that is {} days old. \
```

src/github.rs
```diff
@@ -1076,6 +1076,80 @@ impl Issue {
...
+    /// which typically represents the base branch commit that the PR is based on.
+    ///
+    /// If this parent commit is older than the specified threshold, it suggests the PR
+    /// should be updated/rebased to a more recent version of the base branch.
+    ///
+    /// Returns:
+    /// - Ok(Some(days_old)) - If parent commit is older than the threshold
+    /// - Ok(None) - If not a PR, can't get commit details, or parent is within threshold
+    /// - Err(...) - If an error occurred during processing
+    pub async fn is_parent_commit_too_old(
```
Kobzol, line 1092: This logic is IMO too specialized to live in `Issue` directly. I would just put it in `behind_master.rs`.

src/github.rs
```diff
@@ -1076,6 +1076,80 @@ impl Issue {
...
+                    "Failed to get parent commit details for PR #{}: {}",
+                    self.number,
+                    e
+                );
+                return Ok(None);
+            }
+        };
+
+        // Check if the parent commit is too old
+        let parent_date = parent_commit.commit.author.date;
```
Kobzol, line 1143: Let's put this logic into a separate function, and only return the author date of the base branch commit from this function, to avoid it doing too many things.

src/config.rs
```diff
@@ -49,6 +49,7 @@ pub(crate) struct Config {
     #[serde(alias = "canonicalize-issue-links")]
     pub(crate) issue_links: Option<IssueLinksConfig>,
     pub(crate) no_mentions: Option<NoMentionsConfig>,
+    pub(crate) behind_master: Option<BehindMasterConfig>,
```
Kobzol, line 52: The upstream branch might not be named `master` in all repos. I would propose `behind_upstream`.

Kobzol (commented):
Thank you! I think that the implementation looks good now, modulo a few refactorings that I proposed in comments.

xizheyin:
I did some refactor, mainly as followings:
- rename `behind_master` to `behind_upstream`
- move aging checker to `behind_upstream.rs`
- add `github_commit` in `Repository`
- rename `commits_in_range` to `github_commits_in_range`(Because it return a `GithubCommit` instead of `GitCommit`)
- some other nits

Kobzol (commented):
Great! Please remove the stray backtick and squash the commits, and we can merge it :)

xizheyin:
Ok, I finished it. :)

Urgau:
Please wait before merging, I like to take a look. Will look at it tonight.

src/config.rs
```diff
@@ -427,6 +428,15 @@ pub(crate) struct IssueLinksConfig {}
 #[serde(deny_unknown_fields)]
 pub(crate) struct NoMentionsConfig {}
 
+/// Configuration for PR behind commits checks
+#[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct BehindUpstreamConfig {
+    /// The threshold of days for parent commit age to trigger a warning.
+    /// Default is 14 days if not specified.
+    pub(crate) parent_age_threshold: Option<usize>,
```
Urgau, line 437: nit: it's not very apparent what unit the threshold is, and `parent` feels a bit weird/redudent here (it's kind of implicit with `[behind_upstream]`)

suggestion: what about `days_threshold`?

src/github.rs
```diff
@@ -1499,6 +1499,18 @@ impl Repository {
...
+        &self,
+        client: &GithubClient,
+        sha: &str,
+    ) -> anyhow::Result<GithubCommit> {
+        let url = format!("{}/commits/{}", self.url(client), sha);
+        client
+            .json(client.get(&url))
+            .await
+            .with_context(|| format!("{} failed to get github commit {sha}", self.full_name))
+    }
```
Urgau, line 1512: nit: while `Repository::git_commit` only returns a `GitCommit` it has a `author.date` field, making this addition a bit useless for the purpose of this handler, but this seems like a good addition anyway

src/handlers/check_commits/behind_upstream.rs
```diff
@@ -0,0 +1,112 @@
...
+            log::info!(
+                "PR #{} has a parent commit that is {} days old",
+                event.issue.number,
+                days_old
+            );
+
+            return Some(format!(
+                r"This PR is based on an upstream commit that is {} days old.
+It's recommended to update your branch according to the
+[Rustc Dev Guide](https://rustc-dev-guide.rust-lang.org/contributing.html#keeping-your-branch-up-to-date).",
```
Urgau, line 31: This is going to render weirdly as every warning is a bullet point.

```suggestion
                r"This PR is based on an upstream commit that is {} days old.
    
    *It's recommended to update your branch according to the [rustc-dev-guide](https://rustc-dev-guide.rust-lang.org/contributing.html#keeping-your-branch-up-to-date).*",
```

-----
Which will look like this:
- This PR is based on an upstream commit that is 15 days old.
    
    *It's recommended to update your branch according to the [rustc-dev-guide](https://rustc-dev-guide.rust-lang.org/contributing.html#keeping-your-branch-up-to-date).*

xizheyin:
Thanks both of you! 

I realized that the previous code was wrong.

We shouldn't be checking the parent of `issue.head`, because the head seems to be the latest commit, and its parent is probably the last commit we submitted (if any) instead of the base commit.

The array returned by `pulls/xxx/commits` is accessed in order from oldest to newest, so `commits.first()` gets the earliest commit in the current PR, and its parent is what we need.

I've changed it now though. By the way, I think it's possible to do comments on some structures, e.g. some `repository` field can be annotated whether it's an upstream repository or a forked repository, since the `url` returned by the github api is quite diverse.

Urgau (approved):

Kobzol:
Good point, I kind of forgot that PRs can have more than one commit, lol. Thank you!

xizheyin:
This functionility seems to work well. 😁Do we need to extend it to other code repositories?
https://github.com/rust-lang/rustc-dev-guide/pull/2384#issuecomment-2876631306

Urgau:
The original motivation was to do it for rust-lang/rust, so I would fine if we enable it there. 

Feel free to open a PR and assign it to me.

Issue #1894, opened 2025-01-27
Allow `r? @ghost`


Kobzol:
People often do this after a PR is created, but it just results in an error from triagebot. It should ideally unassign the currently assigned reviewer.

apiraino:
(as discussed with Jakub) In case someone requests `r? @ghost` after a PR has already been assigned, the triagebot should:
- unassign the reviewer
- suggest the command `@rustbot unassign` ([docs](https://rustc-dev-guide.rust-lang.org/rustbot.html#issue-claiming)), so the next time the user knows what to do instead.

PR #1943, opened 2025-04-17
Allow review requesting ghost

Kobzol:
This PR makes it possible for the PR author to unassign the current assignee using `r? ghost`, and also the current assignee to unassign themselves using `@rustbot release-assignment`.

Fixes: https://github.com/rust-lang/triagebot/issues/1894

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -87,6 +87,9 @@ const REVIEWER_ALREADY_ASSIGNED: &str =
 
 Please choose another assignee.";
 
+// Special account that we use to prevent assignment.
+const GHOST_ACCOUNT: &str = "ghost";
+
 #[derive(Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 struct AssignData {
     user: Option<String>,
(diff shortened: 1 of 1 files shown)


src/handlers/assign.rs
```diff
@@ -448,10 +451,12 @@ pub(super) async fn handle_command(
...
-                log::trace!(
-                    "ignoring release on PR {:?}, must always have assignee",
-                    issue.global_id()
-                );
+                let current = &event.user().login;
+                if issue.contain_assignee(current) {
+                    issue
+                        .remove_assignees(&ctx.github, Selection::One(&current))
+                        .await?;
+                }
```
Urgau, line 459: Don't we want to maintain our rule of at least one assignee per PR? If I'm not mistaken this will allow the current assignee to remove them-selves but without selecting anyone else.

How about restarting the selection process with the current assignee excluded?

Kobzol: While there are triagebot comments that suggest so, it's not really a rule today. If you open a PR with r? @ghost, it won't be assigned to anyone, so that is existing practice. This change mostly serves to support the use-case of "oh shit, I forgot to add r? @ghost to the PR body, let me fix this".

Urgau: It might be only me but I see `r? ghost` and `release-assignment` as different, the first being "don't assign this PR to anyone" for which doing `r? ghost` after the fact achieve that, while `release-assignment` "Removes your assignment" (per [documentation](https://forge.rust-lang.org/triagebot/index.html#common-command-summary)) doesn't *to me* means remove all assignments.

Kobzol: Right, and this is how it behaves in this PR, right? release-assignment only releases the comment author. I guess that you propose to "reroll" an assignee after you do release-assignment? That sounds reasonable, but I would rather do it in a separate PR. I will remove the release-assignment changes from this PR, and keep only the ghost change. Is that fine with you?

Urgau: :+1: to postponing release-assignment.

Kobzol:
Ok, removed the `release-assignment` part. I wonder where did these two triagebot comments above come from...

Urgau (approved):

PR #1944, opened 2025-04-17
Ignore bors merge commits in issue links handler as nothing can be done

Urgau:
This PR adds an ignore list of `@bors` merge commits as nothing can be done about them.

Reported in miri https://github.com/rust-lang/miri/pull/4259#issuecomment-2813709896


Kobzol (approved):
Yet another place that we should eventually modify once we make the merge messages contain an absolute issue link :grin:

Thanks!

PR #1945, opened 2025-04-18
Handle issue links in parenthesis

Urgau:
Found here https://github.com/rust-lang/rust-clippy/pull/14648#issue-3004048440


Kobzol (approved):
It's a bit confusing that the two files are named the same :sweat_smile: Maybe we should rename the PR body one to `issue_links_pr_description.rs`, or something like that.

PR #1947, opened 2025-04-18
Take review preferences into account when determining reviewers

Kobzol:
This PR implements step 6 of https://github.com/rust-lang/triagebot/issues/1753. When `[assign.review_prefs]` is configured in `triagebot.toml`, triagebot will now take into account the number of PRs already assigned to a reviewer, and their reviewer preferences.

Users can set their review preferences by sending a DM to triagebot on Zulip, and sending `work set-pr-limit <max-prs>` (this was implemented in https://github.com/rust-lang/triagebot/pull/1919). When a reviewer already has `N` PRs assigned where `N >= max-prs`, they will not be considered for assignment.

Note that the way this is currently implemented, the assigned PR tracking only works for `rust-lang/rust`, so it should not be enabled on any other repo.

There are a couple of questions in the design that we should consider (they do not necessarily block this PR though):
- Should the assigned PR tracking be within a repository or across repositories? In other words, do reviewers want to configure a "max assigned PR limit" across the whole of rust-lang, or do they want to have a limit per repository?  
- Should we only consider open PRs or open + draft PRs to be assigned to a given reviewer?
- Should we ignore this limit in certain situations, e.g. when you request a reviewer explicitly (`r? reviewer`) or when we use the adhoc group fallback assigment logic?

CC @apiraino

Requesting a review regarding the implementation details, but I still want to discuss the design on Zulip first and update Forge, so marking as a draft for now. Best reviewer commit-by-commit. I should probably add more logging before we merge this.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -94,6 +94,10 @@ pub(crate) struct PingTeamConfig {
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
 #[serde(deny_unknown_fields)]
+pub(crate) struct AssignReviewPrefsConfig {}
+
+#[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
 pub(crate) struct AssignConfig {
     /// If enabled, then posts a warning comment if the PR is opened against a
     /// different branch than the default (usually master or main).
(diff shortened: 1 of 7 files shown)


jackh726 (approved):
I'm happy to have this land if someone is ready to revert if there are bugs.

One thought: when we assign a PR to someone, do we immediately manually add that PR to their assigned list, or do we wait until the next sync?

Kobzol:
Actually, it's neither. The periodic sync is just a backup mechanism, most PR assignments are performed through webhooks based on the PR assigned event. We expect that this webhook will be sent, so it works like this:
1) Triagebot assigns a PR
2) GitHub triggers a "PR assigned" webhook
3) Triagebot receives the webhook and updates the PR assignment workqueue

PR #1948, opened 2025-04-18
Update many of our dependencies

Urgau:
This PR updates many of our dependencies, only small changes were required (if at all). I purposely didn't include `hyper` and dependent crate since we need to migrate to the 1.0 which requires many changes.


src/github.rs
```diff
@@ -376,7 +376,7 @@ pub struct Issue {
     /// Note that this is constructed via the [`Issue::repository`] method.
     /// It is not deserialized from the GitHub API.
     #[serde(skip)]
-    pub repository: OnceCell<IssueRepository>,
+    pub repository: OnceLock<IssueRepository>,
```
Kobzol, line 379: We should set `rust-version` in `Cargo.toml` now that we started using "recent" Rust features.

src/main.rs
```diff
@@ -428,8 +428,8 @@ fn is_scheduled_jobs_disabled() -> bool {
 }
 
 #[tokio::main(flavor = "current_thread")]
-async fn main() {
-    dotenv::dotenv().ok();
+async fn main() -> anyhow::Result<()> {
+    dotenvy::dotenv()?;
```
Kobzol, line 432: This means that triagebot will crash if `.env` doesn't exist. That doesn't sound right.

src/changelogs/rustc.rs
```diff
@@ -35,7 +34,10 @@ impl<'a> RustcFormat<'a> {
                     self.store_version(h1, section_ast)?;
                 }
 
-                self.current_h1 = Some(String::from_utf8(child_data.content.clone())?);
+                let Some(h1_child_data) = child.first_child().map(|c| c.data.borrow()) else {
+                    anyhow::bail!("unable to retrieve heading (H1) child from changelog");
+                };
+                self.current_h1 = Some(h1_child_data.value.text().unwrap().to_string());
```
Kobzol, line 40: maybe `unwrap()` -> `?` just to avoid potential panics?

Kobzol (approved):

src/main.rs
Urgau: Well, if we don't have it, we don't have the GitHub and Zulip token, we also don't have access to the database (it's connection conf are in it), so we can't do anything.

Urgau: Agree that for the other bins it maybe be too strict, they don't all required it.

src/github.rs
Urgau: Set to `1.80.0`.

src/main.rs
Kobzol: But it's not strictly needed, right? Someone could just configure the environment variables through other means, there are many ways to do it. I don't think that the `.env` file should be *required* to run triagebot :)

Urgau: I was under the impression that the `dotenvy` crate did more than it actually does, mainly that we were reading the values from it and not from the actual environment variables. I therefor agree that we shouldn't require a `.env`.

PR #1952, opened 2025-04-22
Only consider PRs with certain labels to be in the reviewer workqueue

Kobzol:
This PR adds logic for determining if a PR should be considered to be in the reviewer workqueue. The implemented rules are:

- The PR has to be assigned to the given user (duh).
- The PR has to be open.
- The PR has to be non-draft. 
- The PR **must not** contain `S-blocked`, `S-inactive` or `rollup` labels.
- The PR **must** contain the `S-waiting-on-review` label.
- The PR **must not** be assigned to the PR author.

This implementation is based on discussion in https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/New.20triagebot.20assignment.20logic.20respecting.20PR.20limits/with/513565237.

Prerequisite for https://github.com/rust-lang/triagebot/pull/1947.

### DIFF
--- a/src/handlers/pr_tracking.rs
+++ b/src/handlers/pr_tracking.rs
@@ -130,6 +130,22 @@ pub(super) async fn handle_input<'a>(
     Ok(())
 }
 
+/// Loads the workqueue (mapping of open PRs assigned to users) from GitHub
+pub async fn load_workqueue(client: &Octocrab) -> anyhow::Result<ReviewerWorkqueue> {
+    tracing::debug!("Loading workqueue for rust-lang/rust");
+    let prs = retrieve_pull_request_assignments("rust-lang", "rust", &client).await?;
+
+    // Aggregate PRs by user
+    let aggregated: HashMap<UserId, HashSet<PullRequestNumber>> =
+        prs.into_iter().fold(HashMap::new(), |mut acc, (user, pr)| {
+            let prs = acc.entry(user.id).or_default();
+            prs.insert(pr);
+            acc
+        });
+    tracing::debug!("PR assignments\n{aggregated:?}");
+    Ok(ReviewerWorkqueue::new(aggregated))
+}
+
 /// Retrieve tuples of (user, PR number) where
 /// the given user is assigned as a reviewer for that PR.
 /// Only non-draft, non-rollup and open PRs are taken into account.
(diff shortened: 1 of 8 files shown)


src/handlers/pr_tracking.rs
```diff
@@ -210,30 +222,57 @@ pub async fn get_assigned_prs(ctx: &Context, user_id: UserId) -> HashSet<PullReq
...
+/// batch PR loads).
+///
+/// Note: this functionality is currently hardcoded for rust-lang/rust, other repos might use
+/// different labels.
+fn waits_for_a_review(labels: &[Label], is_open: bool, is_draft: bool) -> bool {
+    let is_blocked = labels
+        .iter()
+        .any(|l| l.name == "S-blocked" || l.name == "S-inactive");
+    let is_rollup = labels.iter().any(|l| l.name == "rollup");
+    let is_waiting_for_reviewer = labels.iter().any(|l| l.name == "S-waiting-on-review");
```
Urgau, line 265: I would prefer if this logic was not hard-coded in the code (yes, I know all this is pretty r-l/r specific anyway but still). Can we at least have "positive" labels (`S-waiting-on-review`) and "negative" labels in a `const` slice? Or better, make them configurable in `ReviewPrefsConfig`?

src/zulip.rs
```diff
@@ -291,8 +291,9 @@ async fn workqueue_commands(
                 None => String::from("Not set (i.e. unlimited)"),
             };
 
-            let mut response = format!("Assigned rust-lang/rust PRs: {prs}\n");
-            writeln!(response, "Review capacity: {capacity}")?;
+            let mut response = format!("`rust-lang/rust` PRs in your review queue: {prs}\n");
+            writeln!(response, "Review capacity: {capacity}\n")?;
```
Urgau, line 295: It might be also useful to see the current number of PR in the queue. *Just a reflection, nothing to act up-on.*

Kobzol: Yeah I was also thinking that it would be nice to see the count, added it.

Urgau (approved):

src/handlers/pr_tracking.rs
Urgau: Makes me sad that we have to hard-code them, but oh well.

PR #1953, opened 2025-04-22
Make label removal check case insensitive

### LINKED ISSUE #1915: Triagebot labels parsing case insensitive
I wouldn't mind the triagebot being able to parse label in a case insensitive way (but I don't remember if this was already discussed).

<details><summary>Screenshotting my lazyness</summary>
<p>

Image

</p>
</details> 

If there are no opposing opinions, I can have a look at this.

apiraino:
When removing a label we check if the label exists. This check is case sensitive and therefore fails if the user isn't writing the label correctly.

This patch makes it so, that removing a label is now case insensitive.

Thanks for checking this out!

Ref: https://github.com/rust-lang/triagebot/issues/1915#issuecomment-2701704816
Closes #1915

r? @ehuss

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -705,7 +705,11 @@ impl Issue {
             name = label,
         );
 
-        if !self.labels().iter().any(|l| l.name == label) {
+        if !self
+            .labels()
+            .iter()
+            .any(|l| l.name.to_lowercase() == label.to_lowercase())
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Unless you want a review from Eric specifically, feel free to merge :)

Issue #1954, opened 2025-04-23
merge_conflict handler failing after loading PR


ehuss:
The merge_conflict handler seems to be failing for an unknown reason. The log message is:

```
ERROR triagebot::handlers::merge_conflicts: failed to scan unknown PRs for merge conflicts: rust-lang/mdBook failed to get issue 608
Caused by:
    invalid type: null, expected struct Repository at line 1 column 3365
```

PR #1955, opened 2025-04-24
Fix missing repo in CommitBase

ehuss:
The `repo` field is optional. For example, it can be unset if the repository gets deleted.

Fixes https://github.com/rust-lang/triagebot/issues/1954


Kobzol (approved):
Thanks!

PR #1956, opened 2025-04-24
Retry when failed in github diff

### LINKED ISSUE #1949: GitHub diffs often fail
I regularly see triagebot fail to fetch the diff of a PR, particularly when it is first opened. The error looks something like this:

```
ERROR request{uuid=b44d80a0-20ca-48af-992f-49aa869bf62e}: triagebot: handling event failed: failed to fetch diff comparison for https://api.github.com/repos/rust-lang/rust/compare/c8f94230282a8e8c1148f3e657f0199aad909228...b70763e306992af3e2b2cc06b7abdd8dfb7f2d88
Caused by:
    0: response: {"message":"Not Found","documentation_url":"https://docs.github.com/rest/commits/commits#compare-two-commits","status":"404"}
    1: HTTP status client error (404 Not Found) for url (https://api.github.com/repos/rust-lang/rust/compare/c8f94230282a8e8c1148f3e657f0199aad909228...b70763e306992af3e2b2cc06b7abdd8dfb7f2d88)
```

For some reason, GitHub is temporarily returning 404 for the `/compare/` endpoint. I'm suspecting some kind of race condition on their backend.

This prevents triagebot from processing requests, particularly for assigning reviewers.

I do not know how to address this. One thought is to retry, maybe after a delay, but I don't think we can do that in the webhook window, which means triagebot would need to separately queue and process events. That seems complicated.

xizheyin:
Closes #1949

I set up a retry mechanism, such as retrying 3 times, each time with an interval of 2 second. And the code provide sufficient logs for subsequent troubleshooting. 

I think it might be the case that Triagebot receives the webhook notification of PR creation and executes it immediately, but at this time the GitHub backend might not have finished calculating the PR diff yet, and this kind of unsynchronized processing between services leads to the 404 error. So setting a reasonable retry interval may reduce the error occurrence.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -968,14 +968,47 @@ impl Issue {
                     "{}/compare/{before}...{after}",
                     self.repository().url(client)
                 );
-                let mut req = client.get(&url);
-                req = req.header("Accept", "application/vnd.github.v3.diff");
-                let (diff, _) = client
-                    .send_req(req)
-                    .await
-                    .with_context(|| format!("failed to fetch diff comparison for {url}"))?;
-                let body = String::from_utf8_lossy(&diff);
-                Ok(parse_diff(&body))
+
+                // Maximum number of retry attempts
+                const MAX_RETRIES: u8 = 3;
+                // Delay between retries (in milliseconds)
+                const RETRY_DELAY_MS: u64 = 2000;
+
+                for attempt in 0..=MAX_RETRIES {
+                    let mut req = client.get(&url);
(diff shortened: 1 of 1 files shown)


xizheyin:
Oh, any other solutions?

xizheyin:
It seems to be called often. 
image

Kobzol:
Right, what I meant was that in the assign handler it's only used for one thing. The other usages are not that critical, I think, but it's important to make assign work robustly. Also I suspect that the diff fails in the assign handler more often than elsewhere, because the assign handler is called immediately after the PR being opened.

PR #1957, opened 2025-04-24
Check if connection is valid while get client from poll

### LINKED ISSUE #1887: Scheduled jobs often fail due to db timeout
The scheduled jobs thread regularly fails with the following:

```
database connection error: error communicating with the server: Connection timed out (os error 110)
thread 'main' panicked at /src/db.rs:239:47:
called `Result::unwrap()` on an `Err` value: Getting jobs data
Caused by:
    connection closed
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2025-01-23T17:33:20.088253Z ERROR triagebot: run_scheduled_jobs task died (error=panic)
```

or

```
database connection error: error communicating with the server: Connection timed out (os error 110)
thread 'main' panicked at src/main.rs:377:26:
called `Result::unwrap()` on an `Err` value: database schedule jobs
Caused by:
    0: Inserting job
    1: connection closed
2025-01-23T17:37:11.484202Z ERROR triagebot: schedule_jobs task died (error=panic)
```

I'm uncertain what might be happening. There is code to check for closed connections, but maybe it is getting dropped without being closed properly? It seems peculiar that it is only affecting the scheduled jobs, since I would expect stale cached connections to affect anything touching the database.

xizheyin:
May Fixes #1887

It is possible that the client connections in the pool have not been used for a long time, and the server side automatically closes the connection with the client, however, is_closed only checks the status of the local client and not whether the server side has a connection. I added the check to ensure the robustness.

Alternatively, there are other ways to ensure a more continuous connection.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -3,8 +3,8 @@ use anyhow::Context as _;
 use chrono::Utc;
 use native_tls::{Certificate, TlsConnector};
 use postgres_native_tls::MakeTlsConnector;
-use std::sync::{Arc, LazyLock, Mutex};
-use tokio::sync::{OwnedSemaphorePermit, Semaphore};
+use std::sync::{Arc, LazyLock};
+use tokio::sync::{Mutex, OwnedSemaphorePermit, Semaphore};
 use tokio_postgres::Client as DbClient;
 
 pub mod issue_data;
@@ -37,8 +37,14 @@ pub struct PooledClient {
 
 impl Drop for PooledClient {
     fn drop(&mut self) {
-        let mut clients = self.pool.lock().unwrap_or_else(|e| e.into_inner());
-        clients.push(self.client.take().unwrap());
+        // We can't await in drop, so we need to spawn a task to handle async lock
+        if let Some(client) = self.client.take() {
+            let pool = self.pool.clone();
+            tokio::spawn(async move {
+                let mut clients = pool.lock().await;
+                clients.push(client);
+            });
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


Urgau:
I know much to nothing about our database connections, maybe @ehuss or @Kobzol knows about it, and if this conceptually a good idea.

ehuss:
Just quickly looking at this, it seems like it might have a few issues.
- It doesn't look like it manages the permit. That means the permit count and the pool count could get out of sync, allowing more connections than the semaphore is intended to allow.
- This doesn't directly address https://github.com/rust-lang/triagebot/issues/1887, which is more of a theory about idle connections being terminated, and not knowing that until it has been popped off the queue.
- If the connection is broken, checking it in `get()` isn't going to help, since the call to `validate_connection` could hang, which would be no different from just running a normal query.

When I have more time, I can maybe look at some suggestions. Or maybe @Kobzol has more time right now.

Kobzol:
Tbh I don't think that this is the right approach, as we aren't even sure what was the cause of the original issue. These timeouts also happen in rustc-perf, and it's IMO caused simply by the DB not responding due to being overloaded or a perhaps a network issue. The DB server has been beefied up recently, so maybe these issues aren't even happening anymore? It seems like they stopped happening on rustc-perf once the server was upgraded. So I would keep the connection pooling like it is for now.

xizheyin:
Ok, because I have encountered similar problems before, but there may be various reasons for the original issue. Thank you three for review! I'll close it.

PR #1958, opened 2025-04-24
Add @rustbot reroll to pick a new reviewer

### LINKED ISSUE #1762: add "bors reroll"
https://github.com/rust-lang/rust/pull/119752#issuecomment-1894548307

@bors reroll

to pick a new reviewer

this has been haunting me

xizheyin:
Closes #1762


Kobzol:
Thanks for the PR! I think that we should first discuss the design though. I'm not sure if rerolling based on the diff is the most intuitive solution in all cases. If someone did `r? compiler` before, then rerolling should IMO select someone else from the compiler team/adhoc group, not determine the reviewer based on the file diff. We might thus have to store some per-issue data to remember that a team was previously requested for a review.

I opened a topic about this on Zulip: https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/.40rustbot.20reroll/with/514122139.

xizheyin:
I modified the logic:
Only when `r? group`, we will select a reviewer that has not been previously ASSIGNED. note that this won't work with `r? @username` because we explicitly specify the name.

I'll comment my changes in the code.

src/handlers/assign.rs
```diff
@@ -242,8 +265,14 @@ async fn set_assignee(issue: &Issue, github: &GithubClient, username: &str) {
             .await
         {
             log::warn!("failed to post error comment: {e}");
+            return Err(e);
         }
     }
+
+    state.data.names.insert(username.to_string());
+    state.save().await?;
```
xizheyin, line 273: When set assignee, we store it into database.

src/handlers/assign.rs
```diff
@@ -290,9 +318,11 @@ async fn determine_assignee(
                 ),
                 Err(
                     e @ FindReviewerError::NoReviewer { .. }
+                    // TODO: only NoReviewer can be reached here!
                     | e @ FindReviewerError::ReviewerIsPrAuthor { .. }
                     | e @ FindReviewerError::ReviewerAlreadyAssigned { .. }
-                    | e @ FindReviewerError::ReviewerOnVacation { .. },
+                    | e @ FindReviewerError::ReviewerOnVacation { .. }
+                    | e @ FindReviewerError::ReviewerPreviouslyAssigned { .. },
```
xizheyin, line 325: It's worth noting here. Should we use `unreachable!()` except `NoReviewer`?

src/handlers/assign.rs
```diff
@@ -804,6 +838,9 @@ fn candidate_reviewers_from_names<'a>(
             .iter()
             .any(|assignee| name_lower == assignee.login.to_lowercase());
 
+        let previous_reviewer_names = get_previous_reviewer_names(ctx, issue).await;
+        let is_previously_assigned = previous_reviewer_names.contains(&candidate);
```
xizheyin, line 842: It is used to know if the `candidate` is a previous reviewer.

src/handlers/assign.rs
```diff
@@ -818,6 +855,12 @@ fn candidate_reviewers_from_names<'a>(
                 Some(FindReviewerError::ReviewerAlreadyAssigned {
                     username: candidate.clone(),
                 })
+            } else if expansion_happened && is_previously_assigned {
+                // **Only** when r? group is expanded, we consider the reviewer previously assigned
+                // `r? @reviewer` will not consider the reviewer previously assigned
+                Some(FindReviewerError::ReviewerPreviouslyAssigned {
+                    username: candidate.clone(),
+                })
```
xizheyin, line 863: I've added logic here to determine if it's a previous Reviewer here **only** when `r? group` (expansion happened).

src/handlers/assign.rs
```diff
@@ -863,3 +906,19 @@ fn candidate_reviewers_from_names<'a>(
...
+    // Get the state of the warnings for this PR in the database.
+    let mut db = ctx.db.get().await;
+    let state: IssueData<'_, Reviewer> =
+        match IssueData::load(&mut db, &issue, PREVIOUS_REVIEWER_KEY).await {
+            Ok(state) => state,
+            Err(_) => return HashSet::new(),
+        };
+
+    state.data.names
+}
```
xizheyin, line 924: This corresponds to `set_assignee`, which stores, and this function is used to read from the database. `cfg!(test)` returns null in order to pass the `tests`.

src/handlers/assign/tests/tests_candidates.rs
```diff
@@ -41,10 +61,12 @@ impl TestCtx {
...
     }
 
-    fn run(self, names: &[&str], expected: Result<&[&str], FindReviewerError>) {
+    async fn run(self, names: &[&str], expected: Result<&[&str], FindReviewerError>) {
+        let ctx = create_test_context();
         let names: Vec<_> = names.iter().map(|n| n.to_string()).collect();
         match (
-            candidate_reviewers_from_names(&self.teams, &self.config, &self.issue, &names),
+            candidate_reviewers_from_names(&ctx, &self.teams, &self.config, &self.issue, &names)
+                .await,
```
xizheyin, line 69: Change the tests into async because `candidate_reviewers_from_names` is async.

ehuss:
Can you please update the description with a detailed explanation of what is changing. From the title it is not clear to me what this is doing.

src/handlers/assign.rs
```diff
@@ -678,7 +711,7 @@ async fn find_reviewer_from_names(
         }
     }
 
-    let candidates = candidate_reviewers_from_names(teams, config, issue, names)?;
+    let candidates = candidate_reviewers_from_names(ctx, teams, config, issue, names).await?;
```
Kobzol, line 714: Note that https://github.com/rust-lang/triagebot/pull/1947 completely refactors these tests and provides them access to an actual database, so it might be better to base the code on top of that.

Kobzol (commented):
Now that I think about it, remembering the previous requested name (`r? name`) would be required for `rustbot reroll`, but it shouldn't be required at all when we are just modifying the request logic to skip the existing reviewer.

In other words, you can get the current assignee directly from the `issue` passed to the handler. Then we should modify the assignment logic to ignore the previous assignee if a team (not a specific user) was requested.

src/handlers/assign.rs
xizheyin: Okay, when #1947 is merged, I'll rewrite the test.

xizheyin:
> Can you please update the description with a detailed explanation of what is changing. From the title it is not clear to me what this is doing.

As discussed in [#triagebot > @rustbot reroll @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/.40rustbot.20reroll/near/514136549), I optimized the logic for `r? group`. Each time an assignee is set, it is stored in the database, and when we use an `r? group`, such as `r? compiler`, I will take out the previously assigned reviewer from the database to avoid to assign a previous assignee.

However. the use of `r? @reviewer` to specify a specific assignee does not exclude the previous assignee, as this is explicitly specified.

xizheyin:
> In other words, you can get the current assignee directly from the issue passed to the handler.

`issue.assignees` seems to only return the current assignees, it may need the database if we want to get all the previous assignees of the issue.

ehuss:
> As discussed in [#triagebot > @rustbot reroll @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/.40rustbot.20reroll/near/514136549), I optimized the logic for `r? group`. Each time an assignee is set, it is stored in the database, and when we use an `r? group`, such as `r? compiler`, I will take out the previously assigned reviewer from the database to avoid to assign a previous assignee.
> 
> However. the use of `r? @reviewer` to specify a specific assignee does not exclude the previous assignee, as this is explicitly specified.

OK, can the description be updated to include that?  Or, if the design hasn't been concluded, can this be marked as a draft until it is ready?

Also, this doesn't seem like it would close https://github.com/rust-lang/triagebot/issues/1762, would it? That seems like a different thing of the ability to say "I want triagebot to pick someone else".

Kobzol:
FWIW, we have been discussing this in https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/.40rustbot.20reroll/with/514415765, and me and Jieyou Xu thought that having an explicit reroll command might not be needed if we make this change to the assignment logic.

xizheyin:
I rebase this PR to the newest version of master branch that is compatible with the latest `test_candidates.rs`.

Kobzol:
Hi, just wanted to note that I'll be at https://rustweek.org/, so I might not be able to take a look at this for a few days, but it's at the top of my TODO list :)

xizheyin:
It looks great. Have fun! I don't think the demand for this pr is so urgent.

src/handlers/assign.rs
```diff
@@ -92,9 +93,23 @@ const REVIEWER_ALREADY_ASSIGNED: &str =
 
 Please choose another assignee.";
 
+const REVIEWER_ASSIGNED_BEFORE: &str = "Requested reviewers are assigned before.
```
Kobzol, line 96:
```suggestion
const REVIEWER_ASSIGNED_BEFORE: &str = "Requested reviewer(s) were already assigned before.
```

src/handlers/assign.rs
```diff
@@ -92,9 +93,23 @@ const REVIEWER_ALREADY_ASSIGNED: &str =
...
 // Special account that we use to prevent assignment.
 const GHOST_ACCOUNT: &str = "ghost";
 
+/// Key for the state in the database
+const PREVIOUS_REVIEWER_KEY: &str = "previous-reviewer";
+
+/// State stored in the database
+#[derive(Debug, Clone, PartialEq, Default, serde::Deserialize, serde::Serialize)]
+struct Reviewers {
+    /// List of the last warnings in the most recent comment.
```
Kobzol, line 109: Wrong comment.

src/handlers/assign.rs
```diff
@@ -92,9 +93,23 @@ const REVIEWER_ALREADY_ASSIGNED: &str =
...
 
+const REVIEWER_ASSIGNED_BEFORE: &str = "Requested reviewers are assigned before.
+
+Please choose another assignee by using `r? @reviewer`.";
+
 // Special account that we use to prevent assignment.
 const GHOST_ACCOUNT: &str = "ghost";
 
+/// Key for the state in the database
+const PREVIOUS_REVIEWER_KEY: &str = "previous-reviewer";
```
Kobzol, line 104:
```suggestion
const PREVIOUS_REVIEWERS_KEY: &str = "previous-reviewers";
```

src/handlers/assign.rs
```diff
@@ -680,6 +710,8 @@ enum FindReviewerError {
     ReviewerIsPrAuthor { username: String },
     /// Requested reviewer is already assigned to that PR
     ReviewerAlreadyAssigned { username: String },
+    /// Requested reviewer is already assigned previously to that PR.
```
Kobzol, line 713:
```suggestion
    /// Requested reviewer was already assigned previously to that PR.
```

src/handlers/assign.rs
```diff
@@ -726,6 +758,13 @@ impl fmt::Display for FindReviewerError {
                     REVIEWER_ALREADY_ASSIGNED.replace("{username}", username)
                 )
             }
+            FindReviewerError::ReviewerPreviouslyAssigned { username } => {
+                write!(
+                    f,
+                    "{}",
+                    REVIEWER_ASSIGNED_BEFORE.replace("{username}", username)
```
Kobzol, line 765: `REVIEWER_ASSIGNED_BEFORE` doesn't seem to contain `{username}`? It shouldn't need any replacement.

src/handlers/assign.rs
```diff
@@ -963,6 +1008,12 @@ async fn candidate_reviewers_from_names<'a>(
                 Some(FindReviewerError::ReviewerAlreadyAssigned {
                     username: candidate.clone(),
                 })
+            } else if expansion_happend && is_previously_assigned {
```
Kobzol, line 1011: You don't need to check `expansion_happened` here. We should decide per-reviewer basis if they were expanded or not. It's enough to check `reviewer_candidate.origin` here.

Kobzol (commented):
Looks good, thank you. I would appreciate some tests though :) You could add a method to `AssignCtx` that will store a set of previously assigned names into the DB, and then check that these names will be ignored when expanded from a group/team, and will not be ignored when requested directly.

src/handlers/assign.rs
xizheyin: For more information, I add `{username}` in `REVIEWER_ASSIGNED_BEFORE`, but it does not seem to matter because `REVIEWER_ASSIGNED_BEFORE` will never show in comments, instead it could only return `FindReviewerError::NoReviewer` in `candidate_reviewers_from_names`.

xizheyin (commented):
I add three tests:
1. `r? team`, while there exists people who have not previous assigned in `team`
2. `r? compiler`, however, all the people in `team` have been previously assigned.
3. `r? @user`, regardless of whether `user` has been assigned or not.

src/handlers/assign.rs
Kobzol: I meant that we don't need to fill the username in this error message, as you said, it won't be shown in the comments.

src/handlers/assign.rs
```diff
@@ -280,8 +303,14 @@ async fn set_assignee(issue: &Issue, github: &GithubClient, username: &str) {
             .await
         {
             log::warn!("failed to post error comment: {e}");
+            return Err(e);
         }
     }
+
+    // Record the reviewer in the database
+    state.data.names.insert(username.to_string());
```
Kobzol, line 311: We should handle normalization. Ideally with a newtype, but that's probably overkill here, the code already uses lowercase conversions all over the place.

Let's:
- Normalize the name to lowercase before putting it into the set
- Add a comment to the `names` field in `Reviewers` that explains that the names are normalized to lowercase
- Check `name_lower` instead of `reviewer_candidate.name` when computing `is_previously_assigned`

src/handlers/assign/tests/tests_candidates.rs
```diff
@@ -71,15 +72,28 @@ impl AssignCtx {
         self
     }
 
+    async fn set_previous_reviewers(mut self, users: &[&User]) -> Self {
+        let mut db = self.test_ctx.db_client_mut();
+        for user in users {
```
Kobzol, line 77: Let's just create a set out of `users` and store it into the DB in one go, instead of iterating over each user. The name of the function is called `set`, not add or something, so it should overwrite the DB contents.

Kobzol (commented):
Thanks for the tests! Left some more comments.

Issue #1962, opened 2025-04-28
Delay reviewer assignment for draft PRs


Zoxc:
I think it would make sense for reviewer assignment to be delayed until a PR no longer has a draft status.

PR #1963, opened 2025-04-30
Do not perform PR assignment for draft PRs without r?

Kobzol:
This PR changes the assignment logic to be less eager.

This is what happens after this PR:
- When a PR is opened normally, assignment is performed.
- When a PR is opened as a draft with r? in PR body, assignment is performed.
- When a PR is opened as a draft without r? OR with `r? ghost` in PR body, assignment is not performed (it is delayed).
- When a PR is converted from draft and there are no current assignees, assignment is performed.

Fixes: https://github.com/rust-lang/triagebot/issues/1962

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -90,13 +90,18 @@ Please choose another assignee.";
 // Special account that we use to prevent assignment.
 const GHOST_ACCOUNT: &str = "ghost";
 
+/// Assignment data stored in the issue/PR body.
 #[derive(Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 struct AssignData {
     user: Option<String>,
 }
 
-/// Input for auto-assignment when a PR is created.
-pub(super) struct AssignInput {}
+/// Input for auto-assignment when a PR is created or converted from draft.
+#[derive(Debug)]
+pub(super) enum AssignInput {
+    Opened { draft: bool },
+    ReadyForReview,
+}
 
 /// Prepares the input when a new PR is opened.
 pub(super) async fn parse_input(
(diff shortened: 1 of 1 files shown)


src/handlers/assign.rs
```diff
@@ -90,13 +90,19 @@ Please choose another assignee.";
...
     user: Option<String>,
 }
 
-/// Input for auto-assignment when a PR is created.
-pub(super) struct AssignInput {}
+/// Input for auto-assignment when a PR is created or converted from draft.
+#[derive(Debug)]
+pub(super) enum AssignInput {
+    Opened,
+    OpenedDraft,
```
Urgau, line 103: nit: I think this would be better expressed as `Opened(bool)` with the boolean being the draft status

src/handlers/assign.rs
```diff
@@ -123,8 +135,31 @@ pub(super) async fn handle_input(
...
+    // - PR was opened as a draft with an explicit r? (but not r? ghost)
+    // - PR was converted from a draft and there are no current assignees
+    let should_assign = match input {
+        AssignInput::Opened => true,
+        AssignInput::OpenedDraft => {
+            // Even if the PR is opened as a draft, we still want to perform assignment if r?
+            // was used. However, historically, r? ghost was meant to mean "do not
+            // perform assignment". So in that case, we skip the assignment and only perform it once
+            // the PR has been marked as being ready for review.
+            assign_command.is_some() && assign_command.as_deref() != Some(GHOST_ACCOUNT)
```
Urgau, line 153: nit: maybe a bit clearer
```suggestion
            assign_command.is_some_and(|a| &a != GHOST_ACCOUNT)
```

Urgau (approved):
r=me with or without the nits

Kobzol:
I will leave this open for a few days to see if there are any concerns. If not, I'll merge it then.

PR #1964, opened 2025-05-03
Silence the `ready` reminder when the author is a member of repository

Urgau:
After a few times I'm starting to find the `@bot ready` reminder to be a bit spammy.

I proposes that we don't emit it when the author is a member of the repository, as the author should already know about the `ready` command and already have the required permissions to update the labels manually anyway.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -4,7 +4,7 @@ use bytes::Bytes;
 use chrono::{DateTime, FixedOffset, Utc};
 use futures::{future::BoxFuture, FutureExt};
 use hyper::header::HeaderValue;
-use octocrab::models::Author;
+use octocrab::models::{Author, AuthorAssociation};
 use regex::Regex;
 use reqwest::header::{AUTHORIZATION, USER_AGENT};
 use reqwest::{Client, Request, RequestBuilder, Response, StatusCode};
@@ -396,6 +396,9 @@ pub struct Issue {
     pub milestone: Option<Milestone>,
     /// Whether a PR has merge conflicts.
     pub mergeable: Option<bool>,
+
+    /// How the author is associated with the repository
+    pub author_association: AuthorAssociation,
 }
 
 #[derive(Debug, serde::Deserialize, Eq, PartialEq)]
--- a/src/handlers/shortcut.rs
+++ b/src/handlers/shortcut.rs
@@ -5,10 +5,11 @@
 use crate::{
     config::ShortcutConfig,
     db::issue_data::IssueData,
-    github::{AuthorAssociation, Event, Label},
+    github::{Event, Label},
     handlers::Context,
     interactions::ErrorComment,
 };
+use octocrab::models::AuthorAssociation;
 use parser::command::shortcut::ShortcutCommand;
 
 /// Key for the state in the database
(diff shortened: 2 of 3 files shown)


src/github.rs
```diff
@@ -468,6 +471,19 @@ pub enum PullRequestReviewState {
...
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
+pub enum AuthorAssociation {
+    Collaborator,
+    Contributor,
+    FirstTimer,
+    FirstTimeContributor,
+    Mannequin,
+    Member,
+    None,
+    Owner,
```
Urgau, line 484: The accepted values can be found on the "Response schema" of the issues endpoint, https://docs.github.com/en/rest/issues/issues?apiVersion=2022-11-28#get-an-issue.

Kobzol: Why not use https://docs.rs/octocrab/latest/octocrab/models/enum.AuthorAssociation.html directly?

Urgau: None of the struct or enums in this file uses it, it would be a first. Not opposed but that's just not what we historically did.

Kobzol: I would prefer defaulting to using third-party/octocrab code where possible, and only falling back to custom code when we have a specific use-case for it. We could probably delete a lot of code from triagebot if we just started using octocrab more :)

Kobzol:
If we take a step back, and try to think of a more "perfect" solution, what about one of these?
1) Only send the response once *per user*, not *per PR*, maybe with some repeating, so e.g. only send the reminder once per user per 3 months, or whatever.
2) Actually detect the situations where a PR's last action is commit by the user, but it is still waiting on author. E.g., when a PR is `S-waiting-on-author`, it received a commit/push after the latest state change, and >2 weeks have received, try to ping the PR author about `@bot ready`.

I think that ultimately, 2) is what we want, and we're trying to emulate it with the `@bot author` response. But it could become a bit spammy too.

Urgau:
cc @RalfJung (as the initiator of #1865)

RalfJung:
No strong opinion either way... but regarding (2), I think 2 weeks is way too much. I'd make it more like 24h.

I also made that proposal before knowing about github's saved replies, so I'd now have a workable approach that does not rely on the bot. But I still think the bot could be useful.

Urgau:
> I think that ultimately, 2) is what we want, and we're trying to emulate it with the `@bot author` response. But it could become a bit spammy too.

Yeah, your second option is ideally what I would want. Also agree with Ralf, that 2 weeks is too long, 24 hrs seems fine. We should also make sure CI is green and other stuffs, like `S-blocked`.

Kobzol:
I used 2 weeks as an example, because I often push something that is WIP, but don't want to mark it as ready for review yet, as I'm planning to push more stuff before re-requesting review. So 24h would be "too fast". It's in general difficult to recognize if a PR is really ready for a review :)

RalfJung:
I think most contributors would be very confused and somewhat frustrated if they push and then nothing happens for 2 weeks and *then* they get told what they should have done 2 weeks ago.

Urgau:
> I used 2 weeks as an example, because I often push something that is WIP, but don't want to mark it as ready for review yet, as I'm planning to push more stuff before re-requesting review. So 24h would be "too fast".

Maybe your workflow would be better represented by using the PR draft state? We could simply not issue the warning if the PR is in draft.

Kobzol:
If we take that to the extreme, we could just use draft for "waiting on author", that would make the "I'm waiting for author" situation very obvious and visual :)

I think that draft PRs shouldn't even be in "S-waiting-on-review" often now that we don't auto assign draft PRs.

Urgau:
That would indeed be the logic consequence.

src/github.rs
Urgau: Switched to octocrab.

Kobzol (approved):

Issue #1965, opened 2025-05-04
Split `issue-links` into "canonicalization" and "links in commit"


tgross35:
Per https://forge.rust-lang.org/triagebot/issue-links.html, `[issue-links]` both enforces canonical links and forbids links in the commits. It would be nice to decouple them; the benefit of being able to crosslink issues in history absolutely outweighs a couple possible extra references showing up in the UI, at least for lower volume repositories. Checking canonical links is still independently relevant, though.

tgross35:
More subjectively, I don't think suggesting against links in commits is ever a good idea. Linking to relevant things is part of keeping good, understandable history, and our git history is going to far outlive a silly issue with the GitHub UI.

Urgau:
This handler was primarily design for our subtrees in rust-lang/rust, where banning issue links in commit is meant to both:
 - prevent the issue from being spammed by back-links from rust-lang/rust (and some forks)
 - but also to prevent accidental `Fixes` to the wrong issue when the changes are up-streamed

It should also be noted that the handler doesn't prevent issue links in the PR description which is embedded in the merge commit, which is what the handler is encouraging users to use (albeit we are not explicit about it).

-----

Having an option to disable the "issue links in commit" part should be easily doable.

tgross35:
Fair enough, it does make more sense there since issues get a bit more noisy. I guess I'm just an avoid user of `git blame` who rarely looks at merge commits. In any case, thank you for the quick fix 🙂

PR #1966, opened 2025-05-04
Add option to disable checking issue links in commits

Urgau:
This PR adds an option (~~`allowed-in-commits`~~) to disable checking issue links in commits, this was requested in rust-lang/triagebot#1965 by @tgross35.

More context in https://github.com/rust-lang/triagebot/issues/1965#issuecomment-2849106561, but summary: repository that aren't subtrees in r-l/r don't necessarily need to prevent issue links in commits.

```toml
[issue-links]
check-commits = false # defaults to true
```

Fixes rust-lang/triagebot#1965

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -465,7 +465,10 @@ pub(crate) struct RenderedLinkConfig {
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
 #[serde(rename_all = "kebab-case")]
 #[serde(deny_unknown_fields)]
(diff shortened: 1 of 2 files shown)


Kobzol (approved):
I would find it more intuitive if the option disabled commit checking, rather than allowed issues in commits (e.g. `check-commits = false`), but it's not super important. Feel free to merge as is.

Urgau:
Switched to `check-commits`.

PR #1967, opened 2025-05-04
Automatically indent warnings messages for `check_commits` sub-handler

Urgau:
cf. https://github.com/rust-lang/triagebot/pull/1942#discussion_r2056505677


Kobzol (commented):
Could you please now add a test that will put some example set of warnings through `warning_from_warnings` and assert how does the result look like?

Urgau:
Sure, done.

And sorry, the test it-self is horrible to look at, even after `#[rustfmt::skip]`.

Kobzol (approved):
Looks good. Thanks!

PR #1968, opened 2025-05-06
Do not apply labels in autolabel when opening a draft PR

Kobzol:
After https://github.com/rust-lang/triagebot/pull/1963, it doesn't make sense to apply `S-waiting-on-review` to PRs opened as a draft. According to https://github.com/search?type=code&q=org%3Arust-lang+%22new_pr+%3D+true%22, this option is only ever used for adding `S-waiting-on-review` in the rust-lang org, so I don't think that we need to complicate the settings by adding some "new_nondraft_pr" setting or something like that.

### DIFF
--- a/src/handlers/autolabel.rs
+++ b/src/handlers/autolabel.rs
@@ -1,3 +1,4 @@
+use crate::db::issue_data::IssueData;
 use crate::{
     config::AutolabelConfig,
     github::{IssuesAction, IssuesEvent, Label},
@@ -5,6 +6,16 @@ use crate::{
 };
 use anyhow::Context as _;
 use tracing as log;
+
+/// Key for the state in the database
+const AUTOLABEL_KEY: &str = "autolabel";
+
+/// State stored in the database
+#[derive(Debug, Default, PartialEq, Clone, serde::Deserialize, serde::Serialize)]
+struct AutolabelState {
+    /// If true, then `autolabel.new_pr` labels have already been applied to this PR.
+    new_pr_labels_applied: bool,
+}
 
 pub(super) struct AutolabelInput {
     add: Vec<Label>,
@@ -26,7 +37,16 @@ pub(super) async fn parse_input(
     // FIXME: This will re-apply labels after a push that the user had tried to
     // remove. Not much can be done about that currently; the before/after on
     // synchronize may be straddling a rebase, which will break diff generation.
-    if event.action == IssuesAction::Opened || event.action == IssuesAction::Synchronize {
+    if matches!(
+        event.action,
+        IssuesAction::Opened | IssuesAction::Synchronize | IssuesAction::ReadyForReview
+    ) {
+        let mut db = ctx.db.get().await;
+        let mut state: IssueData<'_, AutolabelState> =
+            IssueData::load(&mut db, &event.issue, AUTOLABEL_KEY)
+                .await
+                .map_err(|e| e.to_string())?;
+
         let files = event
             .issue
             .diff(&ctx.github)
@@ -69,11 +89,19 @@ pub(super) async fn parse_input(
                         name: label.to_owned(),
                     });
                 }
-                if cfg.new_pr && event.action == IssuesAction::Opened {
-                    autolabels.push(Label {
-                        name: label.to_owned(),
-                    });
-                }
+            }
+
+            // Treat the following situations as a "new PR":
+            // 1) New PRs opened as non-draft
+            // 2) PRs opened as draft that are marked as "ready for review" for the first time.
+            let is_new_non_draft_pr = event.action == IssuesAction::Opened && !event.issue.draft;
+            let is_first_time_ready_for_review =
+                event.action == IssuesAction::ReadyForReview && !state.data.new_pr_labels_applied;
+            if cfg.new_pr && (is_new_non_draft_pr || is_first_time_ready_for_review) {
+                autolabels.push(Label {
+                    name: label.to_owned(),
+                });
+                state.data.new_pr_labels_applied = true;
             }
 
             if event.issue.pull_request.is_none()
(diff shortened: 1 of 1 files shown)


Kobzol:
Actually, this won't work on its own, because when a draft PR is opened, it won't get `S-waiting-on-review`, and once it's converted from draft, it will get assigned someone, but it won't have that label.

Maybe we should modify autolabel to treat both "opened as non draft" and "converted from draft" as triggering `new_pr = true`?

apiraino:
I'm thinking if it makes sense applying *any* kind of post-processing to draft PRs. Is there anything the triagebot needs to do on drafts?

Would it be too radical to set a global `is_draft()` gate on all handlers and just skip everything?

Kobzol:
We might go in that direction, yeah. In theory we could just ignore draft PRs completely, and the first time they are converted from draft to open, we could treat it in triagebot as them opening at that very moment. But that would probably require some larger changes.

Urgau:
> Maybe we should modify autolabel to treat both "opened as non draft" and "converted from draft" as triggering `new_pr = true`?

Make sense to me.

Kobzol:
Ok, I changed the logic, let me know what do you think.

src/handlers/autolabel.rs
```diff
@@ -69,11 +85,18 @@ pub(super) async fn parse_input(
...
-                        name: label.to_owned(),
-                    });
-                }
+            }
+
+            // Newly opened non-draft PRs or PRs that were opened as draft and that were marked as
+            // "ready for review" for the first time are treated as "new PRs".
+            let treat_as_new_pr = (event.action == IssuesAction::Opened && !event.issue.draft)
+                || (event.action == IssuesAction::ReadyForReview
+                    && !state.data.new_pr_labels_applied);
```
apiraino, line 94: The logic here is a bit dense. Just for improving a bit the readability, could it be split into something in this direction:
```suggestion
            // Treat as new PR:
			// - 1) New PRs, exclude drafts
			// - 2) PRs opened as draft and then tagged as "ready for review"
            let is_new_draft_pr = (event.action == IssuesAction::Opened && !event.issue.draft));
            let is_ready_for_review = (event.action == IssuesAction::ReadyForReview
                    && !state.data.new_pr_labels_applied);
			let treat_as_new_pr = !is_new_draft_pr || is_ready_for_review;
```

(unsure about ny rewording, I can't test right now)

src/handlers/autolabel.rs
```diff
@@ -6,6 +7,15 @@ use crate::{
...
 use tracing as log;
 
+/// Key for the state in the database
+const AUTOLABEL_KEY: &str = "autolabel";
+
+/// State stored in the database
+#[derive(Debug, Default, serde::Deserialize, serde::Serialize)]
+struct AutolabelState {
+    new_pr_labels_applied: bool,
+}
```
apiraino, line 17:
```suggestion
#[derive(Debug, Default, serde::Deserialize, serde::Serialize)]
struct AutolabelState {
    has_new_labels: bool,
}
```

?

Kobzol: That sounds a bit confusing/not fully accurate to me, because these labels might not necessarily exist on the PR *right now*, we just remember that we have already tried to apply the `new_pr` labels already. I added a comment to explain this, let me know what you think.

apiraino: > we just remember that we have already tried to apply the new_pr labels already

I see. So in plain English it would be:
> This flag registers if a PR was previously labeled with the predefined set of labels we apply to all new PRs"

Is that correct? Then maybe `already_autolabeled` (or similar)? But I agree, a bit hard to summarize in a short variable name.

(btw, this flag purpose would be more clearer if named `is_new_pr` or similar)

Kobzol: - `new_pr_already_autolabeled`
- `is_new_pr_labels_applied`

Do you like any of these? :laughing: I think that the comment is enough, we probably can't find a completely perfect name for this flag.

apiraino: agreed. let's drop the subject. sorry for wasting some time on this. I hope the comment is enough to a casual reader.

src/handlers/autolabel.rs
```diff
@@ -27,6 +38,12 @@ pub(super) async fn parse_input(
     // remove. Not much can be done about that currently; the before/after on
     // synchronize may be straddling a rebase, which will break diff generation.
     if event.action == IssuesAction::Opened || event.action == IssuesAction::Synchronize {
```
Urgau, line 40: Shouldn't this condition also take into account `IssuesAction::ReadyForReview`? Otherwise the `is_first_time_ready_for_review` below is unreachable, no?

Urgau (approved):

Kobzol:
Seems to work fine: https://github.com/rust-lang/rust/pull/140889.

src/handlers/autolabel.rs
```diff
@@ -69,11 +89,19 @@ pub(super) async fn parse_input(
...
+            // 2) PRs opened as draft that are marked as "ready for review" for the first time.
+            let is_new_non_draft_pr = event.action == IssuesAction::Opened && !event.issue.draft;
+            let is_first_time_ready_for_review =
+                event.action == IssuesAction::ReadyForReview && !state.data.new_pr_labels_applied;
+            if cfg.new_pr && (is_new_non_draft_pr || is_first_time_ready_for_review) {
+                autolabels.push(Label {
+                    name: label.to_owned(),
+                });
+                state.data.new_pr_labels_applied = true;
             }
```
Urgau, line 105: This should have stayed in the condition above, as we are putting new_pr labels on issues instead of only PRs.

I knew this was iffy, but couldn't prove it, thks @jieyouxu.

Urgau: Opened https://github.com/rust-lang/triagebot/pull/1983 to fix it.

PR #1969, opened 2025-05-06
Handle drafts better in PR tracking and do not consider self-assigned PRs for the workqueue

Kobzol:
A few improvements I found after testing the effects of https://github.com/rust-lang/triagebot/pull/1947 in production.

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -298,7 +298,7 @@ async fn workqueue_commands(
                 pluralize("PR", assigned_prs.len())
             );
             writeln!(response, "Review capacity: {capacity}\n")?;
-            writeln!(response, "*Note that only selected PRs that are assigned to you are considered as being in the review queue.*")?;
+            writeln!(response, "*Note that only certain PRs that are assigned to you are included in your review queue.*")?;
             response
         }
         "set-pr-limit" => {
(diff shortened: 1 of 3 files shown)


src/zulip.rs
```diff
@@ -298,7 +298,7 @@ async fn workqueue_commands(
                 pluralize("PR", assigned_prs.len())
             );
             writeln!(response, "Review capacity: {capacity}\n")?;
-            writeln!(response, "*Note that only selected PRs that are assigned to you are considered as being in the review queue.*")?;
+            writeln!(response, "*Note that only some PRs that are assigned to you are considered as being in the review queue.*")?;
```
Urgau, line 301: While we are at it, let's make it clear that we are talking about the user review queue.
```suggestion
            writeln!(response, "*Note that only some PRs that are assigned to you are considered as being in your review queue.*")?;
```
or maybe even this restructured variant,
```suggestion
            writeln!(response, "*Note that only certain PRs assigned to you are included in your review queue.*")?;
```

Urgau (approved):
LGTM, r=me with or without the nit

src/zulip.rs
Kobzol: Reworded, thanks.

PR #1970, opened 2025-05-06
Store vacation status in the database

Kobzol:
This PR adds a notion of being on or off rotation to the DB. This is meant to replace the `on_vacation` array in `triagebot.toml`. The idea is that reviewers will be able to set their vacation/rotation mode through triagebot's DMs (`work set-rotation-mode on/off`), rather than having to modify the `triagebot.toml` file through PRs.

Currently, this PR does not take the DB rotation mode into account when doing reviews, this will be done as a follow-up. Before that, we should resolve if users should be always `r?`-able even if they are on a vacation (https://github.com/rust-lang/triagebot/pull/1920), or if this should be configurable.

I will document this in forge once we get closer to actually using it in practice.

CC @Noratrieb, @apiraino

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -346,4 +346,7 @@ CREATE UNIQUE INDEX IF NOT EXISTS review_prefs_user_id ON review_prefs(user_id);
     "
 ALTER TABLE review_prefs ADD COLUMN IF NOT EXISTS max_assigned_prs INTEGER DEFAULT NULL;
 ",
+    "
+ALTER TABLE review_prefs ADD COLUMN IF NOT EXISTS rotation_mode TEXT NOT NULL DEFAULT 'on-rotation';
+",
 ];
--- a/src/db/review_prefs.rs
+++ b/src/db/review_prefs.rs
@@ -1,13 +1,65 @@
 use crate::db::users::record_username;
 use crate::github::{User, UserId};
 use anyhow::Context;
+use bytes::BytesMut;
+use postgres_types::{to_sql_checked, FromSql, IsNull, ToSql, Type};
 use std::collections::HashMap;
+use std::error::Error;
+
+#[derive(Debug, Default, Copy, Clone)]
+pub enum RotationMode {
+    /// The reviewer can be automatically assigned by triagebot,
+    /// and they can be assigned through teams and assign groups.
+    #[default]
+    OnRotation,
+    /// The user is off rotation (e.g. on a vacation) and cannot be assigned automatically
+    /// nor through teams and assign groups.
+    OffRotation,
+}
+
+impl<'a> FromSql<'a> for RotationMode {
+    fn from_sql(ty: &Type, raw: &'a [u8]) -> Result<Self, Box<dyn Error + Sync + Send>> {
+        let value = <&str as FromSql>::from_sql(ty, raw)?;
+        match value {
+            "on-rotation" => Ok(Self::OnRotation),
+            "off-rotation" => Ok(Self::OffRotation),
+            _ => Err(format!("Unknown value for RotationMode: {value}").into()),
+        }
+    }
+
+    fn accepts(ty: &Type) -> bool {
+        <&str as FromSql>::accepts(ty)
+    }
+}
+
+impl ToSql for RotationMode {
+    fn to_sql(&self, ty: &Type, out: &mut BytesMut) -> Result<IsNull, Box<dyn Error + Sync + Send>>
+    where
+        Self: Sized,
+    {
+        let value = match self {
+            RotationMode::OnRotation => "on-rotation",
+            RotationMode::OffRotation => "off-rotation",
+        };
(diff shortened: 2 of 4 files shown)


src/db.rs
```diff
@@ -345,5 +345,8 @@ CREATE UNIQUE INDEX IF NOT EXISTS review_prefs_user_id ON review_prefs(user_id);
  ",
     "
 ALTER TABLE review_prefs ADD COLUMN IF NOT EXISTS max_assigned_prs INTEGER DEFAULT NULL;
+",
+    "
+ALTER TABLE review_prefs ADD COLUMN IF NOT EXISTS rotation_mode TEXT NOT NULL DEFAULT 'on-rotation';
```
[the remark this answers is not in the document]
Kobzol: It's definitely an option, I just wanted to stay more "forward-compatible" with the design. For example, I was thinking that in the future, we could extend the mode to something like this:
```rust
enum RotationMode {
   OnRotation, // allow r? <user> + r? <team> + autoassign
   AcceptsDirectRequests, // allow only r? <user>
   OnVacation // don't allow anything, I want to REST!
}
```
Lately, when I added new booleans to the DB in some other projects, I then sometimes realized that the situation is more complicated and we had to migrate to strings.

Since this won't be a lot of data, and perf. wise it shouldn't really matter, I'd rather default to something a bit more general, to make it easier to be extended in the future.

src/db/review_prefs.rs
```diff
@@ -88,28 +145,33 @@ pub async fn upsert_review_prefs(
...
+            query,
+            &[&(user.id as i64), &max_assigned_prs, &rotation_mode],
+        )
         .await
         .context("Error upserting review preferences")?;
     Ok(res)
 }
 
 #[cfg(test)]
 mod tests {
```
Urgau, line 173: Unless I'm missing something, we are never checking that the rotation mode is the correct. Could you add test for it.

src/zulip.rs
```diff
@@ -270,6 +274,15 @@ async fn workqueue_commands(
...
     let db_client = ctx.db.get().await;
 
+    let gh_username = username_from_gh_id(&ctx.github, gh_id)
+        .await?
+        .ok_or_else(|| anyhow::anyhow!("Cannot find your GitHub username in the team database"))?;
+    let user = User {
+        login: gh_username.clone(),
+        id: gh_id,
+    };
+    let review_prefs = get_review_prefs(&db_client, gh_id).await?;
```
Urgau, line 284:
```suggestion
    let review_prefs = get_review_prefs(&db_client, gh_id).await.context("Unable to retrieve your review preferences.")?;
```

src/zulip.rs
```diff
@@ -339,9 +356,32 @@ async fn workqueue_commands(
...
+                review_prefs.and_then(|p| p.max_assigned_prs.map(|v| v as u32)),
+                rotation_mode,
+            )
+            .await
+            .context("Error occurred while setting review preferences.")?;
+            tracing::info!("Setting rotation mode `{gh_username}` to {rotation_mode:?}");
+            format!("Rotation mode set to {rotation_mode:?}")
+        }
+        "help" => WORKQUEUE_HELP.to_string(),
         _ => anyhow::bail!("Invalid subcommand."),
```
Urgau, line 385: nit: we could/should mention the `help` command when we encounter an invalid subcommand here

Kobzol: That's already done on the higher level, if anything fails here, help is automatically printed.

src/db/review_prefs.rs
Kobzol: I will definitely add tests, good point. Not sure what do you mean by it being correct, though? It's an enum, it can only be in two states and both are valid.

Urgau: I meant that we should make sure that the updated rotation mode is the one we expect, nothing more.

src/zulip.rs
Urgau: Great.

src/zulip.rs
```diff
@@ -339,9 +358,32 @@ async fn workqueue_commands(
...
+            upsert_review_prefs(
+                &db_client,
+                user,
+                review_prefs.and_then(|p| p.max_assigned_prs.map(|v| v as u32)),
+                rotation_mode,
+            )
+            .await
+            .context("Error occurred while setting review preferences.")?;
+            tracing::info!("Setting rotation mode `{gh_username}` to {rotation_mode:?}");
+            format!("Rotation mode set to {rotation_mode:?}")
```
Urgau, line 384: Seems a bit inconsistent to use the debug representation here while we use a "display" print in the `show` command.

```suggestion
            format!("Rotation mode set to {}", match rotation_mode {
                RotationMode::OnRotation => "on rotation",
                RotationMode::OffRotation => "off rotation",
            })
```

Urgau (approved):

PR #1971, opened 2025-05-06
Support test DB host on a different port

apiraino:
The canonical `DATABASE_URL` string is usually in the form of:

`postgres://username:password@host:port/dbname`


Kobzol (approved):
Thanks! Confirmed that it also works without a port.

apiraino:
only if your DB in on port 5432.

Kobzol:
Yeah, but if it isn't, it wouldn't work anyway (the Postgres code uses `5432` as the default port), so that's fine :)

PR #1972, opened 2025-05-06
Determine whether reviewers were requested directly or not

Kobzol:
This will be useful for using a different assignment logic for direct `r?` requests and automatic/team/group rotation assignments (https://github.com/rust-lang/triagebot/pull/1970).

This PR should contain no functional changes.

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -799,32 +799,60 @@ async fn find_reviewer_from_names(
         .to_string())
 }
 
+#[derive(Eq, PartialEq, Hash)]
+struct ReviewerCandidate {
+    name: String,
+    origin: ReviewerCandidateOrigin,
+}
+
+#[derive(Eq, PartialEq, Hash, Copy, Clone)]
+enum ReviewerCandidateOrigin {
+    /// This reviewer was directly requested for a review.
+    Direct,
+    /// This reviewer was expanded from a team or an assign group.
+    Expanded,
+}
+
 /// Recursively expand all teams and adhoc groups found within `names`.
 /// Returns a set of expanded usernames.
 /// Also normalizes usernames from `@user` to `user`.
-///
-/// Returns `(set of expanded users, expansion_happened)`.
-/// `expansion_happened` signals if any expansion has been performed.
 fn expand_teams_and_groups(
     teams: &Teams,
     issue: &Issue,
     config: &AssignConfig,
     names: &[String],
-) -> Result<(HashSet<String>, bool), FindReviewerError> {
-    let mut expanded = HashSet::new();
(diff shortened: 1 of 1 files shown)


src/handlers/assign.rs
```diff
@@ -799,44 +799,86 @@ async fn find_reviewer_from_names(
...
     let mut seen_names = HashSet::new();
 
+    enum Candidate<'a> {
+        Direct(&'a str),
+        Expanded(&'a str),
+    }
+
     // This is a queue of potential groups or usernames to expand. The loop
     // below will pop from this and then append the expanded results of teams.
     // Usernames will be added to `expanded`.
```
Urgau, line 850: Outdated comment, `expanded` not longer exists.

src/handlers/assign.rs
```diff
@@ -799,44 +799,86 @@ async fn find_reviewer_from_names(
         .to_string())
 }
 
+#[derive(Eq, PartialEq, Hash)]
+enum ReviewerCandidate {
+    /// This reviewer was directly requested for a review.
+    Direct(String),
+    /// This reviewer was expanded from a team or an assign group.
+    Expanded(String),
+}
```
Urgau, line 808: I find this enum to be highly confusing, the origin of the candidate reviewer is a property of the candidate not a variants of the candidate.

I would expect something like this instead:
```rust
struct ReviewerCandidate {
    name: String,
    origin: ReviewerCandidateOrigin,
}

#[derive(Eq, PartialEq, Hash)]
enum ReviewerCandidateOrigin {
    /// This reviewer was directly requested for a review.
    Direct,
    /// This reviewer was expanded from a team or an assign group.
    Expanded,
}
```

src/handlers/assign.rs
```diff
@@ -799,44 +799,86 @@ async fn find_reviewer_from_names(
...
+    let mut selected_candidates: HashSet<String> = HashSet::new();
 
     // Keep track of groups seen to avoid cycles and avoid expanding the same
     // team multiple times.
     let mut seen_names = HashSet::new();
 
+    enum Candidate<'a> {
+        Direct(&'a str),
+        Expanded(&'a str),
+    }
```
Urgau, line 846: Can we re-use `ReviewerCandidate`? Maybe by having a generic `S = String` that would be `&str` here and stay the default elsewhere.

Urgau: Sure, no problem.

src/handlers/assign.rs
Urgau: Maybe my brain is weird. :smile: Thanks.

Urgau (approved):

PR #1974, opened 2025-05-06
notify_zulip: handle labels set when opening the pr

### LINKED ISSUE #1973: `notify-zulip.<LABEL>` should also handle `IssuesAction::Open` and treat it similar to `IssuesAction::Labeled`
Context: https://github.com/rust-lang/rust/pull/140397#discussion_r2072316313

Example: Someone adds various labels to a to-be-submitted PR and then opens it. Then, the `notify_zulip` handler should trigger all registered actions matching the added labels as if they were added after the PR submission. Right now, triagebot simply ignores them and one needs to remove and readd the relevant labels to trigger the actions.

Such a situation happened here: https://github.com/rust-lang/rust/pull/137899. Of course, fixing issue #1184 would also solve this specific case, in a different way.

However, the general issue affects all notify Zulip actions, not just `{beta,stable}-nominated` (e.g., `I-types-nominated` comes to mind), so it's worth fixing regardless (and it'd be a lot easier to fix than #1184).

cc @apiraino

Urgau:
Handle labels set when opening the PR as if they were added in [`notify_zulip` handler](https://forge.rust-lang.org/triagebot/zulip-notifications.html).

We re-use `messages_on_add`, since it's the conceptually same thing as if they were manually added just after the PR was opened.

cc @apiraino @fmease 
Fixes #1973

### DIFF
--- a/src/handlers/notify_zulip.rs
+++ b/src/handlers/notify_zulip.rs
@@ -18,6 +18,7 @@ pub(super) struct NotifyZulipInput {
 }
 
 pub(super) enum NotificationType {
+    Open,
     Labeled,
     Unlabeled,
     Closed,
@@ -45,8 +46,8 @@ pub(super) async fn parse_input(
                 })
                 .map(|input| vec![input]))
         }
-        IssuesAction::Closed | IssuesAction::Reopened => {
-            Ok(Some(parse_close_reopen_input(event, config)))
+        IssuesAction::Opened | IssuesAction::Closed | IssuesAction::Reopened => {
+            Ok(Some(parse_open_close_reopen_input(event, config)))
         }
         _ => Ok(None),
     }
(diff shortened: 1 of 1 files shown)


Kobzol (approved):

src/handlers/notify_zulip.rs
```diff
@@ -45,8 +46,8 @@ pub(super) async fn parse_input(
                 })
                 .map(|input| vec![input]))
         }
-        IssuesAction::Closed | IssuesAction::Reopened => {
-            Ok(Some(parse_close_reopen_input(event, config)))
+        IssuesAction::Opened | IssuesAction::Closed | IssuesAction::Reopened => {
+            Ok(Some(parse_open_close_reopen_input(event, config)))
```
apiraino, line 50: I was looking for a slightly more parseable (ha!) name for this function. `parse_input()`? `get_notification_input()`? Suggestions?

Urgau: We already have a `parse_label_change_input` for label and unlabel events, I rather not make less distinctive.

Better have a descriptive function name than a short one that is confusing.

PR #1975, opened 2025-05-07
Do not write `IssueData` to the DB when it is unchanged

Kobzol:
Noticed this while working on https://github.com/rust-lang/triagebot/pull/1968. I don't think that it's really needed for perf., but I suppose that most of the time we just write back the same data, which seems wasteful, and the structs are simple enough that requiring `Clone` + `Eq` shouldn't be that difficult.

### DIFF
--- a/src/db/issue_data.rs
+++ b/src/db/issue_data.rs
@@ -16,18 +16,19 @@ use tokio_postgres::{Client as DbClient, Transaction};
 
 pub struct IssueData<'db, T>
 where
-    T: for<'a> Deserialize<'a> + Serialize + Default + std::fmt::Debug + Sync,
+    T: for<'a> Deserialize<'a> + Serialize + Default + std::fmt::Debug + Sync + PartialEq + Clone,
 {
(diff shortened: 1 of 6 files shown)


src/db/issue_data.rs
```diff
@@ -51,25 +52,33 @@ where
...
                  ON CONFLICT (repo, issue_number, key) DO UPDATE SET data=EXCLUDED.data",
-                &[&self.repo, &self.issue_number, &self.key, &Json(&self.data)],
-            )
-            .await
-            .context("inserting issue data")?;
+                    &[&self.repo, &self.issue_number, &self.key, &Json(&self.data)],
+                )
+                .await
+                .context("inserting issue data")?;
+            return Ok(());
```
fbstj, line 80: isn't this skipping the below commit?

Kobzol: Oops, that was not supposed to be there, a refactoring gone wrong. Thanks!

Urgau (approved):
Sure, why not.

PR #1979, opened 2025-05-09
Fix build on Windows

Alexendoo:
The direct `openssl` dep failed to compile out of the box, it's replaced with `hmac` & friends

`mio` and `schannel` have been `cargo update`d to drop `ntapi` that failed to compile and fix a runtime crash

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,6 @@ rust-version = "1.80.0"
 
 [dependencies]
 serde_json = "1"
-openssl = "0.10"
 dotenvy = "0.15"
 reqwest = { version = "0.11.4", features = ["json", "blocking"] }
 regex = "1"
@@ -46,6 +45,10 @@ postgres-types = { version = "0.2.4", features = ["derive"] }
 cron = { version = "0.15.0" }
 bytes = "1.1.0"
 structopt = "0.3.26"
+hmac = "0.12.1"
+sha1 = "0.10.6"
+digest = "0.10.7"
+subtle = "2.6.1"
 
 [dependencies.serde]
 version = "1"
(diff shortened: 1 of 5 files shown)


Urgau (approved):

src/payload.rs
```diff
@@ -22,18 +23,12 @@ pub fn assert_signed(signature: &str, payload: &[u8]) -> Result<(), SignedPayloa
         }
     };
 
-    let key = PKey::hmac(
+    let mut mac = Hmac::<Sha1>::new_from_slice(
```
Kobzol, line 26: While we're modifying this, could you please switch to Sha256, and examine the `X-Hub-Signature-256` header rather than `X-Hub-Signature`? The old header is [deprecated](https://docs.github.com/en/webhooks/using-webhooks/validating-webhook-deliveries#troubleshooting). But if you don't want to do it, it's fine, I can do it in a separate PR.

Cargo.toml
```diff
@@ -46,6 +45,10 @@ postgres-types = { version = "0.2.4", features = ["derive"] }
 cron = { version = "0.15.0" }
 bytes = "1.1.0"
 structopt = "0.3.26"
+hmac = "0.12.1"
+sha1 = "0.10.6"
+digest = "0.10.7"
```
Kobzol, line 50: I think that this dependency is unused?
```suggestion

```

Urgau: I had the same reflection and I think it's used as a dependency of `hmac`. It's probably here to set it's version transitively.

Alexendoo: Ah yes it's unused, initially I was trying to use it for https://docs.rs/digest/latest/digest/struct.CtOutput.html#method.ct_eq before I realised it's from the `subtle` crate

Kobzol (approved):
Seems to be working fine, thanks!

PR #1981, opened 2025-05-09
Handle escaped quotes and remove outer quotes in Summary notes

### LINKED ISSUE #1980: Note functionality does not implement string escaping
(Very low-priority enhancement request)

The `@bot note "<string_content>"` form has no way for the user to escape the double-quotes, i.e. `@bot note "hello \"world\""` does not "escape" the double-quotes.

It'd be nice if there was a way to also preserve the escaped double-quotes in the "Summary notes" view, but this is really not a huge deal.

Urgau:
This PR handles escaped quotes and removes the outer quotes in the "Summary notes" section of the note handler/feature.

cc @jieyouxu
Fixes #1980


Kobzol:
Can the note be multi-line? If not, could we just take whatever is after the `note` keyword as being the note, without requiring quotes? That seems like it could avoid the escaping logic.

Also, we could format the note like this, to make the part that forms the note stand out more, now that there are no outer level quotes in the summary list. The username could be rendered without `@` to avoid spamming people (but it could be a link to `https://github.com/<username>` to still allow clicking on it).

### Summary Notes

- [Note with "quotes" in the title](https://github.com/rust-lang/rust/issues/133469#issuecomment-2865978181) by @Urgau

Kobzol:
Oh, actually I see that quotes were not required after `note`. Then why do we need any escaping of quotes at all? :thinking: I thought that only the second commit would be required, why the first one?

Urgau:
It's only for a word, no spaces.

Kobzol:
Ok so maybe I'll ask in a different way, just to make sure I'm not missing anything :grinning: Could we change the behavior of `@bot note` so that it just takes whatever is on the line after `note` and uses that as the summary title? And not add (nor escape any quotes?

Urgau:
It personally makes sense to me to remove them by title, that's how it's done with `@rfcbot resolve` (for concerns).

:+1: for matching on prefix.

Kobzol:
So, to summarize:
- Remove notes by prefix
- Do not escape anything, just take all tokens before `.`
- Do not wrap notes in quotes, and only wrap the note text with a link to better distinguish it from the rest of the line (`by <author>`

?

Urgau:
> Do not escape anything, just take all tokens before `.`

Not sure about that, this seems hacky at best, what if we have `@bot note "This a  summary. title.".` ?

Kobzol:
I mean, I don't expect people to create such notes :laughing: But I don't really see an issue with that? The note would just be named `"This a  summary.`. Is that a problem?

Urgau:
The dot wouldn't be included but I'm also fine with it.

Kobzol:
I would rather just not support anything too complex here to avoid complicated parsing/escaping logic, and just do the simplest thing that makes sense. I haven't seen notes being used much, and I don't think that supporting quotes in some special way is really needed (unless someone comes with a compelling use-case :) ). In most cases the note will likely be just a few words without quotes anyway.

PR #1985, opened 2025-05-10
Add a handler to post a summary comment of Clippy lintcheck runs

Alexendoo:
In Clippy we have a tool called lintcheck that we use in CI as a sort of mini crater run, this handler would post a summary of the run as a comment in the PR thread

It works by having the workflow upload a JSON artifact that the handler downloads, in Clippy we would add

```toml
[lintcheck-summary]
workflow = "Lintcheck"
artifact = "summary"
```

JSON is used instead of uploading a markdown file so that PRs from forks can't have triagebot post arbitrary comments

For https://github.com/rust-lang/rust-clippy/actions/runs/14763277413#summary-41449477485 the following comment would be posted:

---

Lintcheck changes for f980edd0da51eb4da9c15eb423a7829f2c5657fc

| Lint | Added | Removed | Changed |
| ---- | ----: | ------: | ------: |
| [`clippy::collapsible_if`](https://github.com/rust-lang/rust-clippy/actions/runs/14763277413#user-content-collapsible-if) | 0 | 0 | 1 |
| [`clippy::elidable_lifetime_names`](https://github.com/rust-lang/rust-clippy/actions/runs/14763277413#user-content-elidable-lifetime-names) | 30 | 0 | 1125 |
| [`clippy::extra_unused_lifetimes`](https://github.com/rust-lang/rust-clippy/actions/runs/14763277413#user-content-extra-unused-lifetimes) | 1 | 3 | 4 |
| [`clippy::needless_lifetimes`](https://github.com/rust-lang/rust-clippy/actions/runs/14763277413#user-content-needless-lifetimes) | 23 | 0 | 214 |

This comment will be updated if you push new changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,6 +24,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "adler2"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+
+[[package]]
 name = "ahash"
 version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -80,6 +86,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a4668cab20f66d8d020e1fbc0ebe47217433c1b6c8f2040faf858554e394ace6"
 
 [[package]]
+name = "arbitrary"
+version = "1.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dde20b3d026af13f561bdd0f15edf01fc734f0dafcedbaf42bba506a9517f223"
+dependencies = [
+ "derive_arbitrary",
+]
+
+[[package]]
 name = "arc-swap"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -129,7 +144,7 @@ dependencies = [
  "cc",
  "cfg-if",
  "libc",
- "miniz_oxide",
+ "miniz_oxide 0.4.4",
  "object",
  "rustc-demangle",
 ]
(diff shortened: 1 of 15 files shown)


Alexendoo:
A PR workflow doesn't have permission to create comments when it's from a fork but it is possible with a separate `on.workflow_run` workflow, I had a non JSON implementation working [here](https://github.com/Alexendoo/rust-clippy/blob/0faffaf1cf8003ea33f145469b0cd65eb9bea5f4/.github/workflows/lintcheck_pr_comment.yml)

That's when I thought it might not be a good idea to allow comments of arbitrary content, but I did not fancy writing a bunch of JS in YAML

Urgau:
Interesting, the workflow approach seems much more approachable.

So if I understand correctly the motivating factor for doing this in triagebot is over security, but wouldn't adversarial PR still be able to put arbitrary data in some of the output (like name and what not)?

This reminds me that in rust-lang/rust we have had for a few weeks something similar (cc @Kobzol). We have a [post-merge workflow](https://github.com/rust-lang/rust/blob/b10555674f355aca2bfe974e50a0b9ab48eb1d87/.github/workflows/post-merge.yml) that outputs details about the workflow; granted it doesn't download artifact and runs post merge (so much less security risks).

Kobzol:
I know that posting comments from actions on PRs from forks on GitHub is a PITA, but I would like to avoid (ab)using triagebot for this, as this is a very repo-specific functionality and I think it belongs in Clippy, not in triagebot.

The way I do this in the [team](https://github.com/rust-lang/team) is to run a normal `pull_request` workflow that computes the necessary data, and then [uploads it as an artifact](https://github.com/rust-lang/team/blob/master/.github/workflows/main.yml#L67) that contains the PR number and the gathered data. Then there is a second workflow that runs from the main branch and has the `workflow_run` trigger (https://github.com/rust-lang/team/blob/master/.github/workflows/dry-run.yml), which downloads the results and sends the comment. The [bash code](https://github.com/rust-lang/team/blob/master/.github/workflows/dry-run.yml#L148) for that isn't so terrible, but if you want to avoid it, you can also just create e.g. a Python script in the Clippy repo and run it in the workflow. If you upload a JSON and then read some known data out of it, then I think it should be fine security-wise.

Alexendoo:
> wouldn't adversarial PR still be able to put arbitrary data in some of the output (like name and what not)?

Yeah any strings in the JSON are checked to only contain safe chars

> The [bash code](https://github.com/rust-lang/team/blob/master/.github/workflows/dry-run.yml#L148) for that isn't so terrible

True, I was picturing swapping the whole thing over to JS but really I only need to parse the JSON + write a file with it, the rest can still be done via `gh` commands. That should be pretty reasonable

FWIW I found that the PR number is available [in the event](https://github.com/Alexendoo/rust-clippy/blob/0faffaf1cf8003ea33f145469b0cd65eb9bea5f4/.github/workflows/lintcheck_pr_comment.yml#L46), and `gh` has gained the ability to do `--edit-last --create-if-none`

Kobzol:
Oh, nice, good to know about that flag!

Watch out about the PR number though, I also tried to use it from that attribute, but it won't be filled on PRs from forks :facepalm:

Issue #1986, opened 2025-05-14
Allow version numbers to be used in note summaries


samueltardieu:
Feature request: I'd like to be able to use version numbers in note summaries (for example, a note whose summary is "revisit if merged after Rust 1.90 reaches beta"). Right now, the note summary parser will stop at the punctuation mark and does not support quotes anymore.

Cc @Urgau

samueltardieu:
That would be my preferred solution as well.

PR #1987, opened 2025-05-14
Take everything until new-line as the summary in the note command

Urgau:
Fixes https://github.com/rust-lang/triagebot/issues/1986
cc @samueltardieu


Kobzol (approved):
Thanks!

PR #1989, opened 2025-05-15
Remove possibly redundant Zulip env var

apiraino:
I am not sure why we use two env vars to talk to Zulip (`ZULIP_TOKEN` and `ZULIP_API_TOKEN`, see [simpleinfra](https://github.com/rust-lang/simpleinfra/blob/b7ddd5e6bec8a93ec30510cdddec02c5666fefe9/terraform/shared/services/triagebot/main.tf#L95-L101)).
But I think there was a reason why it was added to the infra in commit https://github.com/rust-lang/simpleinfra/commit/d13e1551

I was wondering if one can go, it's slightly confusing.
Anyone in T-infra remembers? cc @Mark-Simulacrum 

If there is a reason for that, then I'll push a small patch to the docs of the triagebot because the latter is not mentioned

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -429,7 +429,7 @@ async fn execute_for_other_user(
         assert_eq!(command.pop(), Some(' ')); // pop trailing space
         command
     };
-    let bot_api_token = env::var("ZULIP_API_TOKEN").expect("ZULIP_API_TOKEN");
+    let bot_api_token = env::var("ZULIP_TOKEN").expect("ZULIP_TOKEN");
 
     let members = ctx
         .github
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
Per existing docs - https://github.com/rust-lang/triagebot/blob/19a277feef088f7ae96f80200266b3d2e8b4f157/.env.sample#L23 - the two environment variables are pretty different. Maybe we can rename ZULIP_TOKEN to ZULIP_WEBHOOK_SECRET?

apiraino:
> Per existing docs -
> 
> https://github.com/rust-lang/triagebot/blob/19a277feef088f7ae96f80200266b3d2e8b4f157/.env.sample#L23
> - the two environment variables are pretty different. Maybe we can rename ZULIP_TOKEN to ZULIP_WEBHOOK_SECRET?

and I even wrote that comment :man_facepalming: 

yeah maybe renaming it is a good idea. I'll take note to do it.

thanks

PR #1991, opened 2025-05-16
Rename ZULIP_TOKEN env var to ZULIP_WEBHOOK_SECRET

apiraino:
From comment https://github.com/rust-lang/triagebot/pull/1989#issuecomment-2884830044

This patch makes it so, that we now accept `ZULIP_WEBHOOK_SECRET` for the token to send messages to Zulip, the name is slightly clearer. Keeping it compatible with `ZULIP_TOKEN` to allow time to change it in the infra.

r? @Mark-Simulacrum did you mean something like this patch?

thanks

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -20,7 +20,7 @@ GITHUB_WEBHOOK_SECRET=MUST_BE_CONFIGURED
 # ZULIP_API_TOKEN=yyy
 
 # Authenticates inbound webhooks from Github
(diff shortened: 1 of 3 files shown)


Kobzol:
Oh, sorry, I missed the mention of it being backwards compatible :man_facepalming: In that case I think this should be safe to merge. I will take a look tomorrow.

Issue #1996, opened 2025-05-19
Labels with space cannot be used with `label` command


samueltardieu:
Context: in the Clippy repository, we recently renamed the `good-first-issue` label into `good first issue`, for consistency with most of other Rust project repositories.

According to https://forge.rust-lang.org/triagebot/labeling.html a `label` used in a `label-delta` part of a `label` command can contain spaces. However, a `@rustbot label +good first issue` [complained](https://github.com/rust-lang/rust-clippy/issues/14847#issuecomment-2890661301) that the `good`, `first`, and `issue` labels do not exist.

Is there another way to get rustbot to label issues with labels containing spaces? And should the documentation get fixed?

Urgau:
The documentation is wrong, there is no way to provide a *space* to a word, the only way I can see this works is by supporting quoted strings (ie. `@bot labels "good first issue"`).

Opened #1997 to address that.

ehuss:
> According to https://forge.rust-lang.org/triagebot/labeling.html a `label` used in a `label-delta` part of a `label` command can contain spaces.

That documentation is explicitly saying it does *not* support a space. The `^` is a common regex form which means "any character except". Spaces are used to separate multiple labels, which does look like the grammar is missing (though the text above it does include).

samueltardieu:
Indeed, I read too fast! Quotes would be great, as:
- "good first issue" is best than "good-first-issue" for discovery, as it is commonly used through the project
- Our guide advices to use triagebot to set the label so that whoever sets the label gets automatically subscribed and can more easily assist beginners.

PR #1997, opened 2025-05-19
Handle quoted labels in relabel commands

Urgau:
Fixes https://github.com/rust-lang/triagebot/issues/1996

### DIFF
--- a/parser/src/command/relabel.rs
+++ b/parser/src/command/relabel.rs
@@ -24,7 +24,9 @@
 //!  is with the previous two variants of this (i.e., ++label and -+label).
 //!  - <label>
 //!
-//! <label>: \S+
+//! <label>:
+//!  - \S+
+//!  - ".+"
 //! ```
 
 use crate::error::Error;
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks!

Can you also post a PR to update the documentation at https://github.com/rust-lang/rust-forge/blob/master/src/triagebot/labeling.md?

At the same time, can you also fix the grammar so that it accepts space separated deltas? There's multiple ways to do that, and I'm not sure which one would be clearest, but the following is one approach:

```diff
diff --git a/src/triagebot/labeling.md b/src/triagebot/labeling.md
index c76d063..278c10b 100644
--- a/src/triagebot/labeling.md
+++ b/src/triagebot/labeling.md
@@ -35,7 +35,7 @@ Formally the grammar is:
 > &nbsp;&nbsp; | `labels`
 >
 > label-list →\
-> &nbsp;&nbsp; &nbsp;&nbsp; *label-delta*\
+> &nbsp;&nbsp; &nbsp;&nbsp; *label-delta*<sup>+</sup>\
 > &nbsp;&nbsp; | *label-delta* `and` *label-list*\
 > &nbsp;&nbsp; | *label-delta* `,` *label-list*\
 > &nbsp;&nbsp; | *label-delta* `,` `and` *label-list*
```

Issue #1992, opened 2025-05-17
rust-lang/rust triagebot ping group might be broken?


jieyouxu:
E.g. https://github.com/rust-lang/rust/pull/140966#issuecomment-2886704667, seems like `@rustbot ping rfl` is not working (at least sometimes?).

jieyouxu:
Or maybe it's just the `rfl` `alias`? The actual `marker-team` in `team` is called `rust-for-linux`.

ehuss:
The label for that ping does not exist:

`Unknown labels: O-rfl`

Perhaps it should be A-rust-for-linux?

jieyouxu:
You're right -- both the labels referenced are outdated in `triagebot.toml` and also in rustc-dev-guide docs. https://github.com/rust-lang/rust/pull/141139 should fix both and thus close this issue.

I do wonder if it might make sense for triagebot to loudly complain if a ping group becomes misconfigured? Idk, it's hard to figure out why it's not working otherwise.

ehuss:
There is very basic validation done in https://github.com/rust-lang/triagebot//blob/HEAD/src/handlers/validate_config.rs. It sounds reasonable to me to continue extending that to check things like "does label exist". I think it should be fine to add that in terms of API points and time budget as long as we don't go overboard.

It also seems reasonable to me to have the ping handler just log errors about missing labels, but otherwise continue to process the ping (instead of just stopping).

Those should be relatively easy changes if anyone is interested in trying.

ZuseZ4:
Relatedly, @jieyouxu (only on my phone ATM)
https://github.com/rust-lang/blog.rust-lang.org/pull/1810#issuecomment-4344853030

PR #1998, opened 2025-05-19
Post comment when failing to add team label in `ping` command

Urgau:
> It also seems reasonable to me to have the ping handler just log errors about missing labels, but otherwise continue to process the ping (instead of just stopping).

From https://github.com/rust-lang/triagebot/issues/1992#issuecomment-2888608180


ehuss (approved):
Thanks!

Issue #1990, opened 2025-05-15
We should handle API errors more consistently


Urgau:
We should handle API errors more consistently

_Originally posted by @Urgau in https://github.com/rust-lang/triagebot/issues/1988#issuecomment-2884874684_

PR #2001, opened 2025-05-20
Consistently handle Zulip API errors

Urgau:
This PR makes it so that we consistently handle Zulip API errors (at all for some of them).

Fixes https://github.com/rust-lang/triagebot/issues/1990
cc @apiraino

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -475,18 +475,8 @@ async fn execute_for_other_user(
     .send(ctx.github.raw())
     .await;
 
-    match res {
-        Ok(resp) => {
-            if !resp.status().is_success() {
-                log::error!(
-                    "Failed to notify real user about command: response: {:?}",
-                    resp
-                );
-            }
-        }
-        Err(err) => {
-            log::error!("Failed to notify real user about command: {:?}", err);
-        }
+    if let Err(err) = res {
+        log::error!("Failed to notify real user about command: {:?}", err);
     }
 
     Ok(Some(output))
@@ -591,7 +581,7 @@ impl<'a> MessageApiRequest<'a> {
         self.recipient.url()
     }
 
-    pub async fn send(&self, client: &reqwest::Client) -> anyhow::Result<reqwest::Response> {
+    pub async fn send(&self, client: &reqwest::Client) -> anyhow::Result<MessageApiResponse> {
         let bot_api_token = env::var("ZULIP_API_TOKEN").expect("ZULIP_API_TOKEN");
 
         #[derive(serde::Serialize)]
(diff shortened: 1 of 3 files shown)


src/zulip.rs
```diff
@@ -591,7 +581,8 @@ impl<'a> MessageApiRequest<'a> {
         self.recipient.url()
     }
 
-    pub async fn send(&self, client: &reqwest::Client) -> anyhow::Result<reqwest::Response> {
+    #[must_use]
+    pub async fn send(&self, client: &reqwest::Client) -> anyhow::Result<MessageApiResponse> {
```
Kobzol, line 585: Isn't Future/Result `#[must_use]` by default?

src/zulip.rs
```diff
@@ -850,18 +850,32 @@ struct AddReaction<'a> {
 }
 
 impl<'a> AddReaction<'a> {
-    pub async fn send(self, client: &reqwest::Client) -> anyhow::Result<reqwest::Response> {
+    #[must_use]
```
Kobzol, line 868: Same here.

src/zulip.rs
```diff
@@ -652,7 +652,8 @@ pub struct UpdateMessageApiRequest<'a> {
 }
 
 impl<'a> UpdateMessageApiRequest<'a> {
-    pub async fn send(&self, client: &reqwest::Client) -> anyhow::Result<reqwest::Response> {
+    #[must_use]
```
Kobzol, line 655: And here.

src/zulip.rs
```diff
@@ -677,7 +678,21 @@ impl<'a> UpdateMessageApiRequest<'a> {
...
+            .await
+            .context("failed to send Zulip API Update Message")?;
+
+        let status = resp.status();
+
+        if !status.is_success() {
+            let body = resp
+                .text()
+                .await
+                .context("fail receiving Zulip API response (when adding a reaction)")?;
```
Kobzol, line 690: Wrong error message.

Kobzol (commented):
It looks like the Zulip API code contains a lot of duplication that would eventually be nice to unify, but this is already a nice improvement over the status quo. Thanks!

src/zulip.rs
Urgau: Fixed the error message.

src/zulip.rs
Urgau: Yes, it is `#[must_use]`, but I though that if I added it it would also require usage of the inner T, but that doesn't seems to be the case. Removed them.

Kobzol: No that's still a WIP feature/idea, IIRC :)

Kobzol (approved):

PR #2006, opened 2025-05-22
Remove our custom diff parsing and prefer Github compare json API

Urgau:
Follow-up to #2005, only the last commit is relevant


Urgau:
> * Some commits in this PR modify **submodules**.

This is really weird, we don't have any submodules in this repository. For someone reason my commit probably has *two* parents, maybe the rebase was weird, idk

btw, this is read for review @Kobzol

Urgau:
Oh, this is funny, our modified submodules handler is looking at the PR diff for "Subproject commit" in it and warns if it finds one, which of course there are, bc I'm modifying the handler tests. :smile: 

https://github.com/rust-lang/triagebot/blob/c335263d86c7d6fe13c0e280e3ab3d7d5b509b4a/src/handlers/check_commits/modified_submodule.rs#L7-L8

Kobzol (approved):
Awesome, always nice to see complex code getting removed!

PR #2014, opened 2025-05-26
Always allow direct review requests

Kobzol:
This PR modifies the assignment logic so that it is always possible to `r?` users, even if they are on a vacation or if they are at their capacity. When a maximum review capacity or vacation is ignored due to using a direct review request, triagebot will post a warning, telling the requested that the reviewer might take a while to respond.

In the future, we can make this behavior configurable.
Requested in https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Review.20queue.20tracking.20in.20triagebot/with/520106062.

Supersedes https://github.com/rust-lang/triagebot/pull/1920 (CC @Noratrieb)

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -268,7 +268,7 @@ async fn set_assignee(
     ctx: &Context,
     issue: &Issue,
     github: &GithubClient,
-    reviewer: &Reviewer,
+    reviewer: &ReviewerSelection,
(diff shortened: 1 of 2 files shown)


Kobzol:
Fair enough! Implemented this in the last commit.

Urgau (approved):

PR #2025, opened 2025-05-29
Add `lookup [github|zulip] <username>` command for determining the GitHub/Zulip username mapping

Kobzol:
This should make it much easier to figure out a GitHub username from Zulip name, and Zulip name from a GitHub username.

Suggested here: [#t-compiler/contrib-private > Nominating mati865 and madsmtm @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/244344-t-compiler.2Fcontrib-private/topic/Nominating.20mati865.20and.20madsmtm/near/520956178).

I know that the code of the Zulip integration is... not great. I just copy-pasted stuff around for this, but after this PR (and after https://github.com/rust-lang/rust-forge/pull/853, which shuffles some Zulip docs and mentions Zulip DM commands), I want to do the following:
- Split the `zulip.rs` file into multiple modules
- Rewrite the command handling using e.g. `clap` (or something similar), to both remove the manual parsing and provide automatically generated error messages for users
- Use an actual Zulip API client struct, rather than copy pasting loading of the Zulip API token from the environment and Zulip API error handling all over the place.
- Document this (and the previous `whoami` command in the Forge)

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -617,7 +617,9 @@ pub(crate) struct ZulipUser {
 }
 
 impl ZulipUser {
-    // The GitHub profile data key is 3873
+    // The custom profile field ID for GitHub profiles on the Rust Zulip
+    // is 3873. This is likely not portable across different Zulip instance,
+    // but we assume that triagebot will only be used on this Zulip instance anyway.
     pub(crate) fn get_github_username(&self) -> Option<&str> {
         self.profile_data.get("3873").map(|v| v.value.as_str())
     }


src/zulip.rs
```diff
@@ -476,6 +479,206 @@ async fn whoami_cmd(
...
+    #[serde(rename = "full_name")]
+    pub(crate) name: String,
+    #[serde(default)]
+    pub(crate) profile_data: HashMap<String, ProfileValue>,
+}
+
+impl ZulipUser {
+    // The GitHub profile data key is 3873
+    pub(crate) fn get_github_username(&self) -> Option<&str> {
+        self.profile_data.get("3873").map(|v| v.value.as_str())
```
Urgau, line 622: How did you figure this out? Is there any documentation about it? Or stability at all?

Urgau:
> I want to do the following

100% on board, as always feel free to assign me; also happy to discuss some details before hand

src/zulip.rs
Kobzol: I have no idea, to be honest :/ I just tried to load the data from Rust Zulip, and this ID is used for GitHub user profiles. I guess that someone created it for our Zulip (https://zulip.com/help/custom-profile-fields), and it got assigned some ID. It's probably not portable across Zulip instances, but hopefully it should stay stable on the Rust Zulip.

Urgau: Hum, I didn't knew it was linked to custom-profile-fields, I always though it was "native" integration.

Would be good to mention that in the function, with the fact that the ID is r-l Zulip instance specific.

Urgau (approved):
The logic looks okay to me.

It's unfortunate that it's anything but trivial, but it's only for a command, so let's go.

PR #2030, opened 2025-05-30
Bring concerns visual to modern GitHub UI feature

Urgau:
@jieyouxu and I had a long and productive discussions about the visual of concerns, and in particular the fact that they can be buried in a wall of text, see https://github.com/rust-lang/compiler-team/issues/864#issue-2999208181 as an example.

We therefor iterated to multiple design and ideas, and we end-up makes 3 choices to drastically improve the visual:
1. the all section is placed inside a `[!CAUTION]` or `[!NOTE]`
2. the number of active concerns is put next to the title, ie. `Concerns (3 active)`
3. we removed the author from the display
   - as it doesn't seems to bring much and can be detrimental to the readability, particularly when a concern is resolved (can still be fetch by just clicking on the issue comment)

------------

> [!CAUTION]
> # Concerns (1 active)
>
> - [This is my concern about concern](https://github.com/fake-comment-1234)
> - ~~[This is a resolved concern](https://github.com/fake-comment-4561)~~ resolved [in this comment](https:://github.com/fake-comment-8888)
>
> *Managed by `@rustbot`—see [help](https://forge.rust-lang.org/triagebot/concern.html) for details.*

------------

> [!NOTE]
> # Concerns (0 active)
>
> - ~~[This is a resolved concern](https://github.com/fake-comment-4561)~~ resolved [in this comment](https:://github.com/fake-comment-8888)
>
> *Managed by `@rustbot`—see [help](https://forge.rust-lang.org/triagebot/concern.html) for details.*

### DIFF
--- a/src/handlers/concern.rs
+++ b/src/handlers/concern.rs
@@ -154,38 +154,48 @@ fn markdown_content(concerns: &[Concern], bot: &str) -> String {
         return "".to_string();
     }
 
+    let active_concerns = concerns
+        .iter()
+        .filter(|c| matches!(c.status, ConcernStatus::Active))
+        .count();
+
     let mut md = String::new();
 
-    let _ = writeln!(md, "\n# Concerns");
     let _ = writeln!(md, "");
+
+    if active_concerns > 0 {
+        let _ = writeln!(md, "> [!CAUTION]");
+    } else {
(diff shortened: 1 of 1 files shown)


Urgau:
I had some ideas with the author field, like auto-ping after a certain times, ... but that would require a proper user id instead, so removed it.

PR #2032, opened 2025-05-30
Migrate `EditIssueBody` to database state storage

Urgau:
This PR migrate our `EditIssueBody` from storing the state inside the issue body it-self to using the database storage.

This includes a migration code, if we find an old data in the issue body we prefer that one over the default one, but of-course we no longer save it

This should avoid issues when multiple commands using `EditIssueBody` are used in the same comment, as they would overlap each other due to `EditIssueBody` [re-using the same issue body](https://github.com/rust-lang/triagebot/blob/7044e5449aaf29eea939bc7db81c63b7c872b9b8/src/interactions.rs#L72) instead of fetching the new one.

What this PR does is pretty straight forward (doesn't use anything new), but I haven't tested it.

Given the small amount of users of `EditIssueBody`, I migrated everything in this PR.

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -111,7 +111,7 @@ struct Reviewers {
 }
 
 /// Assignment data stored in the issue/PR body.
-#[derive(Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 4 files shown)


Kobzol (commented):
I tested it [here](https://github.com/Kobzol/bors-kindergarten/issues/3) and it seems to work fine. Thanks!

Kobzol (approved):

PR #2036, opened 2025-05-31
Use GitHub author association for determine the welcome message

Urgau:
Using [GitHub author association](https://docs.github.com/en/graphql/reference/enums#commentauthorassociation) field for determining if the welcome message should be used should be good enough and contrary to the `is_new_contributor` query is free use, as it is in the issue.

cc #1910 @ehuss

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -31,6 +31,7 @@ use crate::{
     interactions::EditIssueBody,
 };
 use anyhow::{bail, Context as _};
+use octocrab::models::AuthorAssociation;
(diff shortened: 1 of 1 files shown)


ehuss (approved):
Thanks! I assume this should work, but I don't really have an easy way to test without creating a new repo.

I'm going offline so I'm not going to merge right now. Maybe Monday, or earlier if Kobzol is online and can deal with any issues that crop up.

Kobzol:
I think that this is a reasonable way of detecting new contributors, as long as GitHub's description of the first time contributor association doesn't lie :) Let's try it.

PR #2044, opened 2025-06-03
Add a Zulip client API

Kobzol:
First step of cleaning up the `zulip` module. Introduces a separate `ZulipClient` that holds the Zulip instance URL, bot e-mail and also loads the API token.


Urgau:
I assume one of the next steps is going to remove those weird `MessageApiRequest`, `AddReaction`, ... structs and instead call directly the methods on `ZulipClient`, like we do with `GitHubClient`.

PR #2047, opened 2025-06-04
Update dependencies

Kobzol:
I tried to only update things that wouldn't require (almost any) code changes. `cargo build --all-targets` goes from 394 dependencies to 367 dependencies.


Kobzol:
:crossed_fingers: Please let this not break anything :laughing:

PR #2048, opened 2025-06-04
Have triagebot allow T-compiler/ops to do what WG-prio could

apiraino:
Still removing leftovers of the WG-prioritization.

My thinking with this change is to transition what the old WG-prio could do to T-compiler/ops so that nothing for me changes. [WG-prio](https://github.com/rust-lang/team/blob/master/teams/archive/wg-prioritization.toml) and [t-compiler/ops](https://github.com/rust-lang/team/blob/master/teams/compiler-ops.toml) are for now the same.

Makes sense, right?

r? @jieyouxu


jieyouxu:
That definitely makes sense. Although, is compiler/ops a strict subteam of compiler? I.e. are all compiler/ops members part of compiler parent team?

Kobzol (approved):

apiraino:
> I would prefer we completly remove the special check, which was required before due to wg not being included in the special "all" team, but it's replacement T-compiler/ops is a proper team and must be included in it. (I have verified that you are in the "all" members)

for me totally fine to also remove the special check. But I will need to check if `compiler-ops` has the same privileges of the old wg-prio.

> is compiler/ops a strict subteam of compiler

compiler-ops is [wg-prio 2.0](https://rust-lang.zulipchat.com/#narrow/channel/227806-t-compiler.2Fops/topic/Winding.20down.20WG-prioritization/near/519889280) :-) wg-prio members were mostly :ghost: so by starting from a clean slate I get pruning inactive members for free.

apiraino:
@Urgau how does it look now?

Urgau (approved):

PR #2049, opened 2025-06-04
Parse Zulip commands with `clap`

Kobzol:
This removes the annoying manual parsing routines and makes the whole command surface easier to understand and discover. It also automatically generates help.

I didn't actually know what are the meta/notification commands for, so I inferred their behavior from code to generate the doc comments.

### DIFF
--- a/src/zulip/commands.rs
+++ b/src/zulip/commands.rs
@@ -1,11 +1,12 @@
 use crate::db::notifications::Identifier;
 use crate::db::review_prefs::RotationMode;
+use clap::{ColorChoice, Parser};
 use std::num::NonZeroU32;
 use std::str::FromStr;
 
 /// Command sent in a DM with triagebot on Zulip.
 #[derive(clap::Parser, Debug)]
-#[clap(override_usage("<command>"))]
+#[clap(override_usage("<command>"), disable_colored_help(true))]
 pub enum ChatCommand {
     /// Acknowledge a notification
     #[clap(alias = "ack")]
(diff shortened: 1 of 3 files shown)


src/zulip/commands.rs
```diff
@@ -0,0 +1,169 @@
+use crate::db::notifications::Identifier;
+use crate::db::review_prefs::RotationMode;
+use clap::{ColorChoice, Parser};
+use std::num::NonZeroU32;
+use std::str::FromStr;
+
+/// Command sent in a DM with triagebot on Zulip.
+#[derive(clap::Parser, Debug)]
+#[clap(override_usage("<command>"), disable_colored_help(true))]
+pub enum ChatCommand {
```
Urgau, line 10: That that we have the parsing independent from the execution, could you add some basic unit-tests, just to make sure the parsing of commands is as we expect them to.

src/zulip/commands.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+impl FromStr for WorkqueueLimit {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "unlimited" => Ok(Self::Unlimited),
+            v => v.parse::<u32>().map_err(|_|
+                                              "Wrong parameter format. Must be a positive integer or `unlimited` to unset the limit.".to_string(),
```
Urgau: This is weirdly formatted.

src/zulip/commands.rs
Kobzol: Oh that was a great idea :laughing: Turns out I was not passing the first "fake" argument to clap, so it wouldn't work at all :laughing:

src/zulip/commands.rs
Kobzol: Reformatted.

Urgau (approved):

PR #2053, opened 2025-06-05
Store PR title in the workqueue and output it in `work show`

Kobzol:
Requested by @BoxyUwU. I think it makes a lot of sense, in order to provide more useful `work show` output and let people see their queue on Zulip only, without going to GitHub.

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -289,7 +289,7 @@ async fn workqueue_commands(
                 .await
                 .into_iter()
                 .collect::<Vec<_>>();
-            assigned_prs.sort();
+            assigned_prs.sort_by_key(|(pr_number, _)| *pr_number);
 
             let review_prefs = get_review_prefs(&db_client, gh_id)
                 .await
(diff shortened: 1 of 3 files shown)


src/zulip.rs
```diff
@@ -309,9 +309,14 @@ async fn workqueue_commands(
 
             let prs = assigned_prs
                 .iter()
-                .map(|pr| format!("- [#{pr}](https://github.com/rust-lang/rust/pull/{pr})"))
+                .map(|(pr_number, pr)| {
+                    format!(
+                        "- [#{pr_number}](https://github.com/rust-lang/rust/pull/{pr_number}) {}",
```
Urgau, line 314: Should we worry about the same "injection" attack that prompted https://github.com/rust-lang/triagebot/pull/2054 ?

Kobzol: Well, yeah, it could happen for the PR title. But there's probably many places like this, sigh. Do you want me to wrap the PR title in backticks?

Urgau: Yeah, there is probably a lot of other places where this could happen.

I'm not sure putting it in backtick would work, I often use backticks in my own PRs, which would render weird and defeat the purpose.

I feel like if we wanted to do the right we would need to bring out a markdown parser and only collect the pure text portions, but that's much more complicated. Let's defer that.

Urgau (approved):

PR #2055, opened 2025-06-05
Add team API client and cache team API calls

Kobzol:
We are now starting to do more and more team API calls with the new Zulip commands, and frankly the fact that these calls were sprinkled all over the codebase without any caching was triggering me :laughing: This PR adds a team API client to centralize the calls, and adds 2 minute cache to the requests, to avoid the network roundtrip latency for each request.

### DIFF
--- a/src/team_data.rs
+++ b/src/team_data.rs
@@ -6,7 +6,7 @@ use std::time::{Duration, Instant};
 use tokio::sync::RwLock;
 
 #[derive(Clone)]
-pub struct TeamApiClient {
+pub struct TeamClient {
     base_url: String,
     client: Client,
     teams: CachedTeamItem<Teams>,
@@ -14,7 +14,7 @@ pub struct TeamApiClient {
     zulip_mapping: CachedTeamItem<ZulipMapping>,
 }
 
-impl TeamApiClient {
+impl TeamClient {
     pub fn new_from_env() -> Self {
         let base_url = std::env::var("TEAMS_API_URL").unwrap_or(BASE_URL.to_string());
         Self::new(base_url)
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -7,7 +7,7 @@ use serde::{Deserialize, Serialize};
 use tera::{Context, Tera};
 
 use crate::github::{self, GithubClient, Repository};
-use crate::team_data::TeamApiClient;
+use crate::team_data::TeamClient;
 
 #[async_trait]
 pub trait Action {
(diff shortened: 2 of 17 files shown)


src/team_data.rs
```diff
@@ -1,44 +1,67 @@
...
 use anyhow::Context as _;
+use reqwest::Client;
 use rust_team_data::v1::{People, Teams, ZulipMapping, BASE_URL};
 use serde::de::DeserializeOwned;
 
-async fn by_url<T: DeserializeOwned>(client: &GithubClient, path: &str) -> anyhow::Result<T> {
-    let base = std::env::var("TEAMS_API_URL").unwrap_or(BASE_URL.to_string());
-    let url = format!("{}{}", base, path);
+#[derive(Clone)]
+pub struct TeamApiClient {
```
Urgau: Wouldn't it be more consistent to use `TeamClient` instead of `TeamApiGithub`?

After all, we don't have a `GithubApiClient` but just a `GithubClient` despite it being `api.github.com`, probably similar for `ZulipClient`.

src/team_data.rs
```diff
@@ -84,24 +92,68 @@ impl TeamApiClient {
...
+        let now = Instant::now();
+        {
+            let value = self.value.read().await;
+            if let CachedValue::Present {
+                value,
+                last_download,
+            } = &*value
+            {
+                if *last_download + CACHE_DURATION > now {
+                    return Ok(value.clone());
```
Urgau, line 134: Team data can be quite big, it seems a bit wasteful to clone the previous data we have in the cache. I supposed we can easily get a reference out of it.

Urgau:
The needs a rebase, there are some conflicts.

src/team_data.rs
Kobzol: Fair enough! Done.

src/team_data.rs
Kobzol: We store the values in a thread-safe RwLock, so that it can be accessed behind `&self`. We would need to extend the duration of the lock if we returned references, and essentially return some `LockGuard` objects. I think that's really not worth it here. Before, the code did network roundtrip + deserialize + build a new instance of the team data structure everytime triagebot accessed something. Now it's just a clone. I think it's fine performance-wise.

Kobzol:
Rebased.

src/team_data.rs
Urgau: I shouldn't write review comments when I'm tiered. I wanted to say: "**can't** easily get a reference out of it".

> Now it's just a clone. I think it's fine performance-wise.

Yeah, it's probably fine.

Urgau (approved):

Issue #2058, opened 2025-06-06
Edge case: suppress off-rotation or exceeds-capacity message when self-`r?`?


jieyouxu:
I sometimes use `r? jieyouxu` because it might be easier in the middle of a message or sth. Can the off-rotation or exceed PR limit message be suppressed if the person themselves is writing `r? <myself>`?

Kobzol:
Hmm, it didn't occur [here](https://github.com/rust-lang/rust/pull/112049#issuecomment-2949035696), so not sure what happened there.

jieyouxu:
[There](https://github.com/rust-lang/rust/pull/112049#issuecomment-2949065183) 😼

Kobzol:
Oh, I see, I got it wrong. We have a special case for the *author of the PR*, not the *author of the comment*. So I guess that we should also allow the auto-assign for the comment author.

PR #2060, opened 2025-06-06
Fix self-assign fast path

Kobzol:
We were checking the author of the PR, rather than the author of the comment.

Fixes: https://github.com/rust-lang/triagebot/issues/2058

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -397,6 +397,7 @@ async fn determine_assignee(
             &teams,
             config,
             &event.issue,
+            &event.issue.user.login,
             &[name],
         )
         .await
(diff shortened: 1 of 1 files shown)


jieyouxu (approved):
Looks good, I don't work on the triagebot codebase tho 😄

jieyouxu:
r? triagebot

Kobzol:
It should be the same, because the `determine_assignee` is only called on PR being opened or converted from draft. It seemed more explicit like this, because `event.user().login` always confuses me a bit. But if you prefer the `user()` function, I'll switch it.

Urgau:
I don't have a preference. Let's merge as is.

Urgau (approved):

Issue #2041, opened 2025-06-02
Have typed-functions instead `replaces` when creating messages


Urgau:
> One more place that be turned from a constant into a function, using these `replaces` on multiple places in code is brittle (doesn't need to be in this PR).

_Originally posted by @Kobzol in https://github.com/rust-lang/triagebot/pull/2037#discussion_r2120322320_

--------

We have multiples places in triagebot that uses the pattern const+`.replace()` to create a message with a changing content, this is brittle, as we may miss to replace placeholders.

It would be better to have typed-functions instead.

```rust
// Before
const WELCOME_WITH_REVIEWER: &str = "@{assignee} (or someone else)";

fn my_logic() {
    let assignee = "rustbot";
    let message = WELCOME_WITH_REVIEWER.replace("{assignee}", assignee);
}

// After
fn my_logic() {
    let assignee = "rustbot";
    let message = welcome_with_reviewer(assignee);
}

fn welcome_with_reviewer(assignee) -> String {
    format!("@{assignee} (or someone else)")
}
```

PR #2064, opened 2025-06-07
Typed-functions instead `const` + `replace` pattern when creating messages

shard77:
Hello, this PR fixes the addressed issue https://github.com/rust-lang/triagebot/issues/2041.

I have also made slight modifications to some variable names, e.g. I have changed `who_text` to `assignee_text` because it seems to make more sense. But let me know if I should rollback those changes.

PS: Maybe it is worth opening another issue for this, but what's the point of doing:
```rust
REVIEWER_IS_PR_AUTHOR.replace("{username}", username)
...
REVIEWER_ALREADY_ASSIGNED.replace("{username}", username)
```
Those two original constants, do not contain any `username` placeholders.

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -43,60 +43,13 @@ use tokio::sync::RwLock;
 use tokio_postgres::Client as DbClient;
 use tracing as log;
 
+mod messages;
+
 #[cfg(test)]
 mod tests {
     mod tests_candidates;
     mod tests_from_diff;
 }
-
-const NEW_USER_WELCOME_MESSAGE: &str = "Thanks for the pull request, and welcome! \
-The Rust team is excited to review your changes, and you should hear from {who} \
-some time within the next two weeks.";
-
-const CONTRIBUTION_MESSAGE: &str = "Please see [the contribution \
-instructions]({contributing_url}) for more information. Namely, in order to ensure the \
-minimum review times lag, PR authors and assigned reviewers should ensure that the review \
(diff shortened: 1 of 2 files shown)


src/handlers/assign.rs
```diff
@@ -49,33 +49,46 @@ mod tests {
     mod tests_from_diff;
 }
 
-const NEW_USER_WELCOME_MESSAGE: &str = "Thanks for the pull request, and welcome! \
-The Rust team is excited to review your changes, and you should hear from {who} \
-some time within the next two weeks.";
+fn new_user_welcome_message(reviewer: &str) -> String {
```
Urgau: All of those messages function now take up a lot of space, how we move them to a `assign/messages.rs` submodule?

We would then refer to them by explicit module naming: `messages::new_user_welcome_message(...)`.

Urgau (commented):
Thanks for the contribution, this looks pretty good to me.

Urgau:
> Maybe it is worth opening another issue for this, but what's the point of doing: [...]
> 
> Those two original constants, do not contain any `username` placeholders.

That's a great find. They serve no purpose, we should remove the `replace`s.

src/handlers/assign.rs
shard77: Good point, I also added the constants there.

shard77:
> > Maybe it is worth opening another issue for this, but what's the point of doing: [...]
> > Those two original constants, do not contain any `username` placeholders.
> 
> That's a great find. They serve no purpose, we should remove the `replace`s.

Removed them :+1:

Urgau (approved):

PR #2051, opened 2025-06-04
Announce concerns on major change in Zulip as well

Urgau:
This PR makes it so that when the "concerns label" is added to a major change a message is issued on Zulip about added concern(s).

To know the label that should be used to know if there are concerns a new (optional) configuration is used `concerns_label`.

When raising concerns:
> Concern(s) have been raised on the [associated GitHub issue]({}). This proposal is now blocked until those concerns are fully resolved.

When resolving all concerns:
> All concerns on the [associated GitHub issue]({}) have been resolved, this proposal is no longer blocked, and will be approved in 10 days if no (new) objections are raised.

or

> All concerns on the [associated GitHub issue]({}) have been resolved, this proposal is no longer blocked.

cc @jieyouxu

PR #2066, opened 2025-06-07
Add automatic closing of major change after their waiting period

Urgau:
This PR adds automatic closing of major change after their waiting period.

This is made possible by https://github.com/rust-lang/triagebot/pull/2022 and https://github.com/rust-lang/triagebot/pull/2051.

The way it works is by creating one-shot schedule jobs, based on the (newly added `waiting_period` config).

> The final comment period is now complete, this major change is now accepted.
> 
> As the automated representative, I would like to thank the author for their work and everyone else who contributed to this major change proposal.

I haven't tested it, as it requires postgress, but I intend to test it on this repo before enabling it anywhere else. I should have added sufficient logging to be able to debug any issues that comes up.

cc @apiraino

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -400,6 +400,9 @@ pub(crate) struct MajorChangeConfig {
     pub(crate) meeting_label: String,
     /// This label signals there are concern(s) about the proposal.
     pub(crate) concerns_label: Option<String>,
+    /// An optional duration for the waiting period after second for the
+    /// major change to become automaticaly accepted.
+    pub(crate) waiting_period: Option<u16>,
     /// The Zulip stream ID where the messages about the status of
     /// the major changed should be relayed.
     pub(crate) zulip_stream: u64,
--- a/src/handlers/major_change.rs
+++ b/src/handlers/major_change.rs
@@ -1,3 +1,4 @@
+use crate::jobs::Job;
 use crate::zulip::api::Recipient;
 use crate::{
     config::MajorChangeConfig,
@@ -6,7 +7,10 @@ use crate::{
     interactions::ErrorComment,
 };
 use anyhow::Context as _;
+use async_trait::async_trait;
+use chrono::{DateTime, Duration, Utc};
 use parser::command::second::SecondCommand;
+use serde::{Deserialize, Serialize};
 use tracing as log;
 
 #[derive(Clone, PartialEq, Eq, Debug)]
@@ -274,6 +278,41 @@ pub(super) async fn handle_command(
         false,
     )
     .await
+    .context("unable to process second command")?;
+
+    // Add MCP to the MCP acceptence queue to be close automaticaly
+    if let Some(waiting_period) = &config.waiting_period {
+        let seconded_at = Utc::now();
+        let accept_at = if issue.repository().full_repo_name() == "rust-lang/triagebot" {
+            // Hack for the triagebot repo, so we can test more quickly
+            seconded_at + Duration::minutes(5)
+        } else {
+            seconded_at + Duration::days((*waiting_period).into())
+        };
+
+        let major_change_seconded = MajorChangeSeconded {
+            repo: issue.repository().full_repo_name(),
+            issue: issue.number,
+            seconded_at,
+            accept_at,
+        };
+
+        tracing::info!(
+            "major_change inserting to acceptence queue: {:?}",
+            &major_change_seconded
+        );
+
+        crate::db::schedule_job(
+            &*ctx.db.get().await,
+            MAJOR_CHANGE_ACCEPTENCE_JOB_NAME,
+            serde_json::to_value(major_change_seconded).unwrap(),
+            accept_at,
+        )
+        .await
+        .context("failed to add the major change to the automatic acceptance queue")?;
+    }
+
+    Ok(())
 }
 
 async fn handle(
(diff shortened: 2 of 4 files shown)


src/handlers/major_change.rs
```diff
@@ -362,3 +401,156 @@ fn zulip_topic_from_issue(issue: &ZulipGitHubReference) -> String {
...
+    issue
+        .add_labels(
+            &ctx.github,
+            vec![Label {
+                name: config.accept_label.clone(),
+            }],
+        )
+        .await
+        .context("unable to add the accept label")?;
+    issue
```
fmease, line 573: I know this is a feature request on top of a feature request and needs separate discussion (and a separate PR most likely):

It would be more than awesome if we could automatically open a T-compiler tracking issue over at r-l/r to avoid accepted MCP falling into oblivion (that's been discussed in the past like maybe a year ago).

Urgau: Sure, that should be technically doable.

Could you open a issue with what you would except triagebot to put in the tracking issue (MCP title, author?, template?, ...).

fmease: Opened https://github.com/rust-lang/triagebot/issues/2067. I will also open a Zulip topic I guess or is there one already for this feature (namely, more automated MCP processes)?

Urgau: We (T-triagebot) discussed it a bit when adding `concern` in https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/triagebot.20concern/with/520721923 but that's was only internal.

Otherwise, no there isn't any active Zulip topic discussing "automated MCP processes".

src/config.rs
```diff
@@ -400,6 +400,9 @@ pub(crate) struct MajorChangeConfig {
     pub(crate) meeting_label: String,
     /// This label signals there are concern(s) about the proposal.
     pub(crate) concerns_label: Option<String>,
+    /// An optional duration for the waiting period after second for the
```
Kobzol: Would be nice to specify the unit here. It is days, right?

src/handlers/major_change.rs
```diff
@@ -362,3 +401,156 @@ fn zulip_topic_from_issue(issue: &ZulipGitHubReference) -> String {
...
+
+    let repo = ctx
+        .github
+        .repository(&major_change.repo)
+        .await
+        .context("failed retrieving the repository informations")?;
+
+    let config = crate::config::get(&ctx.github, &repo)
+        .await
+        .context("failed to get the repositor configuration")?;
```
Kobzol:
```suggestion
        .context("failed to get the repository configuration")?;
```

Kobzol (commented):
The implementation looks reasonable, although if there ever was some "permanent" error in it (e.g. someone removed `major_change` from the triagebot config of the repo), the job would just get repeated an inifinite number of times. We should probably modify the job to return a more semantic error information, that would distinguish "retry" vs "do not retry". It is done partially already, but the `context(...)?` calls will just cause a retry.

I like the idea of automating this, but I'd also like to hear from apiraino.

src/config.rs
Urgau: yes, it's in days

Urgau:
> We should probably modify the job to return a more semantic error information

Added a proper error type, with more semantic[^1] error types. There could still be some errors, like if the repository was deleted (as opposed to a transient error), but I think that could be dealt later.

[^1]: I called them "logic error" in the code, let me know if you want me to change it.

src/handlers/major_change.rs
```diff
@@ -362,3 +404,183 @@ fn zulip_topic_from_issue(issue: &ZulipGitHubReference) -> String {
...
+    }
+}
+
+async fn process_seconded(
+    ctx: &super::Context,
+    major_change: &MajorChangeSeconded,
+    now: DateTime<Utc>,
+) -> anyhow::Result<()> {
+    if major_change.accept_at < now {
+        anyhow::bail!(SecondedLogicError::NotYetAcceptenceTime {
```
Kobzol, line 513: This error should actually cause a retry, I guess, because of timer differences.

src/handlers/major_change.rs
```diff
@@ -362,3 +404,183 @@ fn zulip_topic_from_issue(issue: &ZulipGitHubReference) -> String {
...
+
+    let issue = repo
+        .get_issue(&ctx.github, major_change.issue)
+        .await
+        .context("unable to get the associated issue")?;
+
+    if issue
+        .labels
+        .iter()
+        .any(|l| Some(&l.name) == config.concerns_label.as_ref())
```
Kobzol, line 542: So what happens if someone does a second and then someone adds concerns? Is the automatic accept never retried? I wonder if we should also trigger this job once concern(s) are removed and the MCP was already seconded before?

src/handlers/major_change.rs
Urgau: We fetch jobs to run with only a 1 minute of interval,
https://github.com/rust-lang/triagebot/blob/3a6a2907cb211454fec5d0fb0ee646cb26db0170/src/jobs.rs#L61
and the SQL query specifically look for jobs `scheduled_at <= now()`
https://github.com/rust-lang/triagebot/blob/3a6a2907cb211454fec5d0fb0ee646cb26db0170/src/db/jobs.rs#L106
so I don't think it should ever be possible to reach this condition.

Kobzol: Hopefully the SQL and Rust times will be in sync. You can second a job twice :laughing: But then it should also be scheduled twice, so the second run should succeed. Anyway, we don't need to solve all potential edge-cases here, agreed.

src/handlers/major_change.rs
Urgau: > So what happens if someone does a second and then someone adds concerns? Is the automatic accept never retried?

We get into a weird state, where the job we fired at the `second` will execute and reject the acceptance, unless the concern was resolved before, which it weird. Worse thing at happens it the MCP is sometimes closed a little early, it's not dramatic, and concerns after second is pretty rare anyway.

I don't really have a great solution for it yet, the only thing I could think of would be to delete the previous jobs for that major change. That would require some big changes, maybe for a follow-up PR.

> I wonder if we should also trigger this job once concern(s) are removed and the MCP was already seconded before?

Yes, we should, forgot about that case. Fixed.

Urgau:
I'm gonna go ahead and merge this pull request now, as I have some time now to tests and fix any errors that might occur.

Before deploying it in production, I will do a summary and make sure everyone is okay with it.

src/handlers/major_change.rs
```diff
@@ -274,6 +283,52 @@ pub(super) async fn handle_command(
...
     .await
+    .context("unable to process second command")?;
+
+    // Schedule acceptence job to automaticaly close the MCP
+    schedule_acceptence_job(ctx, config, issue).await?;
+
+    Ok(())
+}
+
+async fn schedule_acceptence_job(
```
apiraino, line 294:
```suggestion
async fn schedule_acceptance_job(
```

also in other places

src/handlers/major_change.rs
```diff
@@ -362,3 +417,183 @@ fn zulip_topic_from_issue(issue: &ZulipGitHubReference) -> String {
...
+            open: issue.is_open()
+        });
+    }
+
+    if !issue.labels.iter().any(|l| l.name == config.accept_label) {
+        // Only post the comment if the accept_label isn't set yet, we may be in a retry
+        issue
+            .post_comment(
+                &ctx.github,
+                "The final comment period is now complete, this major change is now accepted.\n\nAs the automated representative, I would like to thank the author for their work and everyone else who contributed to this major change proposal."
```
apiraino, line 559: Since this is a bot closing a human-driven process, I would add a bit or wording along the line of "feel free to reopen if there open concerns"

apiraino:
Did I miss a discussion about this patch?

I ... have written on Zulip in different occasions how I was hesitant to autoclose MCP proposals without a cursory overview to check if a) the discussion settled and b) there were no open concerns (registered or not). Sometimes seconding a proposal arrives [too early](https://rust-lang.zulipchat.com/#narrow/channel/233931-t-compiler.2Fmajor-changes/topic/WebAssembly.20for.20procedural.20macros.20compiler-team.23876/near/521069552).

I guess at this point we will see if the autoclosing will be too eager.

src/handlers/major_change.rs
Urgau: Sure, will add that.

Urgau: Opened https://github.com/rust-lang/triagebot/pull/2072

PR #2072, opened 2025-06-09
Improve a major change handling of concerns and auto-close

Urgau:
Address https://github.com/rust-lang/triagebot/pull/2066#discussion_r2135836050
Address https://github.com/rust-lang/triagebot/pull/2066#discussion_r2135840308
As well as the interaction when a `second` is done while there are concerns.


Urgau:
Let's merge it. If @apiraino has other improvements, we can do them later.

Issue #2074, opened 2025-06-12
MCP Playground 2


Urgau:
For testing #2072 mixed with concerns.

> [!NOTE]
> # Concerns (0 active)
>
> - ~~[my-concern](https://github.com/rust-lang/triagebot/issues/2074#issuecomment-2967618259)~~ resolved in [this comment](https://github.com/rust-lang/triagebot/issues/2074#issuecomment-2967645916)
> - ~~[blocking-the-second](https://github.com/rust-lang/triagebot/issues/2074#issuecomment-2967666491)~~ resolved in [this comment](https://github.com/rust-lang/triagebot/issues/2074#issuecomment-2967688537)
>
> *Managed by `@rustbot`—see [help](https://forge.rust-lang.org/triagebot/concern.html) for details.*

PR #2075, opened 2025-06-12
Small fixes for major changes

Urgau:
cf. https://github.com/rust-lang/triagebot/issues/2074#issuecomment-2967661355 and [#triagebot > MCP Playground 2 triagebot#2074 @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/MCP.20Playground.202.20triagebot.232074/near/523801066)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -400,9 +400,12 @@ pub(crate) struct MajorChangeConfig {
     pub(crate) meeting_label: String,
     /// This label signals there are concern(s) about the proposal.
     pub(crate) concerns_label: Option<String>,
-    /// An optional duration (in days) for the waiting period after second for the
-    /// major change to become automaticaly accepted.
-    pub(crate) waiting_period: Option<u16>,
+    /// Waiting period after second for the major change to become accepted.
+    #[serde(default = "MajorChangeConfig::waiting_period_default")]
+    pub(crate) waiting_period: u16,
+    /// Enables automatic closing of the major change when the waiting period is completed.
+    #[serde(default)]
+    pub(crate) auto_closing: bool,
     /// The Zulip stream ID where the messages about the status of
     /// the major changed should be relayed.
     pub(crate) zulip_stream: u64,
@@ -416,6 +419,9 @@ impl MajorChangeConfig {
     }
     fn accept_label_default() -> String {
         String::from("major-change-accepted")
+    }
+    fn waiting_period_default() -> u16 {
+        10
     }
 }
 
--- a/src/handlers/major_change.rs
+++ b/src/handlers/major_change.rs
@@ -211,7 +211,7 @@ pub(super) async fn handle_input(
                 // Re-schedule acceptance job to automaticaly close the MCP
                 schedule_acceptance_job(ctx, config, &event.issue).await?;
 
-                format!("All concerns on the [associated GitHub issue]({}) have been resolved, this proposal is no longer blocked, and will be approved in 10 days if no (new) objections are raised.", event.issue.html_url)
+                format!("All concerns on the [associated GitHub issue]({}) have been resolved, this proposal is no longer blocked, and will be approved in {} days if no (new) objections are raised.", event.issue.html_url, config.waiting_period)
             } else {
                 format!("All concerns on the [associated GitHub issue]({}) have been resolved, this proposal is no longer blocked.", event.issue.html_url)
             },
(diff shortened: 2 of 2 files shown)


src/handlers/major_change.rs
```diff
@@ -210,8 +210,9 @@ pub(super) async fn handle_input(
             }) {
                 // Re-schedule acceptance job to automaticaly close the MCP
                 schedule_acceptance_job(ctx, config, &event.issue).await?;
+                let waiting_period = config.waiting_period.unwrap_or(10);
```
Kobzol: The default `10` is now configured at two separate places. Either create a function for this, or just turn `waiting_period` from `Option<u16>` to `u16` and configure the default using serde?

src/handlers/major_change.rs
```diff
@@ -587,10 +588,10 @@ async fn process_seconded(
             .post_comment(
                 &ctx.github,
                 &*format!(
-                    "The final comment period is now complete, this major change is now **accepted**.\
-                    \
-                    As the automated representative, I would like to thank the author for their work and everyone else who contributed to this major change proposal.\
-                    \
+                    "The final comment period is now complete, this major change is now **accepted**.\n\
```
Kobzol: Using a `r"` string would be a bit nicer, to avoid the `\` newline escapes.

src/handlers/major_change.rs
Urgau: I don't yet want to use serde default, as to would instantaneously enable it which it shouldn't, I will create a method.

src/handlers/major_change.rs
Urgau: Though of it, but the at the same time it shifts the content way to the left which isn't nice.

src/handlers/major_change.rs
Urgau: Create a `displayable_waiting_period` method.

src/handlers/major_change.rs
Urgau: Putting every lines to the far left is better, pushed that.

src/handlers/major_change.rs
apiraino: could we round this to 10+1 days so we're sure that it is 10 days all around the world?

Urgau: I'm not I understand. This "10 days" is purely for display, we don't do any calculation with it.

This change is about keeping the current display, while using the new waiting period config field if specified, and we currently say 10 days everywhere.

Kobzol: > I don't yet want to use serde default, as to would instantaneously enable it which it shouldn't, I will create a method.

Actually, this confuses me. The code only does auto accept if `waiting_period` is `Some`, but here the message about accepting is always sent. Does that mean that `Some` means "accept happens automatically" and `None` means "accept happens manually"? What if I want to accept manually with a different waiting period? :D I think we should separate the waiting period from enabling auto accept.

Urgau: > Does that mean that `Some` means "accept happens automatically" and `None` means "accept happens manually"?

Yes, that's how the field is documented.

> What if I want to accept manually with a different waiting period?

Not a scenario I envisioned. The only reason the waiting period is even configurable is because I didn't wanted to wait 10 days for testing it (but then I still hacked it to 5 minutes in the end ;-)).

> I think we should separate the waiting period from enabling auto accept.

Done, I added a new `auto_closing` field, it defaults to false.

Kobzol: It's likely a use-case that no one's gonna ever hit, so sorry for bothering you with it :laughing: I still think that this is a clearer design, so thank you.

src/config.rs
```diff
@@ -400,9 +400,12 @@ pub(crate) struct MajorChangeConfig {
     pub(crate) meeting_label: String,
     /// This label signals there are concern(s) about the proposal.
     pub(crate) concerns_label: Option<String>,
-    /// An optional duration (in days) for the waiting period after second for the
-    /// major change to become automaticaly accepted.
-    pub(crate) waiting_period: Option<u16>,
+    /// Waiting period after second for the major change to become accepted.
+    #[serde(default = "MajorChangeConfig::waiting_period_default")]
+    pub(crate) waiting_period: u16,
+    /// Enables automatic closing of the major change when the waiting period.
```
Kobzol: Looks like the comment is unfinished?

src/handlers/major_change.rs
```diff
@@ -587,11 +585,11 @@ async fn process_seconded(
...
                 &ctx.github,
                 &*format!(
-                    "The final comment period is now complete, this major change is now **accepted**.\
-                    \
-                    As the automated representative, I would like to thank the author for their work and everyone else who contributed to this major change proposal.\
-                    \
-                    *If you think this major change shouldn't have been accepted, feel free to remove the `{}` label and reopen this issue.*",
+r#"The final comment period is now complete, this major change is now **accepted**.
+
+As the automated representative, I would like to thank the author for their work and everyone else who contributed to this major change proposal.\n\
```
Kobzol: No need for `\n\` since it's now a `r` string.

Kobzol (approved):
Thanks! Feel free to merge with the nits fixed.

src/handlers/major_change.rs
Urgau: Weird, must a wrong rebase. Fixed.

PR #2076, opened 2025-06-13
Add enhanced GitHub Actions raw logs viewer

Urgau:
This PR adds an enhanced (as in colored) GitHub Actions raw logs viewer.

This is useful for us as Github Actions logs viewer is just unusable in rust-lang/rust given the size of our logs (it loads indefinitely) and the raw logs are full of ANSI sequence cluttering the output.

This works by adding a new endpoint `/gha-logs/:owner/:repo/:log-id` (restricted to team repos) that fetch from GitHub REST api the raw logs, embeds it directly in the HTML which are then processed by the [`ansi_up`](https://github.com/drudru/ansi_up) javascript library (loaded from jsdelivr.com CDN).

image

To prevent XSS injection (or just bugs in `ansi_up`) I added the `Content-Security-Policy` with restriction in `script-src`. I also tried adding integrity for `ansi_up` but jsdelivr doesn't guaranteed the same content for minified.

After this, we could adjust rust-log-analyzer to link to us instead of GitHub to ease the use of this functionality.

cc @Mark-Simulacrum


src/gha_logs.rs
```diff
@@ -0,0 +1,111 @@
...
+                .status(StatusCode::INTERNAL_SERVER_ERROR)
+                .body(Body::from(format!("{:?}", e)))
+                .unwrap());
+        }
+    };
+
+    Ok(res)
+}
+
+async fn process_logs(
```
Kobzol, line 78: Could we add some cache for this in `Context`? Maybe a LRU hashmap or something like that? The logs can be large and we probably don't want to spam the GH API when the page is refreshed or a couple of people go visit a link when a job fails.

src/gha_logs.rs
```diff
@@ -0,0 +1,111 @@
...
+    <style>
+        body {{
+            font: 14px SFMono-Regular, Consolas, Liberation Mono, Menlo, monospace;
+            background: #0C0C0C;
+            color: #CCCCCC;
+            white-space: pre;
+        }}
+    </style>
+    <script type="module" nonce="{nonce}">
+        import {{ AnsiUp }} from '{ANSI_UP_URL}'
```
Kobzol, line 148: This is now possible in browsers? Wow :)

src/gha_logs.rs
```diff
@@ -0,0 +1,111 @@
...
+    );
+
+    return Ok(Response::builder()
+        .status(StatusCode::OK)
+        .header(
+            CONTENT_SECURITY_POLICY,
+            format!("script-src 'nonce-{nonce}' {ANSI_UP_URL}"),
+        )
+        .body(Body::from(html))
+        .unwrap());
```
Kobzol:
```suggestion
        ?);
```

src/gha_logs.rs
```diff
@@ -0,0 +1,111 @@
...
+            &github::IssueRepository {
+                organization: owner.to_string(),
+                repository: repo.to_string(),
+            },
+            log_id,
+        )
+        .await
+        .context("unable to get the raw logs")?;
+
+    let json_logs = serde_json::to_string(&logs).context("unable to JSON-ify the raw logs")?;
```
Kobzol: I thought that we could just dump the raw string into the `<body>`, but that's not ideal if it actually contains HTML tags :) I tried it and it breaks stuff.

Kobzol (commented):
This is a really wonderful idea :heart: And I can't wait to use it! I can take care of the RLA integration once/if this is merged.

Kobzol:
We could also just host the ansiup file ourselves through triagebot, it's relativey small (~14 KiB).

src/gha_logs.rs
Urgau: Yeah, I also tried. I had issues with multilines, "octal numbers", escape sequence being interpetred, ... it's not worth the pain, especialy when serde_json can do it for us so simply.

src/gha_logs.rs
Urgau: Yeah same, I wasn't aware of JS modules[^1], but it's been around since ES6 (ECMAScript 2015) and has a wide compatibility[^2].

[^1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#importing_features_into_your_script
[^2]: https://caniuse.com/es6-module

src/gha_logs.rs
Urgau: I though of it, but given some of the size of the logs (some of them go 3-4 Mb each) I'm worried we'll have a huge cache miss due to the limited number of logs we'll be able to cache.

I also don't know how much we can safely cache, I don't how much RAM the triagebot machine has left. Or maybe we'll want to have the cache on disk?

Kobzol: I wouldn't go for a disk cache. I think that even storing a Vec of ~10-100 last results would be enough. Logs are usually accessed only for a brief period of time after CI fails, when a bunch of people may get notified and they go check the logs. After a few days, people usually don't care about the log anymore. So remembering the last N logs to have a history of at least a few hours would help to avoid needless API requests, IMO.

Not sure about RAM size, but storing e.g. 20 results should hopefully be fine. @mark-simulacrum should know the machine specs, I think.

Urgau: Okay, added a small cache of 50Mb for the raw logs in RAM.

src/gha_logs.rs
```diff
@@ -0,0 +1,167 @@
...
+            .context("unable to get the raw logs")?;
+
+        ctx.gha_logs
+            .write()
+            .await
+            .put(log_uuid.clone(), logs.into())
+            .clone()
+    };
+
+    let json_logs = serde_json::to_string(&*logs).context("unable to JSON-ify the raw logs")?;
```
Kobzol: Micro-opt: we could store the JSON encoded logs in the cache. It should be the ~same size, and it saves one JSON encoding per page load. But that doesn't really matter :)

Kobzol (commented):
Impl looks good, but I'd like to hear from Mark or someone else from t-infra about potential security concerns.

src/gha_logs.rs
```diff
@@ -0,0 +1,167 @@
...
+        } else if self.capacity + value.len() as u64 >= MAX_CACHE_CAPACITY_BYTES {
+            self.entries.pop_back()
+        } else {
+            None
+        };
+        if let Some(removed) = removed {
+            self.capacity -= removed.1.len() as u64;
+        }
+        // Insert at the front (most recently used)
+        self.capacity += value.len() as u64;
```
Mark-Simulacrum, line 51: Can we check here that we don't store logs exceeding MAX_CACHE_CAPACITY_BYTES? I think as-is we'd be able to store 1 such log (at least until the next request)

src/gha_logs.rs
Mark-Simulacrum: We currently allocate 1/4 of a CPU and 0.5 GB of memory: https://github.com/rust-lang/simpleinfra/blob/bcba51d544106fba945c904bf213d64a7bb7a473/terraform/shared/services/triagebot/main.tf#L47-L48

Looks like most of the memory is currently unused:

image

I think a small cache isn't bad, but I'd probably suggest that if we want this in the long run it should be backed by S3 (with a, say, 15 day TTL) -- that's pretty cheap and avoids worrying about how much space it uses, and S3 rate limiting is basically impossible. If we did that we could also serve the logs from a CDN rather than through triagebot, which seems also nice.

src/gha_logs.rs
```diff
@@ -0,0 +1,167 @@
...
+use crate::handlers::Context;
+use anyhow::Context as _;
+use hyper::header::{CONTENT_SECURITY_POLICY, CONTENT_TYPE};
+use hyper::{Body, Response, StatusCode};
+use std::collections::VecDeque;
+use std::str::FromStr;
+use std::sync::Arc;
+use uuid::Uuid;
+
+const ANSI_UP_URL: &str = "https://cdn.jsdelivr.net/npm/ansi_up@6.0.6/+esm";
```
Mark-Simulacrum: I think I'd prefer we pin this by hosting the file ourselves.

src/gha_logs.rs
```diff
@@ -0,0 +1,167 @@
...
+    let logs = 'logs: {
+        if let Some(logs) = ctx.gha_logs.write().await.get(&log_uuid) {
+            tracing::info!("gha_logs: cache hit for {log_uuid}");
+            break 'logs logs.clone();
+        }
+
+        tracing::info!("gha_logs: cache miss for {log_uuid}");
+        let logs = ctx
+            .github
+            .raw_job_logs(
```
Mark-Simulacrum, line 111: So, I'm wondering if we actually have to make the request from triagebot. Could the page we serve use JS to request directly from GitHub, using the users own GitHub credentials? The plain logs appear to be served with `CORS: *` -- though not sure if the redirect to them is.

I seem to recall we had even built (or found?) such a page at some point...

Urgau: How would we get the user GitHub credentials? By using a GitHub App or OAuth app for triagebot?

src/gha_logs.rs
Urgau: Sure, add more code serve it ourselves.

src/gha_logs.rs
Kobzol: Doing GH login in triagebot is not trivial and it has been rejected before for the back-office review work, so I don't suppose that's what Mark meant (?). I guess what we could do is just fetch the plain logs URL from the user's browser.

However, it doesn't seem to work? I tried this:
```html
<html>
    <head>
    </head>
    <body>
        <div id="body"></div>
        <script>
            async function load() {
                const data = await fetch("https://github.com/rust-lang/rust/commit/d04c7466850b10b56c28211ccb880ae34915d431/checks/44186934959/logs");
                document.getElementById("body").innerHTML = data;
            }
            load();
        </script>
    </body>
</html>
```
and got this:
image

Kobzol: You can get a `*` CORS URL by visiting the original logs URL, but it is specific for each user and temporary (it lasts like 5 minutes, IIRC). And you can only get it after you visit the original URL in your browser, not via CORS. So I think that we'll have to download it through the API.

Urgau: I see. Just tried with [the API](https://api.github.com/repos/rust-lang/rust/actions/jobs/43666277064/logs), and it leads to a 403 about being unauthenticated:
```json
{
  "message": "Must have admin rights to Repository.",
  "documentation_url": "https://docs.github.com/rest/actions/workflow-jobs#download-job-logs-for-a-workflow-run",
  "status": "403"
}
```

Mark-Simulacrum: Ok, yeah, sounds like the final link is available - in theory, that means we could still leave the download for users, right? I.e., on triagebot side we cache for ~1 minute ttl the resolve of the API URL (stable) to the short-lived plaintext logs link and then have the user's browser download that and render it as needed. That way we don't have triagebot in the critical path for the big object and caches are much smaller per-element, etc.

That said I personally suspect that we might want to keep the rendering plausibly on the server, I'm not sure how well browsers will cope with the megabytes of HTML + JavaScript rendering all the ansi codes... it might be better to strip them on the server side.

Mark-Simulacrum: So let's not change anything here but keep it in mind if we run into issues.

src/gha_logs/ansi_up@6.0.6.min.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Bundled by jsDelivr using Rollup v2.79.2 and Terser v5.39.0.
+ * Original file: /npm/ansi_up@6.0.6/ansi_up.js
+ */
```
Mark-Simulacrum, line 27: What's the license on this? Can we get it copy/pasted into this comment?

src/gha_logs.rs
Kobzol: I don't think there's an API endpoint for figuring out the temporary URL though.

src/gha_logs/ansi_up@6.0.6.min.js
Urgau: Oh yeah, forgot about the license. It's MIT license. Included it.

src/gha_logs.rs
Urgau: > That said I personally suspect that we might want to keep the rendering plausibly on the server, I'm not sure how well browsers will cope with the megabytes of HTML + JavaScript rendering all the ansi codes... it might be better to strip them on the server side.

I have looked at our longest CI jobs (like 5/6Mb of raw jobs) and the Firefox renders it surprisingly fine (less than 1s). The longest part is downloading the raw jobs logs, not rendering it.

> I don't think there's an API endpoint for figuring out the temporary URL though.

It's the same API, but instead of following the redirects we need to look at the `Location:` header of the GitHub API call response instead of following it to it's destination.

Urgau:
Seems like we're in a pretty good shape. Should we merge it @Kobzol ?

Urgau:
> Did you want to keep the JS vs using the Rust code that is now available through your crate?

I'm fine keeping the JS for now, if we run into issue, or want better perf we can always switch to `anstyle_svg`.

Urgau:
:tada: https://triage.rust-lang.org/gha-logs/rust-lang/rust/43666277064

PR #2077, opened 2025-06-13
Fix off rotation count in team-stats

dtolnay:
Fixes the wrong count being used in the header of "OFF rotation".


Kobzol (approved):
Pff, I thought I checked this ten times, but I still made a copy-paste error :facepalm: Thanks!

Issue #2078, opened 2025-06-14
Some character encoding issue with the `team-stats` command on Zulip


fmease:
First noticed in PR https://github.com/rust-lang/triagebot/pull/2077, see the screenshot contained in its description.

The output of `team-stats libs-contributors` contains a butchered display name for [@]joboet, namely `Jonas BÃ¶ttiger` (`Ã¶` U+00C3 U+00B6 in UTF-8 encoded bytes: `C3 83 C2 B6`) while it should be `Jonas Böttiger` (`ö` U+00F6[^1] in UTF-8 encoded bytes: `C3 B6`)

I don't know *where* the (re)encoding issue occurs but at some point in the pipeline the `ö` (U+00F6) gets UTF-8 encoded as `C3 B6` **but read back as ISO-8859-1 or WINDOWS-1252 instead of UTF-8** leading to `Ã¶` (U+00C3 U+00B6) getting output.

[^1]: In *this* case it is — it could've also been a composition of two code points (U+006F and U+0308)) but I checked by copy/pasting from GH.

fmease:
To follow at home, the `ö` (U+0000F6) gets UTF-8-encoded the following way:

* We have 0x00_00_7F < **0x00_00_F6** ≤ 0x07_FF and thus we need two bytes
* The bytes have the form `110xxxyy 10yyzzzz` where 0bxxxyyyyzzzz is the code point
* 0x00_00_F6 is 0b000_1111_0110
* Thus we end up having `11000011 0b10110110` which is `C3 B6`

Let's read the UTF-8-encoded `ö` back as ISO-8859-1 or WINDOWS-1252 into our UTF-8-based terminal emulator:

```console
> printf '\xc3\xb6\n' | iconv - -f ISO-8859-1 -t UTF-8
Ã¶
> printf '\xc3\xb6\n' | iconv - -f WINDOWS-1252 -t UTF-8
Ã¶
```

Kobzol:
Oh no, it's the UTF-8 mess again. I noticed this while implementing a previous command, but didn't realize it would also happen here. Previously, we "fixed" it by rendering Zulip names using Zulip ID, but here the user might not even have Zulip, so we should render the name properly. I'll take a look.

Kobzol:
Fixed by https://github.com/rust-lang/triagebot/pull/2081. Thanks for reporting the issue!

PR #2081, opened 2025-06-16
Return JSON content type in Zulip webhook responses

Kobzol:
I suspect that this might have caused the [encoding problems](https://github.com/rust-lang/triagebot/issues/2078). If not, then I'm out of ideas, on the Rust side everything is UTF-8, I checked the contents of the team API and the strings are correct there.


Kobzol:
Yay!

PR #2086, opened 2025-06-17
Use `ErrorComment` for emitting the validation config error

Urgau:
So it looks like [this](https://github.com/rust-lang/triagebot/pull/1942#issuecomment-2812130163), instead of [this](https://github.com/rust-lang/triagebot/pull/2085#issuecomment-2981899722).

Kind-of related, looking at both our check-commits warnings and this error, I'm wondering if we should use GitHub UI warning and error block.

### DIFF
--- a/src/handlers/check_commits.rs
+++ b/src/handlers/check_commits.rs
@@ -4,6 +4,7 @@ use anyhow::Context as _;
 use anyhow::bail;
 
 use super::Context;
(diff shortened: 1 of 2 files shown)


Kobzol (approved):
Using the block would be more visually distinctive, but I wouldn't put the whole message in it, just some header (e.g. warnings detected).

PR #2088, opened 2025-06-23
GH caching issue

apiraino:
@ehuss do you have thoughts about this approach?

(without looking too much at the messy code)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,5 +1,6 @@
 use crate::changelogs::ChangelogFormat;
 use crate::github::{GithubClient, Repository};
+use octocrab::Octocrab;
 use std::collections::{HashMap, HashSet};
 use std::fmt;
 use std::sync::{Arc, LazyLock, RwLock};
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -525,14 +525,40 @@ fn default_true() -> bool {
...
+    repo: &Repository,
+) -> Result<Arc<Config>, ConfigurationError> {
+    let mut content_items = octocrab::instance()
+        .repos(repo.owner(), repo.name())
+        .get_content()
+        .path(CONFIG_FILE_NAME)
+        .r#ref(&repo.default_branch)
+        .send()
+        .await
+        .map_err(|e| ConfigurationError::Http(Arc::new(e.into())))?;
```
apiraino: Getting the file using the git API reduces the caching time but not completely. It looks like this way it's much shorter. Empirically (looking with `RUST_LOG=trace`) a couple of seconds, just until the connection is closed.

apiraino: IIUC this basically creates everytime a new HTTP client, not great, but this codepath is run just once at the start of triagebot and everytime the `triagebot.toml` of the rust repository is updated.

apiraino: I don't know yet what to think of `REFRESH_EVERY_SECS`.

Kobzol: We can just pass the existing `Context::octocrab` instance to the `get` function to avoid creating a new client. Otherwise this looks fine. I guess that `REFRESH_EVERY_SECS` should work exactly the same as before?

apiraino: I did some more tests: my solution is not reliable, GH cache is just really sticky and I can't figure out a way around it. Certainly I can't have the pretension to outsmart them. I will leave this code here as a memento in case someone wants to try :man_shrugging:

PR #2090, opened 2025-06-29
Also support `concern resolve` to resolve concerns

Urgau:
This PR adds an alias to `@bot resolve` in the form of `@bot concern resolve`, this adds a parallel with `@bot note remove`.

This came up here https://github.com/rust-lang/rust/pull/142824#issuecomment-3016579109.

cc @jieyouxu

### DIFF
--- a/parser/src/command/concern.rs
+++ b/parser/src/command/concern.rs
@@ -25,8 +25,15 @@ impl fmt::Display for ParseError {
 impl ConcernCommand {
     pub fn parse<'a>(input: &mut Tokenizer<'a>) -> Result<Option<Self>, Error<'a>> {
         let mut toks = input.clone();
-        if let Some(Token::Word(action @ ("concern" | "resolve"))) = toks.peek_token()? {
+        if let Some(Token::Word(mut action @ ("concern" | "resolve"))) = toks.peek_token()? {
             toks.next_token()?;
+
+            if action == "concern" {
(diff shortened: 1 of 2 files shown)


parser/src/command/concern.rs
```diff
@@ -25,9 +25,16 @@ impl fmt::Display for ParseError {
 impl ConcernCommand {
     pub fn parse<'a>(input: &mut Tokenizer<'a>) -> Result<Option<Self>, Error<'a>> {
         let mut toks = input.clone();
-        if let Some(Token::Word(action @ ("concern" | "resolve"))) = toks.peek_token()? {
+        if let Some(Token::Word(mut action @ ("concern" | "resolve"))) = toks.peek_token()? {
             toks.next_token()?;
 
+            if action == "concern" {
```
Kobzol, line 31: Hmm, what is someone tries to create a concern (sentence) that starts with the word "resolve"? 😆 Well, probably a too niche of an edge-case.

Kobzol (approved):

parser/src/command/concern.rs
Urgau: I asked jieyouxu about it and I was told "that seems somewhat... strange xD", so left it. We can always add it later.

jieyouxu: Yeah, I thought that was even more niche, and maybe more of a "well don't do that" xD

PR #2091, opened 2025-06-29
Remove `new_pr` labels when converting PR to draft or closing

Urgau:
This PR updates the logic in `[autolabel]` to remove the `new_pr = true` labels when a PR is converted to draft or closed. (we re-add them when reopening a pr or marking the pr as ready for review)

It doesn't add draft labels (ie `S-waiting-on-draft`) but that's pre-existing (opening a draft PR also doesn't add it).

Related to https://github.com/rust-lang/triagebot/pull/1768#issuecomment-3012963312.

### DIFF
--- a/src/handlers/autolabel.rs
+++ b/src/handlers/autolabel.rs
@@ -26,23 +26,32 @@ pub(super) async fn parse_input(
     // FIXME: This will re-apply labels after a push that the user had tried to
     // remove. Not much can be done about that currently; the before/after on
     // synchronize may be straddling a rebase, which will break diff generation.
-    if matches!(
+    let can_trigger_files = matches!(
         event.action,
-        IssuesAction::Opened
-            | IssuesAction::Closed
-            | IssuesAction::Reopened
-            | IssuesAction::Synchronize
-            | IssuesAction::ReadyForReview
-            | IssuesAction::ConvertedToDraft
-    ) {
-        let files = event
-            .issue
-            .diff(&ctx.github)
-            .await
-            .map_err(|e| {
-                log::error!("failed to fetch diff: {:?}", e);
-            })
-            .unwrap_or_default();
+        IssuesAction::Opened | IssuesAction::Synchronize
+    );
+
+    if can_trigger_files
+        || matches!(
+            event.action,
+            IssuesAction::Closed
+                | IssuesAction::Reopened
+                | IssuesAction::ReadyForReview
+                | IssuesAction::ConvertedToDraft
+        )
+    {
+        let files = if can_trigger_files {
+            event
+                .issue
+                .diff(&ctx.github)
+                .await
+                .map_err(|e| {
+                    log::error!("failed to fetch diff: {:?}", e);
+                })
+                .unwrap_or_default()
+        } else {
+            Default::default()
+        };
 
         let mut autolabels = Vec::new();
         let mut to_remove = Vec::new();
--- a/src/handlers/autolabel.rs
+++ b/src/handlers/autolabel.rs
@@ -39,7 +39,10 @@ pub(super) async fn parse_input(
     // synchronize may be straddling a rebase, which will break diff generation.
     if matches!(
         event.action,
-        IssuesAction::Opened | IssuesAction::Synchronize | IssuesAction::ReadyForReview
+        IssuesAction::Opened
+            | IssuesAction::Synchronize
+            | IssuesAction::ReadyForReview
+            | IssuesAction::ConvertedToDraft
     ) {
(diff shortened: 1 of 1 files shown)


src/handlers/autolabel.rs
```diff
@@ -108,6 +113,14 @@ pub(super) async fn parse_input(
                     });
                     state.data.new_pr_labels_applied = true;
                 }
+
+                // If a PR is converted to draft remove all the "new PR" labels
+                if cfg.new_pr && event.action == IssuesAction::ConvertedToDraft {
+                    to_remove.push(Label {
+                        name: label.to_owned(),
+                    });
+                    state.data.new_pr_labels_applied = false;
```
Kobzol: Hmm, not 100% sure about this one. But I guess that it makes sense; before we handled PRs opened as drafts as a special case, now being in draft is not so special anymore, and everytime you convert from draft to non-draft, it's similar to a new PR being opened 🤔

Kobzol (commented):
When a PR is converted to draft, the `trigger_files` autolabelling probably shouldn't be triggered 🤔

src/handlers/autolabel.rs
Urgau: Yeah, I was a bit confused by this. It personally make sense that we treat draft to non-draft as new PR, the only place we currently use it is for `S-waiting-on-review` on r-l/r anyway.

Also I don't think we're using `state.data.new_pr_labels_applied` anymore, should I remove it?

Kobzol: Well we use it here: https://github.com/rust-lang/triagebot/blob/205897ccd821bf7b2de7ccc0644a90e34ed24e47/src/handlers/autolabel.rs#L103. But after this PR, we should remove that usage, to always apply the `new_pr` labels when a PR is undrafted.

Urgau: Yeah, that's what I meant. Removing `new_pr_labels_applied` because `ReadyForReview` <-> `ConvertedToDraft`.

Kobzol: Yeah, let's do that.

src/handlers/autolabel.rs
```diff
@@ -97,16 +85,21 @@ pub(super) async fn parse_input(
...
+                let is_ready_for_review = event.action == IssuesAction::ReadyForReview;
+                if cfg.new_pr && (is_new_non_draft_pr || is_ready_for_review) {
                     autolabels.push(Label {
                         name: label.to_owned(),
                     });
-                    state.data.new_pr_labels_applied = true;
+                }
+
+                // If a PR is converted to draft remove all the "new PR" labels
+                if cfg.new_pr && event.action == IssuesAction::ConvertedToDraft {
```
Kobzol: I wonder if we should also do this when the PR is closed :thinking: There are many closed PRs marked as waiting on review.

Kobzol: https://github.com/rust-lang/rust/pulls?q=is%3Apr+label%3AS-waiting-on-review+

Urgau: Sure, sounds reasonable. Done.

src/handlers/autolabel.rs
```diff
@@ -97,16 +87,27 @@ pub(super) async fn parse_input(
 
                 // Treat the following situations as a "new PR":
```
Kobzol, line 97: We should only apply the `trigger_files` labels when the PR was pushed to or when it was opened/converted from draft.

src/handlers/autolabel.rs
```diff
@@ -97,16 +96,27 @@ pub(super) async fn parse_input(
...
                 // 1) New PRs opened as non-draft
```
Kobzol:
```suggestion
                // 1) PRs that were (re)opened and are not draft
```

src/handlers/autolabel.rs
```diff
@@ -97,16 +96,27 @@ pub(super) async fn parse_input(
...
-                // 2) PRs opened as draft that are marked as "ready for review" for the first time.
+                // 2) PRs opened as draft that are marked as "ready for review".
```
Kobzol:
```suggestion
                // 2) PRs that have been converted from a draft to being "ready for review".
```

src/handlers/autolabel.rs
```diff
@@ -97,16 +96,27 @@ pub(super) async fn parse_input(
...
                 let is_new_non_draft_pr =
```
Kobzol:
```suggestion
                let is_opened_non_draft =
```

Kobzol (approved):
Feel free to merge with/without the rewordings.

PR #456, opened 2020-04-10
Implement the `prioritize` command

LeSeulArtichaut:
Closes #453 (hopefully :D)
cc @Mark-Simulacrum @spastorino

PR #2092, opened 2025-06-30
Auto-nominate for backport a pull request fixing a regression

apiraino:
From #1184:

We should nominate PRs for backport if any of the following are true:
1. Description mentions a issue which is labeled as [regression]
2. Description mentions a issue which is milestoned with the beta version number

---

This patch implements point (1) and adds a check when a new PR is opened. The new `backport.rs` handler tries to figure to out if a PR should be nominated for backport.

This handler is enabled by adding a new `[backport]` config item. The format allows multiple instances of the config so teams can handle backports as they wish.

```
# "foo" is just to disambiguate multiple instances of this config
[backport.foo]
# The pull request MUST have one of these labels
required_pr_labels = ["T-compiler"]
# The regression MUST have this label
required_issue_labels = "regression-from-stable-to-beta"
# if the above conditions matches, the PR will receive these labels
add_labels = ["beta-nominated"]

[backport.bar]
required_pr_labels = ["T-compiler"]
required_issue_labels = "regression-from-stable-to-stable"
add_labels = ["beta-nominated", "stable-nominated"]

[backport.baz]
required_pr_labels = ["T-libs", "T-libs-api"]
required_issue_labels = "regression-from-stable-to-stable"
add_labels = ["stable-nominated"]
```

The workflow that models the decision to add the backport label is:

0. Someone files a regression `#123`, with a t-label (ex. "T-compiler")
1. Contributor opens PR `#456`, adds comment "Fixes `#123`"
2. triagebot receives the event "new PR"
3. triagebot parses the opening comment and retrieves originating regression(s)
  3a. checks if PR has the required team labels (reads from `required_pr_labels`), loads the appropriate config
  3b. checks if issue has a `regression-from-*` label (reads from `required_issue_labels`)
  3c. checks the P-label of the regression, discard P-low and P-medium regressions   
5. triagebot adds backport label(s) to PR #456 (reads from `add_labels`)

### TODOs

Situations not yet handled where this handler should trigger as well:

- [ ] PR is opened without a T-label (can that happen? or is that automatically applied?), user adds a T-label to the PR later
- [ ] The regression receives a P-label when the PR fixing it already exists
- [x] The expected marker text ("Fixes, resolves, etc.") is added later 
- [ ] probably other cases ...

This work supersedes #1380 (**Important:** before merge this commit should be Co-authored[^1])

r?

[^1]: Not doing now to avoid spamming the original author with multiple forced git push :slightly_smiling_face:

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -47,6 +47,7 @@ pub(crate) struct Config {
     pub(crate) issue_links: Option<IssueLinksConfig>,
     pub(crate) no_mentions: Option<NoMentionsConfig>,
     pub(crate) behind_upstream: Option<BehindUpstreamConfig>,
+    pub(crate) backport: Option<BackportConfig>,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -522,6 +523,25 @@ fn default_true() -> bool {
     true
 }
 
+#[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+pub(crate) struct BackportConfig {
+    // Config identifier -> labels
+    #[serde(flatten)]
+    pub(crate) configs: HashMap<String, BackportRuleConfig>,
+}
+
+#[derive(Default, PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub(crate) struct BackportRuleConfig {
+    /// Prerequisite label(s) (one of them) to trigger this handler for a specific team
+    pub(crate) required_pr_labels: Vec<String>,
+    /// Prerequisite label for an issue to qualify as regression
+    pub(crate) required_issue_label: String,
+    /// Labels to be added to a pull request closing the regression
+    pub(crate) add_labels: Vec<String>,
+}
+
 fn get_cached_config(repo: &str) -> Option<Result<Arc<Config>, ConfigurationError>> {
     let cache = CONFIG_CACHE.read().unwrap();
     cache.get(repo).and_then(|(config, fetch_time)| {
@@ -655,6 +675,11 @@ mod tests {
 
             [behind-upstream]
             days-threshold = 14
+
+            [backport.teamRed]
+            required-pr-labels = ["T-libs", "T-libs-api"]
+            required-issue-label = "regression-from-stable-to-stable"
+            add-labels = ["stable-nominated"]
         "#;
         let config = toml::from_str::<Config>(&config).unwrap();
         let mut ping_teams = HashMap::new();
@@ -679,6 +704,20 @@ mod tests {
         nominate_teams.insert("release".to_owned(), "T-release".to_owned());
         nominate_teams.insert("core".to_owned(), "T-core".to_owned());
         nominate_teams.insert("infra".to_owned(), "T-infra".to_owned());
+
+        let mut backport_configs = HashMap::new();
+        backport_configs.insert(
+            "teamRed".into(),
+            BackportRuleConfig {
+                required_pr_labels: vec!["T-libs".into(), "T-libs-api".into()],
+                required_issue_label: "regression-from-stable-to-stable".into(),
+                add_labels: vec!["stable-nominated".into()],
+            },
+        );
+        let backport_team_config = BackportConfig {
+            configs: backport_configs,
+        };
+
         assert_eq!(
             config,
             Config {
@@ -727,6 +766,7 @@ mod tests {
                 concern: Some(ConcernConfig {
                     labels: vec!["has-concerns".to_string()],
                 }),
+                backport: Some(backport_team_config)
             }
         );
     }
@@ -812,6 +852,7 @@ mod tests {
                 behind_upstream: Some(BehindUpstreamConfig {
                     days_threshold: Some(7),
                 }),
+                backport: None
             }
         );
     }
--- a/src/handlers/backport.rs
+++ b/src/handlers/backport.rs
@@ -1,0 +1,252 @@
+use std::collections::HashMap;
+use std::sync::LazyLock;
+
+use crate::config::BackportConfig;
+use crate::github::{IssuesAction, IssuesEvent, Label};
+use crate::handlers::Context;
+use anyhow::Context as AnyhowContext;
+use futures::future::join_all;
+use regex::Regex;
+use tracing as log;
+
+// See https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue
+// See tests to see what matches
+static CLOSES_ISSUE_REGEXP: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new("(?i)(?P<action>close[sd]*|fix([e]*[sd]*)?|resolve[sd]*)(?P<spaces>:? +)(?P<org_repo>[a-zA-Z0-9_-]*/[a-zA-Z0-9_-]*)?#(?P<issue_num>[0-9]+)").unwrap()
+});
+
+const BACKPORT_LABELS: [&str; 4] = [
+    "beta-nominated",
+    "beta-accepted",
+    "stable-nominated",
+    "stable-accepted",
+];
+
+const REGRESSION_LABELS: [&str; 3] = [
+    "regression-from-stable-to-nightly",
+    "regression-from-stable-to-beta",
+    "regression-from-stable-to-stable",
+];
+
+// auto-nominate for backport only patches fixing high/critical regressions
+// For `P-{medium,low}` regressions, let the author decide
+const PRIORITY_LABELS: [&str; 2] = ["P-high", "P-critical"];
+
+#[derive(Default)]
+pub(crate) struct BackportInput {
+    // Issue(s) fixed by this PR
+    ids: Vec<u64>,
+    // Handler configuration, it's a compound value of (required_issue_label -> add_labels)
+    labels: HashMap<String, Vec<String>>,
+}
+
+pub(super) async fn parse_input(
+    _ctx: &Context,
+    event: &IssuesEvent,
+    config: Option<&BackportConfig>,
+) -> Result<Option<BackportInput>, String> {
+    let config = match config {
+        Some(config) => config,
+        None => return Ok(None),
+    };
+
+    // Only handle events when the PR is opened or the first comment is edited
+    let should_check = matches!(event.action, IssuesAction::Opened | IssuesAction::Edited);
+    if !should_check || !event.issue.is_pr() {
+        log::debug!(
+            "Skipping backport event because: IssuesAction = {:?} issue.is_pr() {}",
+            event.action,
+            event.issue.is_pr()
+        );
+        return Ok(None);
+    }
+    let pr = &event.issue;
+
+    let pr_labels: Vec<&str> = pr.labels.iter().map(|l| l.name.as_str()).collect();
+    if contains_any(&pr_labels, &BACKPORT_LABELS) {
+        log::debug!("PR #{} already has a backport label", pr.number);
+        return Ok(None);
+    }
+
+    // Retrieve backport config for this PR, based on its team label(s)
+    // If the PR has no team label matching any [backport.*.required-pr-labels] config, the backport labelling will be skipped
+    let mut input = BackportInput::default();
+    let valid_configs: Vec<_> = config
+        .configs
+        .iter()
+        .clone()
+        .filter(|(_cfg_name, cfg)| {
+            let required_pr_labels: Vec<&str> =
+                cfg.required_pr_labels.iter().map(|l| l.as_str()).collect();
+            if !contains_any(&pr_labels, &required_pr_labels) {
+                log::warn!(
+                    "Skipping backport nomination: PR is missing one required label: {:?}",
+                    pr_labels
+                );
+                return false;
+            }
+            input
+                .labels
+                .insert(cfg.required_issue_label.clone(), cfg.add_labels.clone());
+            true
+        })
+        .collect();
+    if valid_configs.is_empty() {
+        log::warn!(
+            "Skipping backport nomination: could not find a suitable backport config. Please ensure the triagebot.toml has a `[backport.*.required-pr-labels]` section matching the team label(s) for PR #{}.",
+            pr.number
+        );
+        return Ok(None);
+    }
+
+    // Check marker text in the opening comment of the PR to retrieve the issue(s) being fixed
+    for caps in CLOSES_ISSUE_REGEXP.captures_iter(&event.issue.body) {
+        let id = caps
+            .name("issue_num")
+            .ok_or_else(|| format!("failed to get issue_num from {caps:?}"))?
+            .as_str();
+
+        let id = match id.parse::<u64>() {
+            Ok(id) => id,
+            Err(err) => {
+                return Err(format!("Failed to parse issue id `{id}`, error: {err}"));
+            }
+        };
+        if let Some(org_repo) = caps.name("org_repo")
+            && org_repo.as_str() != event.repository.full_name
+        {
+            log::info!(
+                "Skipping backport nomination: Ignoring issue#{id} pointing to a different git repository: Expected {0}, found {org_repo:?}",
+                event.repository.full_name
+            );
+            continue;
+        }
+        input.ids.push(id);
+    }
+
+    if input.ids.is_empty() || input.labels.is_empty() {
+        return Ok(None);
+    }
+
+    log::debug!(
+        "Will handle event action {:?} in backport. Regression IDs found {:?}",
+        event.action,
+        input.ids
+    );
+
+    Ok(Some(input))
+}
+
+pub(super) async fn handle_input(
+    ctx: &Context,
+    _config: &BackportConfig,
+    event: &IssuesEvent,
+    input: BackportInput,
+) -> anyhow::Result<()> {
+    let pr = &event.issue;
+
+    // Retrieve the issue(s) this pull request closes
+    let issues = input
+        .ids
+        .iter()
+        .copied()
+        .map(|id| async move { event.repository.get_issue(&ctx.github, id).await });
+    let issues = join_all(issues).await;
+
+    // Add backport nomination label to the pull request
+    for issue in issues {
+        if let Err(ref err) = issue {
+            log::warn!("Failed to get issue: {:?}", err);
+            continue;
+        }
+        let issue = issue.context("failed to get issue")?;
+        let issue_labels: Vec<&str> = issue.labels.iter().map(|l| l.name.as_str()).collect();
+
+        // Check issue for a prerequisite priority label
+        // If none, skip this issue
+        if !contains_any(&issue_labels, &PRIORITY_LABELS) {
+            continue;
+        }
+
+        // Get the labels to be added the PR according to the matching (required) regression label
+        // that is found in the configuration that this handler has received
+        // If no regression label is found, skip this issue
+        let add_labels = issue_labels.iter().find_map(|l| input.labels.get(*l));
+        if add_labels.is_none() {
+            log::warn!(
+                "Skipping backport nomination: nothing to do for issue #{}. No config found for regression label ({:?})",
+                issue.number,
+                REGRESSION_LABELS
+            );
+            continue;
+        }
+
+        // Add backport nomination label(s) to PR
+        let mut new_labels = pr.labels().to_owned();
+        new_labels.extend(
+            add_labels
+                .expect("failed to unwrap add_labels")
+                .iter()
+                .cloned()
+                .map(|name| Label { name }),
+        );
+        log::debug!(
+            "PR#{} adding labels for backport {:?}",
+            pr.number,
+            add_labels
+        );
+        let _ = pr
+            .add_labels(&ctx.github, new_labels)
+            .await
+            .context("failed to add backport labels to the PR");
+    }
+
+    Ok(())
+}
+
+fn contains_any(haystack: &[&str], needles: &[&str]) -> bool {
+    needles.iter().any(|needle| haystack.contains(needle))
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::handlers::backport::CLOSES_ISSUE_REGEXP;
+
+    #[tokio::test]
+    async fn backport_match_comment() {
+        let test_strings = vec![
+            ("close #10", vec![10]),
+            ("closes #10", vec![10]),
+            ("closed #10", vec![10]),
+            ("Closes #10", vec![10]),
+            ("close  #10", vec![10]),
+            ("close rust-lang/rust#10", vec![10]),
+            ("cLose: rust-lang/rust#10", vec![10]),
+            ("fix #10", vec![10]),
+            ("fixes #10", vec![10]),
+            ("fixed #10", vec![10]),
+            ("resolve #10", vec![10]),
+            ("resolves #10", vec![10]),
+            ("resolved #10", vec![10]),
+            (
+                "Fixes #20, Resolves #21, closed #22, LOL #23",
+                vec![20, 21, 22],
+            ),
+            ("Resolved #10", vec![10]),
+            ("Fixes #10", vec![10]),
+            ("Closes #10", vec![10]),
+        ];
+        for test_case in test_strings {
+            let mut ids: Vec<u64> = vec![];
+            let test_str = test_case.0;
+            let expected = test_case.1;
+            for caps in CLOSES_ISSUE_REGEXP.captures_iter(test_str) {
+                // eprintln!("caps {:?}", caps);
+                let id = &caps["issue_num"];
+                ids.push(id.parse::<u64>().unwrap());
+            }
+            // eprintln!("ids={:?}", ids);
+            assert_eq!(ids, expected);
+        }
+    }
+}
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -28,6 +28,7 @@ impl fmt::Display for HandlerError {
 
 mod assign;
 mod autolabel;
+mod backport;
 mod bot_pull_requests;
 mod check_commits;
 mod close;
@@ -225,6 +226,7 @@ macro_rules! issue_handlers {
 issue_handlers! {
     assign,
     autolabel,
+    backport,
     issue_links,
     major_change,
     mentions,


src/config.rs
```diff
@@ -522,6 +523,24 @@ fn default_true() -> bool {
...
 }
 
+#[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+pub(crate) struct BackportTeamConfig {
+    // Config identifier -> labels
+    #[serde(flatten)]
+    pub(crate) configs: HashMap<String, BackportConfig>,
+}
+
+#[derive(Default, PartialEq, Eq, Debug, serde::Deserialize)]
```
Urgau, line 533: Almost all of our fields use `kebab-case`, let's use that here as well.
```suggestion
#[derive(Default, PartialEq, Eq, Debug, serde::Deserialize)]
#[serde(rename_all = "kebab-case")]
```

src/config.rs
```diff
@@ -522,6 +523,24 @@ fn default_true() -> bool {
...
+#[serde(deny_unknown_fields)]
+pub(crate) struct BackportConfig {
+    /// Prerequisite label(s) (one of them) to trigger this handler for a specific team
+    pub(crate) team_labels: Vec<String>,
```
Urgau: I find the name `team_labels` quite misleading, it's not restricted to "team" labels, it can be any labels. What about `required_pr_labels`? That makes it clear that it's a list of required labels on the pull request.
```suggestion
    pub(crate) required_pr_labels: Vec<String>,
```

src/config.rs
```diff
@@ -522,6 +523,24 @@ fn default_true() -> bool {
...
+    /// Prerequisite label for an issue to qualify as regression
+    pub(crate) needs_label: String,
```
Urgau: Similar thing here, "needs_labels" doesn't indicate at all that it's for the issue.
```suggestion
    pub(crate) required_issue_labels: String,
```

src/config.rs
```diff
@@ -727,6 +746,7 @@ mod tests {
                 concern: Some(ConcernConfig {
                     labels: vec!["has-concerns".to_string()],
                 }),
+                backport: None
```
Urgau: Could you populate the toml above (this test data) or the one below and add the expected de-serialized output here. Mainly for sanity check.

src/handlers/backport.rs
```diff
@@ -0,0 +1,227 @@
...
+
+use crate::config::BackportTeamConfig;
+use crate::github::{IssuesAction, IssuesEvent, Label};
+use crate::handlers::Context;
+use regex::Regex;
+use tracing as log;
+
+// see https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue
+static CLOSES_ISSUE_REGEXP: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new("(?i)(close[sd]*|fix([e]*[sd]*)?|resolve[sd]*) #(\\d+)").unwrap());
```
Urgau: This doesn't take into account canonicalized issue links, ie `rust-lang/rust#123` which are norm now a days on rust-lang/rust. It also doesn't take into account the s pac es. It's also missing the `:` in `closes: #123`.

I took the liberty of "fixing" your regex.
```suggestion
    LazyLock::new(|| Regex::new("(?i)(?P<action>close[sd]*|fix([e]*[sd]*)?|resolve[sd]*)(?P<spaces>:? +)(?<org_repo>[a-zA-Z0-9_-]*\/[a-zA-Z0-9_-]*)?(?P<issue>#[0-9]+)").unwrap());
```
*Note that I named the capture groups so ease of read, can be removed if you don't like them.*

src/handlers/backport.rs
```diff
@@ -0,0 +1,227 @@
...
+            ("resolve #10", vec![10]),
+            ("resolves #10", vec![10]),
+            ("resolved #10", vec![10]),
+            (
+                "Fixes #20, Resolves #21, closed #22, LOL #23",
+                vec![20, 21, 22],
+            ),
+            ("Resolved #10", vec![10]),
+            ("Fixes #10", vec![10]),
+            ("Closes #10", vec![10]),
```
Urgau, line 237:
```suggestion
            ("Closes #10", vec![10]),
            ("close  #10", vec![10]),
            ("close rust-lang/rust#10", vec![10]),
            ("cLose: rust-lang/rust#10", vec![10]),
```

src/handlers/backport.rs
```diff
@@ -0,0 +1,227 @@
...
+        ];
+        for test_case in test_strings {
+            let mut ids: Vec<u64> = vec![];
+            let test_str = test_case.0;
+            let expected = test_case.1;
+            for caps in CLOSES_ISSUE_REGEXP.captures_iter(test_str) {
+                // println!("caps {:?}", caps);
+                let id = caps.get(3).unwrap().as_str();
```
Urgau:
```suggestion
                let id = caps["issue"].as_str();
```

src/handlers/backport.rs
```diff
@@ -0,0 +1,227 @@
...
+    // Retrieve the issue(s) this pull request closes
+    let issues = input
+        .ids
+        .iter()
+        .copied()
+        .map(|id| async move { event.repository.get_issue(&ctx.github, id).await });
+
+    // auto-nominate for backport only patches fixing high/critical regressions
+    // For `P_{medium,low}` regressions, let the author decide
+    let priority_labels = ["P-high", "P-critical"];
```
Urgau: I feel like this should be configurable instead of being hard-coded.

src/handlers/backport.rs
```diff
@@ -0,0 +1,227 @@
...
+                l.name.as_str()
+            })
+            .collect();
+
+        // Check issue for a prerequisite regression label
+        let regression_labels = [
+            "regression-from-stable-to-nightly",
+            "regression-from-stable-to-beta",
+            "regression-from-stable-to-stable",
+        ];
```
Urgau: Same with those, I think it we would be better if they were configurable, but that's not load bearing, we can merge without making them configurable at first.

src/handlers/backport.rs
```diff
@@ -0,0 +1,227 @@
...
+                .iter()
+                .cloned()
+                .map(|name| Label { name }),
+        );
+        log::debug!(
+            "PR#{} adding labels for backport {:?}",
+            pr.number,
+            new_labels
+        );
+        return pr.add_labels(&ctx.github, new_labels).await;
```
Urgau:
```suggestion
        pr.add_labels(&ctx.github, new_labels).await.context("failed to add backport labels to the PR")?;
```

src/handlers/backport.rs
```diff
@@ -0,0 +1,227 @@
...
+    _ctx: &Context,
+    event: &IssuesEvent,
+    config: Option<&BackportTeamConfig>,
+) -> Result<Option<BackportInput>, String> {
+    let config = match config {
+        Some(config) => config,
+        None => return Ok(None),
+    };
+
+    if !matches!(event.action, IssuesAction::Opened) && !event.issue.is_pr() {
```
Urgau: Contributors often add labels after opening the PR, we should probably also listen to "Edited" events, but that becomes expansive.

What about only triggering the logic when the PR is merged? That way we reduce the false-positives and false-negatives.

src/handlers/backport.rs
```diff
@@ -0,0 +1,227 @@
...
+        log::warn!(
+            "Skipping backport nomination: could not find a suitable backport config. Please ensure the triagebot.toml has a `[backport.*.team_labels]` section matching the team label(s) for PR #{}.",
+            pr.number
+        );
+        return Ok(None);
+    }
+
+    // Check marker text in the opening comment of the PR to retrieve the issue(s) being fixed
+    for caps in CLOSES_ISSUE_REGEXP.captures_iter(&event.issue.body) {
+        let id = caps.get(3).unwrap().as_str();
```
Urgau:
```suggestion
        let id = caps.name("issue").unwrap().as_str();
```

src/handlers/backport.rs
Urgau: I've seen contributor (probably my-self included using) `Fixes (after beta-backport) #123`). How do we want to take into account that extra marker?

src/handlers/backport.rs
```diff
@@ -0,0 +1,227 @@
...
+    if valid_configs.is_empty() {
+        log::warn!(
+            "Skipping backport nomination: could not find a suitable backport config. Please ensure the triagebot.toml has a `[backport.*.team_labels]` section matching the team label(s) for PR #{}.",
+            pr.number
+        );
+        return Ok(None);
+    }
+
+    // Check marker text in the opening comment of the PR to retrieve the issue(s) being fixed
+    for caps in CLOSES_ISSUE_REGEXP.captures_iter(&event.issue.body) {
```
Urgau, line 103: FWIW, capturing over the issue body without taking into account the markdown is insufficient as people might use `Fixes #123` and not expect us to trigger on it.

Preventing that is bit finicky so let's defer that.

src/handlers/backport.rs
apiraino: Yeah, I saw also Michael doing that. To my understanding it is to block GH autoclosing the issue, correct?

Your observation makes sense. But that syntax can be very aleatoric (`Fixes basically-anything-I-can-think-of #123`), I would probably think about it for a second before adding to the regexp :thinking: 

I'll make a note to discuss this in a second patch, would it be fine for you?

apiraino: I would just like to be sure avoiding introducing *false* triggers

src/handlers/backport.rs
apiraino: Hm ... in general I think that having more time to evaluate backporting a patch is better, therefore nominating it while the PR is in progress is a good thing. We live in different time zones and sometimes the "voting" time window is short due to various factors.

src/handlers/backport.rs
apiraino: For reference: do you have an example? So I can look at that later.

src/handlers/backport.rs
apiraino: Good idea and good suggestion! (fixed a small typo in the regexp)

Urgau: Sure, fine to address a second patch.

One thing I just think of, contributors put the extra marker to prevent GitHub from closing it too early but it's preventing us from seeing as well. What if we automatically added the extra marker after detecting that it's linked to a regression issue?

src/handlers/backport.rs
Urgau: Fine by me.

We should still however look at the "Edited" events, can be done later of-course.

src/handlers/backport.rs
apiraino: Was discussed with David and Wesley so I left a comment to explain the rationale. We think it's fine to start with just auto-nominating p-high and p-critical which are important. We can always change this if needed.

src/handlers/backport.rs
apiraino: Not completely against but these labels are basically a `const` in our workflows.

Urgau: My reasoning is that this handler is probably quite useful for other repos, like clippy and cargo. I would prefer not to specialize it too much against the compiler case.

src/handlers/backport.rs
apiraino: That's a clever idea :+1:  I'll look into the github webhook [pull request events](https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request) after I have fixed this branch; at first sight I don't see a specific event for "this PR was linked to an issue".

As you point out parsing all edits could be a bit expensive, we have TONS of pull requests, it's a hot path for the triagebot.

Urgau: > at first sight I don't see a specific event for "this PR was linked to an issue".

I don't think there is one, I'm pretty sure we'll have to listen to the "Edited" event, which might not be that bad if we are clever and only analyze the changes (that should avoid unnecessary checking of labels when the fixes hasn't changed).

Kobzol: You can get it from the Timeline API (https://docs.github.com/en/rest/issues/timeline?apiVersion=2022-11-28, https://docs.github.com/en/rest/using-the-rest-api/issue-event-types?apiVersion=2022-11-28#cross-referenced).

apiraino: (self-note) after a quick test in my test repository I think we can retrieve the cross reference event parsing like this

```sh
$ curl -L \
	-H "Accept: application/vnd.github+json" \
	-H "Authorization: Bearer $GITHUB_TOKEN" \
	-H "X-GitHub-Api-Version: 2022-11-28"	\
	https://api.github.com/repos/apiraino/test-triagebot/issues/110/timeline \
	| jq '.[].event="cross-referenced"' \
	| jq .[].source.issue.number
```

Translated:
- filter all events of type `cross-referenced`
- filter the one element that has a `source` -> `issue` -> `number`

apiraino: Can we experiment with this in a follow-up patch?

Urgau: I'm not sure how finding the cross-references with the Timeline api helps (since it's not a webhook), but sure, we can do that later.

apiraino: > I don't think there is one, I'm pretty sure we'll have to listen to the "Edited" event, which might not be that bad if we are clever and only analyze the changes (that should avoid unnecessary checking of labels when the fixes hasn't changed).

Just to be sure I understand: by "edit" events you mean all the comments added to the pull request, looking for the marker indicating that it fixes a regression. Correct?

Urgau: No, I'm talking about the edit of the PR description, which (I think) are observed from the "Edited" event.

apiraino: Ah Ok, then maybe I'm just overthinking it. I don't expect the opening comment to be edited *that* often. I'll just handle `IssuesEvent::{Opened|Edited}` :+1: 

When editing a comment the events are different (`IssueCommentEvent`).

src/handlers/backport.rs
apiraino: (I moved this out of the for loop and fixed the check)

src/handlers/backport.rs
```diff
@@ -0,0 +1,241 @@
...
+) -> Result<Option<BackportInput>, String> {
+    let config = match config {
+        Some(config) => config,
+        None => return Ok(None),
+    };
+
+    // Only handle events when the PR is opened or the first comment is edited
+    if !matches!(event.action, IssuesAction::Opened | IssuesAction::Edited) && !event.issue.is_pr()
+    {
+        log::info!(
```
Urgau: `info!` seems a bit much, as it will be triggered for every event, let's downgrade it to debug

```suggestion
        log::debug!(
```

src/handlers/backport.rs
```diff
@@ -0,0 +1,241 @@
...
+    event: &IssuesEvent,
+    config: Option<&BackportTeamConfig>,
+) -> Result<Option<BackportInput>, String> {
+    let config = match config {
+        Some(config) => config,
+        None => return Ok(None),
+    };
+
+    // Only handle events when the PR is opened or the first comment is edited
+    if !matches!(event.action, IssuesAction::Opened | IssuesAction::Edited) && !event.issue.is_pr()
```
Urgau:
```suggestion
    if !matches!(event.action, IssuesAction::Opened | IssuesAction::Edited) || !event.issue.is_pr()
```

src/handlers/backport.rs
```diff
@@ -0,0 +1,241 @@
...
+                .profile_labels
+                .insert(cfg.required_issue_label.clone(), cfg.add_labels.clone());
+            true
+        })
+        .collect();
+    if valid_configs.is_empty() {
+        log::warn!(
+            "Skipping backport nomination: could not find a suitable backport config. Please ensure the triagebot.toml has a `[backport.*.required_pr_labels]` section matching the team label(s) for PR #{}.",
+            pr.number
+        );
```
Urgau: Is this supposed to be a user-facing warning? Given how it's written, it feels like it.

src/handlers/backport.rs
```diff
@@ -0,0 +1,241 @@
...
+    for caps in CLOSES_ISSUE_REGEXP.captures_iter(&event.issue.body) {
+        let id = caps.name("issue_num").unwrap().as_str();
+        let id = match id.parse::<u64>() {
+            Ok(id) => id,
+            Err(err) => {
+                return Err(format!("Failed to parse issue id `{id}`, error: {err}"));
+            }
+        };
+        input.ids.push(id);
+    }
```
Urgau, line 125: We probably want a early-exit here, to avoid unnecessary work later.

```suggestion
    }
    
    if input.ids.is_empty() || input.profile_labels.is_empty() {
        return Ok(None);
    }
```

src/handlers/backport.rs
```diff
@@ -0,0 +1,241 @@
...
+    input: BackportInput,
+) -> anyhow::Result<()> {
+    let pr = &event.issue;
+
+    // Retrieve the issue(s) this pull request closes
+    let issues = input
+        .ids
+        .iter()
+        .copied()
+        .map(|id| async move { event.repository.get_issue(&ctx.github, id).await });
```
Urgau, line 153: Futures don't do anything unless polled, we can either just retrive them one after the other or use something like `let issues = join_all(issues).await;` from `use futures::future::join_all;`.

src/handlers/backport.rs
```diff
@@ -0,0 +1,241 @@
...
+                    regression_label = l.name.clone();
+                }
+                l.name.as_str()
+            })
+            .collect();
+
+        // Check issue for a prerequisite regression label
+        if regression_label.is_empty() || !contains_any(&REGRESSION_LABELS, &[&regression_label]) {
+            return Ok(());
+        }
```
Urgau: Let's avoid hard-coded knowledge.
```suggestion
        let mut regression_label = None;
        let issue_labels: Vec<&str> = issue
            .labels
            .iter()
            .map(|l| {
                // save regression label for later
                if regression_label.is_none() && let Some(label) = REGRESSION_LABELS.iter().find(|rl| rl == l) {
                    regression_label = Some(label);
                }
                l.name.as_str()
            })
            .collect();

        let Some(regression_label) = regression_label else {
            continue;
        };
```

src/handlers/backport.rs
```diff
@@ -0,0 +1,241 @@
...
+
+        // Check issue for a prerequisite priority label
+        if !contains_any(&issue_labels, &PRIORITY_LABELS) {
+            return Ok(());
```
Urgau: I don't think we want to skip the other issues.
```suggestion
            continue;
```

src/handlers/backport.rs
```diff
@@ -0,0 +1,241 @@
...
+
+        // figure out the labels to be added according to the regression label
+        let add_labels = input.profile_labels.get(&regression_label);
+        if add_labels.is_none() {
+            log::warn!(
+                "Skipping backport nomination: nothing to do for issue #{}. No config found for regression label ({:?})",
+                issue.number,
+                REGRESSION_LABELS
+            );
+            return Ok(());
```
Urgau:
```suggestion
            continue;
```

src/handlers/backport.rs
```diff
@@ -0,0 +1,241 @@
...
+                .iter()
+                .cloned()
+                .map(|name| Label { name }),
+        );
+        log::debug!(
+            "PR#{} adding labels for backport {:?}",
+            pr.number,
+            new_labels
+        );
+        return pr
```
Urgau:
```suggestion
        pr
```

src/handlers/backport.rs
Urgau: Btw, why are we even trying to find a regression label? Shouldn't look at if any of the issue labels match the required issue label (now in `profile_labels`)?

In other words,
```rust
    let add_labels = issue_labels.iter().find_map(|l| input.profile_labels.get(&l));
```

src/handlers/backport.rs
```diff
@@ -0,0 +1,241 @@
...
+                issue.number,
+                REGRESSION_LABELS
+            );
+            return Ok(());
+        }
+
+        // Add backport nomination label(s) to PR
+        let mut new_labels = pr.labels().to_owned();
+        new_labels.extend(
+            add_labels
```
Urgau, line 187: No need to do that, the `add_labels` call below already appends to the existing list of labels. You can just pass the `add_labels`.

src/handlers/backport.rs
apiraino: Ahh booleans got me again, thanks!

src/handlers/backport.rs
apiraino: I thought to just log this case to be visible to triagebot maintainers in case someone comes to us and says *"why wasn't my patch auto-nominated for backport?"*. Since this is an automated task I wouldn't know to who and where to send this warning message. Do you have opinions or suggestions?

I realize the message itself is not great. I tried hard looking for the right words :D

src/handlers/backport.rs
apiraino: correct, they are polled in the following `for` iteration: what does it improve if they're all run immediately?

src/handlers/backport.rs
apiraino: ~~Why `continue` here instead of returning?~~

(ah ok we are iterating a list)

src/handlers/backport.rs
apiraino: The function `add_labels()` wants a Vec<Label>`. The variable `add_labels` is a `Vec<String>` (and I'd prefer leaving it like that for ease of handling)

src/handlers/backport.rs
apiraino: You're right, I can skip reduce the amount of work here

src/handlers/backport.rs
Urgau: I think it fine, even as written. I was just a bit surprised to see what looked like a user-message being only logged. Let's leave it that way.

src/handlers/backport.rs
Urgau: The issue could have a unlimited amount of linked-issues (max I've seen is 5) and we only have maximum of 10s per webhook, and network IO can be unreliable, so the maximum we can do concurrently we should do it.

src/handlers/backport.rs
apiraino: oh wait maybe you mean people actually writing in the comment:
```
`Fixes #123`
```
to avoid the GitHub automated issue cross-reference?

src/handlers/backport.rs
```diff
@@ -0,0 +1,238 @@
...
+    let issues = input
+        .ids
+        .iter()
+        .copied()
+        .map(|id| async move { event.repository.get_issue(&ctx.github, id).await });
+    let issues = join_all(issues).await;
+
+    // Add backport nomination label to the pull request
+    for issue in issues {
+        let issue = issue.expect("Failed to get issue from iterator");
```
Urgau: We shouldn't panic if there was an error about an issue, that would take down the other handlers.

```suggestion
        let issue = issue.context("failed to get issue")?;
```

apiraino: good point. I logged a warning instead of blocking the iteration

src/handlers/backport.rs
Urgau: Yes, I'm talking about those cases. We have some (kind-of) complex logic to handle all kind of markdown input in https://github.com/rust-lang/triagebot/blob/master/parser/src/ignore_block.rs, we unfortunately don't have a simple wrapper for regex.

It's also not very common, so I'm fine deferring it until someone complains.

src/config.rs
```diff
@@ -522,6 +523,25 @@ fn default_true() -> bool {
     true
 }
 
+#[derive(PartialEq, Eq, Debug, serde::Deserialize)]
+pub(crate) struct BackportTeamConfig {
```
Urgau: Didn't noticed it before but it's not very consistent for the the config struct to not be named in the form `<handler-name>Config`, even more so that it's children is actually named the "right" way.

What about `BackportConfig` here and `BackportRuleConfig` for the childrens? (to avoid `ConfigConfig`)

src/handlers/backport.rs
```diff
@@ -0,0 +1,242 @@
...
+    event: &IssuesEvent,
+    config: Option<&BackportTeamConfig>,
+) -> Result<Option<BackportInput>, String> {
+    let config = match config {
+        Some(config) => config,
+        None => return Ok(None),
+    };
+
+    // Only handle events when the PR is opened or the first comment is edited
+    if !matches!(event.action, IssuesAction::Opened | IssuesAction::Edited) || !event.issue.is_pr()
```
Urgau: We just had an "incident" about those exclusion conditions in https://github.com/rust-lang/triagebot/pull/2110, and since we'll probably want to handle more events in the future, let's make it better from the start.


```suggestion
    // Only handle events when the PR is opened or the first comment is edited
    let should_check  matches!(event.action, IssuesAction::Opened | IssuesAction::Edited);
    
    if !should_check || !event.issue.is_pr()
```

src/handlers/backport.rs
```diff
@@ -0,0 +1,242 @@
...
+    let valid_configs: Vec<_> = config
+        .configs
+        .iter()
+        .clone()
+        .filter(|(_cfg_name, cfg)| {
+            let required_pr_labels: Vec<&str> =
+                cfg.required_pr_labels.iter().map(|l| l.as_str()).collect();
+            if !contains_any(&pr_labels, &required_pr_labels) {
+                log::warn!(
+                    "Skipping backport nomination: PR is missing one required team label: {:?}",
```
Urgau:
```suggestion
                    "Skipping backport nomination: PR is missing one required PR label: {:?}",
```

src/handlers/backport.rs
```diff
@@ -0,0 +1,242 @@
...
+            pr.number
+        );
+        return Ok(None);
+    }
+
+    // Check marker text in the opening comment of the PR to retrieve the issue(s) being fixed
+    for caps in CLOSES_ISSUE_REGEXP.captures_iter(&event.issue.body) {
+        let id = caps
+            .name("issue_num")
+            .expect("Failed to cast issue_num to string")
```
Urgau:
```suggestion
            .with_context(|| format!("failed to get issue_num from {caps:?}"))?
```

src/config.rs
apiraino: I'm totally fine with any naming scheme you prefer :+1:

src/handlers/backport.rs
apiraino: I don't fully understand this suggestion but I'm fine

src/handlers/backport.rs
apiraino: Does this change actually compile for you? I can't make it compile and it seems to be a bit fiddly to make it work.

<details><summary>Details</summary>
<p>

```

error[E0277]: `?` couldn't convert the error to `std::string::String`
   --> src/handlers/backport.rs:106:79
    |
106 |             .with_context(|| format!("failed to get issue_num from {caps:?}"))?;
    |              -----------------------------------------------------------------^ the trait `From<anyhow::Error>` is not implemented for `std::string::String`
    |              |
    |              this can't be annotated with `?` because it has type `Result<_, anyhow::Error>`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait

```

</p>
</details>

Urgau: I didn't try it locally but I'm surprised it doesn't work.

Looking around, I think it's due to `handle_input` not returning `anyhow::Result` but std-`Result`.

apiraino: I changed it a little bit to make the intent clearer  (if the parsing fails, don't exit but delegate the u64 cast to check if there is an error)

src/handlers/backport.rs
```diff
@@ -0,0 +1,243 @@
...
+            .ok_or_else(|| format!("failed to get issue_num from {caps:?}"))?
+            .as_str();
+
+        let id = match id.parse::<u64>() {
+            Ok(id) => id,
+            Err(err) => {
+                return Err(format!("Failed to parse issue id `{id}`, error: {err}"));
+            }
+        };
+        input.ids.push(id);
```
Urgau, line 124: This is not handling cases where there is a an org/repo: `rust-lang/cargo` or whatever. We should either reject those for now (I think we should do that) or handle them here and in the `get_issue` below.


```suggestion
        if let Some(org_repo) = caps.name("org_repo") {
            log::info!("ignoring issue {id} bc of repo: {org_repo:?}");
            continue;
        }
        
        input.ids.push(id);
```

Urgau:
Modulo the last nit, this looks good to me. Thanks for putting it through my reviews.

Next step (preferably before merging) will be to open a PR against the forge to add documentation for this feature, with a description of what it does, it's currently limitations, how to activate it, ... (take inspiration from the current pages).

Regarding deployment, do you want to go straight to "prod" in rust-lang/rust or do you want to use this repo as a small playground before using it rust-lang/rust?

apiraino:
> Will merge the PR after you format the code (to fix CI).

Oh, I didn't notice the CI rustfmt failure. In this repo we have a config file that tells rustfmt to use edition 2018 but I *think* the CI is using the 2024 edition formatting options:
```
$ rustfmt --check --edition 2018 backport.rs

$ rustfmt --check --edition 2024 backport.rs
Diff in backport.rs:115:
         if let Some(org_repo) = caps.name("org_repo")
             && org_repo.as_str() != event.repository.full_name
         {
-            log::info!("Skipping backport nomination: Ignoring issue#{id} pointing to a different git repository: Expected {0}, found {org_repo:?}", event.repository.full_name);
+            log::info!(
+                "Skipping backport nomination: Ignoring issue#{id} pointing to a different git repository: Expected {0}, found {org_repo:?}",
+                event.repository.full_name
+            );
             continue;
         }
         input.ids.push(id);
```

Hmm is there a discrepancy between the `rustfmt.toml` and the CI? :thinking:

Urgau:
Let's go.

PR #2093, opened 2025-07-01
Handle base branch changed in `[autolabel]` and check-commits

Urgau:
This PR modifies the `[autolabel]` and check-commits handler to also trigger them-selves when the base branch has been changed (eg. `master` -> `beta`).

Example of the current misbehaving at https://github.com/rust-lang/rust/pull/143275

Based on https://github.com/rust-lang/triagebot/pull/2091, only last commit relevant.


src/handlers/autolabel.rs
```diff
@@ -20,21 +20,26 @@ pub(super) async fn parse_input(
...
         event.action,
         IssuesAction::Opened
             | IssuesAction::Closed
             | IssuesAction::Reopened
             | IssuesAction::Synchronize
             | IssuesAction::ReadyForReview
             | IssuesAction::ConvertedToDraft
-    ) {
+    );
+    let has_base_changed = matches!(event.action, IssuesAction::Edited)
```
Kobzol: Can we add this as a method on `event`?

Urgau: Sure, done.

Kobzol (approved):

PR #2095, opened 2025-07-04
Add self-referential anchors to every timestamps in our GHA logs viewer

Urgau:
This PR adds self-referential anchors (`<a>`) to every timestamps at the start of the logs line in our GHA logs viewer.

This meant to be able easily to share a position in the logs.

To not be too distracting I made it so the anchor are only visible when hovering the timestamps.

image

This was asked at [#t-compiler > Enhanced plain logs with rla @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Enhanced.20plain.20logs.20with.20rla/near/527031952)

### DIFF
--- a/src/gha_logs.rs
+++ b/src/gha_logs.rs
@@ -129,7 +129,7 @@ async fn process_logs(
     let nonce = Uuid::new_v4().to_hyphenated().to_string();
 
     let html = format!(
-        r#"<!DOCTYPE html>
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Ooh, that's clever, and we don't even need line numbers! Cool.

PR #2097, opened 2025-07-05
Improve automatic GHA logs scrolling for desktop and mobile

Urgau:
This PR improves the automatic GHA logs scrolling for desktop and mobile by:
 1. putting the error marker at the bottom of the viewport -15vh
     - (instead of being at the center and potentially missing part of the error)
 2. by masking by default the timestamps on mobile/small viewport

This should improve the scrolling by making it more likely to be at the right place from the start without having the user needing to adjust it.

| Mobile | Desktop - 1 | Desktop - 2 |
|--------|-------|---|
| image | Screenshot 2025-07-06 at 00-24-38 rust-lang_rust$43666277064 - triagebot | Screenshot 2025-07-06 at 00-24-53 rust-lang_compiler-builtins$45387265068 - triagebot |

### DIFF
--- a/src/gha_logs.rs
+++ b/src/gha_logs.rs
@@ -150,6 +150,9 @@ async fn process_logs(
         .timestamp:hover {{
             text-decoration: underline;
         }}
+        .error-marker {{
+            scroll-margin-bottom: 15vh;
+        }}
     </style>
     <script type="module" nonce="{nonce}">
         import {{ AnsiUp }} from '{ANSI_UP_URL}'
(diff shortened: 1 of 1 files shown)


src/gha_logs.rs
```diff
@@ -183,9 +186,11 @@ async fn process_logs(
         
         // 6. If no anchor is given, scroll to the last error
         if (location.hash === "" && errorCounter >= 0) {{
+            const hasSmallViewport = window.innerWidth <= 750;
             document.getElementById(`error-${{errorCounter}}`).scrollIntoView({{
                 behavior: 'smooth',
```
Kobzol: Maybe we could also make this "instant"? I don't think the smooth scrolling gets us anything, and it takes a bit of time.

Urgau: I agree that on slow devices (mobile phones) it doesn't bring us anything as we scroll to a part which isn't rendered yet.

On desktop I'm a bit more ambivalent, I think it's makes it clear that are a bunch of logs before that one, but it does take a bit of time which isn't strictly necessary.

So changed it to `instant`.

Kobzol (approved):

Issue #2099, opened 2025-07-07
Major change handler doesn't take into account resolved-concerns between second and acceptance time


Urgau:
Seen at https://github.com/rust-lang/compiler-team/issues/893#issuecomment-3043394002, where the last resolved concerns was 3 days before the acceptance instead of the 10 days.

We should either remove the acceptance job (when the first concern is raised) or look at the events and discard the job as the concern label was set in between the second and now.

PR #2100, opened 2025-07-07
Cancel previous acceptance job when concerns are added on major change

Urgau:
This PR adds the necessary logic to retrieve the job ID when scheduling a one-off job so we can store it after scheduling the major change acceptance job which is then used to cancel the acceptance when concerns are added on major change.

I haven't tested the updated SQL query, it looks fine to me, but who knows.

Fixes https://github.com/rust-lang/triagebot/issues/2099

### DIFF
--- a/src/handlers/major_change.rs
+++ b/src/handlers/major_change.rs
@@ -11,6 +11,7 @@ use crate::{
 use anyhow::Context as _;
 use async_trait::async_trait;
 use chrono::{DateTime, Duration, Utc};
+use futures::TryStreamExt as _;
 use parser::command::second::SecondCommand;
 use serde::{Deserialize, Serialize};
 use tracing as log;
@@ -452,13 +453,27 @@ enum SecondedLogicError {
         accept_at: DateTime<Utc>,
         now: DateTime<Utc>,
     },
+    IssueStateChanged {
+        at: DateTime<Utc>,
+        draft: bool,
+        open: bool,
+    },
     IssueNotReady {
         draft: bool,
         open: bool,
     },
-    NotAMajorChange,
+    EnablingLabelAbsent,
+    EnablingLabelRemoved {
+        at: DateTime<Utc>,
+    },
     SecondLabelAbsent,
+    SecondLabelRemoved {
+        at: DateTime<Utc>,
+    },
     ConcernsLabelPresent,
+    ConcernsLabelAdded {
+        at: DateTime<Utc>,
+    },
     NoMajorChangeConfig,
 }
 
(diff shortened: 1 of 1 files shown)


Urgau:
Another idea might be to use the Timeline/Events API inside the job and see if there is an interesting event that happened between the acceptance and the job running.

I feel like this would be easier as we would only need to check things once (in the job) and we wouldn't need to have the logic for rembering the job and canceling it.

I think I want to try that instead, what do you think?

Kobzol:
I think that this job spawning is running into race condition limits that I'm not sure we're able to handle in triagebot. I think that a more robust approach would just be to run a "cron" job e.g. every hour or so, which would check all open MCPs, and accept those that have been waiting after second without any concerns for N days.

Urgau:
I don't see how that changes much, with a cron job we would still need to look at the Timeline api to see if there was any concerns or issue closed or whatever. At which point the cron job doesn't change much from the one-off jobs.

> I think that this job spawning is running into race condition limits that I'm not sure we're able to handle in triagebot.

I don't see how we would have a race if used the Timeline API on the one-off job, we would be able to see that the concern label was added after the second, preventing the acceptance.

(The only race I could see with that is between the Timeline api call and us deciding to close it, which would be incredibly unlucky, less than 1s, which would also happen with the cron job)

Kobzol:
Ok, if you want to do that, feel free to! Happy to take a look. I never used the timeline API, so I'm interested in seeing what it can do (I have other use-cases for it).

Urgau:
Yes, that seems right.

src/handlers/major_change.rs
```diff
@@ -568,24 +598,87 @@ async fn process_seconded(
...
+                } else if label.name == config.second_label {
+                    anyhow::bail!(SecondedLogicError::SecondLabelRemoved { at });
+                }
+            } else if event.event == Event::Labeled {
+                let label = event.label.context("labeled event without label")?;
+
+                if Some(&label.name) == config.concerns_label.as_ref() {
+                    anyhow::bail!(SecondedLogicError::ConcernsLabelAdded { at })
+                }
+            } else if event.event == Event::Closed || event.event == Event::ConvertToDraft {
```
Kobzol, line 650: MCPs cannot be on PRs, right? So `ConvertToDraft` shouldn't happen (it cannot be used on issues). If it was possible to convert to draft, we wouldn't ever restart the job anyway after it would be undrafted.

Something similar would happen if someone closed and then reopened the issue though, the job wouldn't be restarted.

Urgau: > MCPs cannot be on PRs, right?

We don't currently do that, but I don't think there is anything preventing it.

To me converted to draft and closing are hard stop, I don't think we want a re-opened to restart the clock (and currently it restart it MCP from the start anyway), same for draft->ready.

Kobzol: Ok, fair enough.

Kobzol (approved):

Issue #2102, opened 2025-07-08
Edge case: draft PR marked as S-waiting-on-author receives S-waiting-on-review on undraft in addition instead of flip


jieyouxu:
### Summary

Un-drafting a PR should cause existing `S-waiting-on-author` label to be replaced by `S-waiting-on-review` and not have both co-exist.

Example: https://github.com/rust-lang/rust/pull/142357

### Repro steps

- Open non-draft PR (no `S-waiting-on-*` labels by default)
- `@rustbot author` -> PR receives `S-waiting-on-author`
- Mark PR as draft
- Undraft PR -> PR receives `S-waiting-on-review` but `S-waiting-on-author` isn't removed.

### Current behavior

Both `S-waiting-on-{author,review}` labels are present.

### Expected behavior

Only one of `S-waiting-on-{author,review}` labels are present; should become `S-waiting-on-review` on un-draft.

apiraino:
> Undraft PR -> PR receives S-waiting-on-review but S-waiting-on-author isn't removed.

Do we have a check on labels so that these two labels are mutually exclusive? Would that solve the issue at the root?

jieyouxu:
> > Undraft PR -> PR receives S-waiting-on-review but S-waiting-on-author isn't removed.
> 
> Do we have a check on labels so that these two are mutually exclusive? Would that solve the issue at the root?

Another edge case is that you can manually apply labels, so it doesn't have to be *perfect*

apiraino:
> Another edge case is that you can manually apply labels

Both of them at the same time or did you mean something else? (I'm sorry maybe I didn't understand your comment)

PR #2104, opened 2025-07-08
Add `new_draft` kind of labels in `[autolabel]`

Urgau:
This PR adds a new kind `new_draft` for labels in `[autolabel]`, this is so that we can handle `S-waiting-on-author` in a similar way to `S-waiting-on-review` (ie add it when the PR is in draft state and remove it when it's not).

```toml
[autolabel."S-waiting-on-author"]
new_draft = true
```

Fixes https://github.com/rust-lang/triagebot/issues/2102

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -282,6 +282,8 @@ pub(crate) struct AutolabelLabelConfig {
     pub(crate) new_pr: bool,
     #[serde(default)]
     pub(crate) new_issue: bool,
+    #[serde(default)]
+    pub(crate) new_draft: bool,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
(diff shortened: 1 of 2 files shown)


Urgau:
> we would say that label A is mutually exclusive with label B

I don't think it would necessarily help as removing a label (e.g. `S-waiting-on-review`) wouldn't add `S-waiting-on-author`, only adding `S-waiting-on-author` would remove `S-waiting-on-review`, right?

I say this as we have many exclusive labels around `S-waiting-on-*`.

In any case let's merge this as I think it's useful even if we later add an exclusive mechanism.

jieyouxu:
FWIW, I don't think it's possible to enforce exclusive labels, because can't you manually apply both labels at the same time?

Issue #2108, opened 2025-07-10
Duplicate pings from the `ping-goals` command


tomassedovic:
Calling e.g. `ping-goals` to ask every Goal owner to send an update to the tracking issue often results in duplicate (triplicate or even more) pings one after the other.

Repro:
* Send `@triagebot ping-goals 15 July-20` (just once)
* Wait for all the pings to be sent
* Look at the amount of messages each relevant Zulip topic receives.

Expected: each Goal-related topic receives exactly one message.

Actual: each topic receives anywhere between 2-4 messages.

See e.g. here:

[#project-goals > Prepare const traits for stabilization (goals#106)](https://rust-lang.zulipchat.com/#narrow/channel/435869-project-goals/topic/Prepare.20const.20traits.20for.20stabilization.20.28goals.23106.29/with/528094383)

Also, this is what triagebot told me after I sent that command:

```
Failed to await at this time: rust-lang/rust-project-goals failed to get repo

Caused by:
0: error sending request
1: client error (SendRequest)
2: connection closed before message completed
```

Urgau:
Looking at the logs can see 4 webhook coming to our `/zulip-hook` at ~5s interval.

```
2025-07-10T14:21:08.234041Z  INFO request{uuid=34e7e1fc-9b8c-492c-8792-eb93fcc2173e}: triagebot: request = Request { method: POST, uri: /zulip-hook, ..., body: Body(Streaming) }
2025-07-10T14:21:13.480020Z  INFO request{uuid=8b85d205-8153-4f7b-aeac-a545f833b1ca}: triagebot: request = Request { method: POST, uri: /zulip-hook, ..., body: Body(Streaming) }
2025-07-10T14:21:18.707339Z  INFO request{uuid=89d0c318-b008-4831-99ac-20af7303ddb1}: triagebot: request = Request { method: POST, uri: /zulip-hook, ..., body: Body(Streaming) }
2025-07-10T14:21:23.957444Z  INFO request{uuid=7bddf336-bf1f-45b2-95a4-424b806c790d}: triagebot: request = Request { method: POST, uri: /zulip-hook, ..., body: Body(Streaming) }
```

Only that last one seems to have a reply (with the error seen above).
```
[...]

PR #2109, opened 2025-07-10
Execute `ping-goals` command in the background to avoid timeout and infinite retries

Urgau:
This PR updates the handling of the `ping-goals` command to be executed in the background to avoid timeout and infinite retries from Zulip side, which resulted in multiple pings being sent to people.

cc @tomassedovic
Fixes https://github.com/rust-lang/triagebot/issues/2108

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -168,7 +168,7 @@ async fn serve_req(
         return Ok(Response::builder()
             .status(StatusCode::OK)
             .header("Content-Type", "application/json")
-            .body(Body::from(triagebot::zulip::respond(&ctx, req).await))
+            .body(Body::from(triagebot::zulip::respond(ctx, req).await))
             .unwrap());
     }
     if req.uri.path() != "/github-hook" {
--- a/src/main.rs
+++ b/src/main.rs
@@ -151,6 +151,7 @@ async fn serve_req(
             payload.extend_from_slice(&chunk);
         }
 
+        log::info!("/zulip-hook request body: {:?}", str::from_utf8(&payload));
         let req = match serde_json::from_slice(&payload) {
             Ok(r) => r,
             Err(e) => {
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -151,6 +151,7 @@ async fn serve_req(
             payload.extend_from_slice(&chunk);
         }
 
+        log::info!("/zulip-hook request body: {:?}", str::from_utf8(&payload));
```
Kobzol, line 154: Not that it matters too much, but why log and not tracing? And also maybe this should be `debug!`.

Kobzol (approved):
`tokio::spawn` has its issues, but it's the simplest thing to do here, and I don't think that we have to handle backpressure or do anything more complicated. So :+1:, let's try this.

src/main.rs
Urgau: It's the same thing, we just renamed it for some reason. https://github.com/rust-lang/triagebot/blob/1cdce7513d8a9ee4d3f5c1573b78fd4db7194c69/src/main.rs#L13

Urgau: > And also maybe this should be `debug!`.

We do log at `info!` the input of the github hook, so I wanted it to be the same.

src/zulip.rs
```diff
@@ -262,29 +270,58 @@ async fn handle_command<'a>(
...
+                    tracing::error!("ping_project_goals_owners: {err:?}");
+                    format!("ERROR\n\n```\n{err:#?}\n```\n")
+                }
+            };
+
+            let res = MessageApiRequest {
+                recipient: Recipient::Private {
+                    id: message.sender_id,
+                    email: &message.sender_email,
+                },
```
Urgau: This is not actually right, the command can be sent from both private message and stream. I need to change that.

Kobzol: Good catch, I forgot that, I find it kind of weird that it can be sent from both, but it is indeed supported.

Urgau: Fixed.

PR #2110, opened 2025-07-11
Fix check-commits trigger to not require base change

Urgau:
We should not required a PR base change, we should only allow it to check the commits.

Found when looking at why https://github.com/rust-lang/rust/pull/143794 didn't get the beta branch warning.


Kobzol (approved):
Heh, I remember staring hard at this and I still missed this "obvious" bug. Complex conditions are.. complex :)

PR #2111, opened 2025-07-12
Extract check-commits handler trigger logic and add tests for it

Urgau:
This PR extracts the check-commits super-handler trigger logic and add tests for it, so we can make sure it doesn't regress accidentally.

Related to https://github.com/rust-lang/triagebot/pull/2110

### DIFF
--- a/src/handlers/check_commits.rs
+++ b/src/handlers/check_commits.rs
@@ -39,25 +39,42 @@ struct CheckCommitsState {
     last_labels: Vec<String>,
 }
 
-pub(super) async fn handle(ctx: &Context, event: &Event, config: &Config) -> anyhow::Result<()> {
-    let Event::Issue(event) = event else {
-        return Ok(());
-    };
+fn should_handle_event(event: &IssuesEvent) -> bool {
+    // Reject non-PR
+    if !event.issue.is_pr() {
+        return false;
+    }
 
-    let should_check = matches!(
+    // Reject rollups and draft pr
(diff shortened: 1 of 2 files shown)


src/handlers/check_commits.rs
```diff
@@ -350,3 +367,139 @@ r#":warning: **Warning** :warning:
      - 123456789"#
     );
 }
+
+#[cfg(test)]
```
Kobzol: Maybe wrap this whole thing into `#[cfg(test)] mod tests { ... }`, as we usually do?

src/handlers/check_commits.rs
```diff
@@ -350,3 +367,139 @@ r#":warning: **Warning** :warning:
...
+fn make_opened_issues_event() -> IssuesEvent {
```
Kobzol:
```suggestion
fn make_opened_pr_event() -> IssuesEvent {
```
Would be a bit easier to understand, because most tests actually test the PR stuff, and it was weird seeing a function called "create issue" called, and then we modified it to actually create an issue (in `test_not_pr`) :)

Kobzol (approved):
Thanks, this is very nice. Left two nits, but feel free to merge as you want.

src/handlers/check_commits.rs
Urgau: `issues_event` actually referenced the name of the struct `IssuesEvent`, but I can see the confusion, rename it to `make_opened_pr_event` as suggested.

PR #2115, opened 2025-07-17
make matching case-insensitive on zulip

jdonszelmann:
I often use `work show` from my phone, which automatically capitalizes the first letter in my sentence. I see no reason why we would care about case in these commands, so just converting everything to lowercase shouldn't make a difference to most people while it makes the bot do the right thing more often.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -476,9 +476,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.39"
+version = "4.5.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fd60e63e9be68e5fb56422e397cf9baddded06dae1d2e523401542383bc72a9f"
+checksum = "1c1f056bae57e3e54c3375c41ff79619ddd13460a17d7438712bd0d83fda4ff8"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -486,9 +486,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.5.39"
+version = "4.5.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "89cc6392a1f72bbeb820d71f32108f61fdaf18bc526e1d23954168a67759ef51"
+checksum = "b3e7f4214277f3c7aa526a59dd3fbe306a370daee1f8b7b8c987069cd8e888a8"
 dependencies = [
  "anstream",
  "anstyle",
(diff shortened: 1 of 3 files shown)


jdonszelmann:
let me know if this makes sense, I thought I wouldn't open an issue since it's such a small change, just making the changes and discussing it here makes more sense to me :)

Urgau:
I don't think we currently have any zulip commands that cares about it's argument capitalization, so this would be technically fine.

I still wonder if it wouldn't be better to only lowercase the very first letter of the command.

cc @Kobzol

jdonszelmann:
We currently do not, though I just realized that commands that take parameters probably shouldn't get their parameter lowercased. So maybe we should also only do the first word in the arguments, or only do so after the unchanged version failed to parse

Urgau:
> So maybe we should also only do the first word in the arguments, or only do so after the unchanged version failed to parse

Both make sense to me. The first one would be simpler to implement.

Urgau:
I didn't knew about that setting, it indeed seems much better.

Kobzol:
Sadly AFAIK it cannot be currently applied to the whole command, so we need to add it to each argument separately...

jdonszelmann:
I looked for that but didn't find it earlier. Would be nice to apply globally ye...

jdonszelmann:
Am interested, just didn't have time. Go right ahead and feel free to r me for review?

jdonszelmann:
Mhm, but most people don't use command line parsers as a bot interface. I do like doing this though, but it makes quite some sense for bot commands to be case insensitive. I think using to lower makes most sense, or otherwise a pr to clap :shrug:

Urgau:
> I think using to lower makes most sense

That's what I meant (us ignoring the casing).

> otherwise a pr to clap

That could do it as well, but it's more involved.

Urgau:
After some back and forth the clap maintainer in https://github.com/clap-rs/clap/issues/6097 about adding an ignore case for commands, clap maintainer is proposing a workaround where we add aliases for each commands with the first letter upper-cased.

I think only having the first letter upper-cased would still work for us given how mobile phone keyboards work. That means both `whoami` and `Whoami` would match, but not `wHoami`.

We would use the newly added [`Command::mut_subcommands`](https://docs.rs/clap/latest/clap/struct.Command.html#method.mut_subcommands) method so we wouldn't have to do it manually. *The example of that method is the implantation that we would want.*

What do you think @jdonszelmann?

jdonszelmann:
Seems reasonable to me? Neat, thanks for looking into this :3

Kobzol (approved):
That's a pretty cool solution :)

Issue #2114, opened 2025-07-16
Misleading error message when concern is left in a file comment


samueltardieu:
The [concern entered here](https://github.com/rust-lang/rust/pull/143949#discussion_r2211697687) was rejected with [this message](https://github.com/rust-lang/rust/pull/143949#issuecomment-3081291097): "Only organization members can add or resolve concerns."

However, the problem was not organization membership, as the author was a member of rust-lang, but apparently that the concern was written in a file-specific comment. The error message is misleading.

Another (better?) resolution would be to accept concern registered in such messages.

Cc @ChrisDenton

Urgau:
There is no fundamental reason it can't work with a pull-request review comment and the error message actually indicates that it tried adding the concern but failed for another reason.

The actual reason is the same as https://github.com/rust-lang/triagebot/issues/2094, we use GitHub's `author_association` from the webhook response but since your membership status is "Private", it makes GitHub not report to us that you are part of the organization and thus we reject the command.

https://github.com/orgs/rust-lang/people?query=samu
<img width="982" height="143" alt="Image" src="https://github.com/user-attachments/assets/f7ffa6f9-e91b-49b9-be6c-4b7f2554a7b6" />

Short-term solution is to put your membership status as "Public". Longer-term is for us (triagebot) to not us it anymore.

samueltardieu:
Thanks @Urgau, I had no idea that my membership status for the rust-lang organization was private. I've switched it to public.

Issue #2094, opened 2025-07-02
[GitHub REST/Webhook API] `author_association` returns different value than the web UI


Urgau:
While analyzing why triagebot (our bot) [commented this message](https://github.com/rust-lang/rust/pull/143316#issuecomment-3027057882), while it shouldn't as the PR author is a member of the rust-lang organization (as can be seen on the web UI), we came across this issue.

Looking at the webhook event we got from [the previous comment](https://github.com/rust-lang/rust/pull/143316#issuecomment-3027057630), we can see that the `issue.author_association` is set at `CONTRIBUTOR`, not to be confused by the author of the comment who's `author_association` is correctly set at `MEMBER`.

However we can see on the web UI that `@Kobzol` is a member of the rust-lang organization, we would therefor expect the `issue.author_association` to be `MEMBER` instead of `CONTRIBUTOR` as:

> `MEMBER`: The author is a member of the organization that owns the repository.

This can be reproduced using the REST `pulls` endpoint, https://api.github.com/repos/rust-lang/rust/pulls/143316:

```json
{
  // ...
  "author_association": "CONTRIBUTOR",
}
```

It seems like other people have noticed the same bug, [community#18690](https://github.com/orgs/community/discussions/18690).

Urgau:
As mentioned in https://github.com/orgs/community/discussions/18690#discussioncomment-4232389 it's linked to the author membership status (private/public) to the organization.

If it's private then the webhook and unauthenticated API request don't return `CONTRIBUTOR` instead of `MEMBER` with authenticated API requests.

PR #2117, opened 2025-07-17
Use team repo to determine membership for the concern command

Urgau:
This PR moves the concern command off from the `author_association` field in favor of our team repo. This is done in order to avoid us rejecting valid user just because they have their membership private.

Helps with https://github.com/rust-lang/triagebot/issues/2094
Fixes https://github.com/rust-lang/triagebot/issues/2114

### DIFF
--- a/src/handlers/concern.rs
+++ b/src/handlers/concern.rs
@@ -1,13 +1,12 @@
 use std::fmt::Write;
 
 use anyhow::{Context as _, bail};
-use octocrab::models::AuthorAssociation;
 
 use crate::{
     config::ConcernConfig,
     github::{Event, Label},
     handlers::Context,
-    interactions::EditIssueBody,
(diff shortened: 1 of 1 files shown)


src/handlers/concern.rs
```diff
@@ -68,22 +67,22 @@ pub(super) async fn handle_command(
...
+    if !issue_comment
+        .comment
+        .user
+        .is_team_member(&ctx.team)
+        .await
+        .context("failed to verify that the user is a team member")?
+    {
         tracing::info!(
-            "{}#{} tried to register a concern, but author association isn't right: {:?}",
+            "{}#{} tried to register a concern, but author comment {:?} is not part of the team repo",
```
Kobzol:
```suggestion
            "{}#{} tried to register a concern, but comment author {:?} is not part of the team repo",
```

Kobzol (approved):

Issue #2116, opened 2025-07-17
Rendered link does not point to correct file


apiraino:
The automation to add a rendered link to a PR touching documentation sometimes does not point to the correct file.

Example in [rust-forge#918](https://togithub.com/rust-lang/rust-forge/pull/918#issue-3239217119) it points to the file `SUMMARY.md` instead of the new file `repositories.md`, which is more useful to read.

I should probably test which files reach [this point](https://github.com/rust-lang/triagebot/blob/817268ed945e3d42048b5b52a53f527a1cacf6d9/src/handlers/rendered_link.rs#L71) and how the file is choosen from the list.

(not a big deal, in any case)

Urgau:
Opened https://github.com/rust-lang/triagebot/pull/2118 to fix this.

PR #2118, opened 2025-07-17
Improve `[rendered-link]` guess of the relevant file to point to

Urgau:
This pull-request improves the guess of the relevant file to point to by:
1. introducing an exclusion list of filenames ~~(defaults to `SUMMARY.md`)~~
2. by sorting all the matching files by the total number of lines changed (additions + removals + changes)

This should drastically improve our guess as before we took the first file we match on.

cc @apiraino
Fixes https://github.com/rust-lang/triagebot/issues/2116

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -496,7 +496,9 @@ pub(crate) struct BotPullRequests {}
 #[serde(rename_all = "kebab-case")]
 #[serde(deny_unknown_fields)]
 pub(crate) struct RenderedLinkConfig {
+    /// List of paths to watch for modifications
     pub(crate) trigger_files: Vec<String>,
+    /// List of paths to exclude from watching for modifications
     #[serde(default)]
     pub(crate) exclude_files: Vec<String>,
 }


src/handlers/rendered_link.rs
```diff
@@ -47,6 +47,10 @@ async fn add_rendered_link(
                     .trigger_files
                     .iter()
                     .any(|tf| f.filename.starts_with(tf))
+                    && !config
+                        .exclude_files
+                        .iter()
+                        .any(|tf| f.filename.starts_with(tf))
```
Kobzol, line 53: This won't work, because the filename contains the whole path from root (try `gh api repos/rust-lang/rust-forge/pulls/918/files`). Maybe we could check only the filename itself here, not the whole path?

Probably it would be also more accurate to rename `trigger_files` to `trigger_paths`...

src/handlers/rendered_link.rs
```diff
@@ -52,7 +52,15 @@ async fn add_rendered_link(
                         .iter()
                         .any(|tf| f.filename.starts_with(tf))
             })
-            .and_then(|file| {
+            .collect();
+
+        // Sort the relavant files by the total number of lines changed, as to
+        // improve our guess for the relevant file to show the link to.
+        relevant_files.sort_by_key(|f| f.additions + f.deletions + f.changes);
```
Kobzol: You can just use `max_by` instead of sorting the whole list and then taking the last element, but it probably doesn't matter that much here.

src/handlers/rendered_link.rs
Urgau: Well, it's on purpose. `exclude_files` is the mirror of `trigger_paths` which works on the filepath.

Kobzol: Yeah, I know. It's just that the `SUMMARY.md` default isn't very useful here.

src/handlers/rendered_link.rs
Urgau: Didn't know about `Iterator::max_by`, switch to it.

src/handlers/rendered_link.rs
Urgau: I though the `SUMMARY.md` was at the top for most repos but that doesn't seems to be the case. Removed it as the default.

Kobzol (approved):

src/handlers/rendered_link.rs
Urgau: There is also a `Iterator::max_by_key`. Switched to it as it simplifies the implementation.

apiraino:
Did we understand what caused #2116? This feels like a (legit) workaround but I feel I didn't understand the root cause of it.

I infer that the reason is that we don't have a way to tell which .md files are more relevant for the rendered link, correct?

src/config.rs
```diff
@@ -497,6 +497,8 @@ pub(crate) struct BotPullRequests {}
 #[serde(deny_unknown_fields)]
 pub(crate) struct RenderedLinkConfig {
```
apiraino, line 498: this struct is public, I would add a little documentation about its field (although the name of the fields are sort-of self-explanatory)

Urgau: Documentation for the handler and it's fields are at https://forge.rust-lang.org/triagebot/rendered-link.html

src/handlers/rendered_link.rs
apiraino: Is the intention to put index .md files into `exclude_files`? What happens if a patch to the rust-forge repository only changes the file `SUMMARY.md` and that file is in the exclude list? That might be a corner case.

maybe put `exclude_files` back in if the filtering returned nothing? :thinking:

apiraino: How does this handler work when the patchset does not find .md files to render in `trigger_files`?

src/config.rs
apiraino: I like having it while exploring the code but ... ok, I guess it's a matter of taste

(btw: that's user documentation, not developer documentation)

src/handlers/rendered_link.rs
Urgau: If there no relevant files, then nothing happens, the `Rendered link` link is not put in the body.

(technically it's a bit more complex since there could have been a match from a previous link from a previous push, so we need to remove it, but that's already handled below)

src/config.rs
Urgau: Added documentation about the fields

PR #2120, opened 2025-07-19
Add anchor around repository paths in our GHA logs viewer

Urgau:
This PR adds an automatic anchor around paths to the repository in our GHA logs viewer. This is meant to ease navigation to the file where the error occurred or the linked file.

image

Implementation wise we retrieve the SHA of the current job from the jobs details endpoint, as well as the top-level Git tree for the given SHA (GitHub supports recursive but our repos are way to big for that). 

To reduce the number of false-positives the root of each path (modulo `/checkout`) is checked against the root directory of the repository, that way `/checkout/obj`, `obj/build`, `/home/runner` or `$DIR` are immediately excluded and won't have a anchor that would lead no-where.

This is perfect as some relative paths are not relative to the root of the repo (looking at ou miri test suite), but there isn't much we can do about those unfortunately.

The regex can be tested at https://regexr.com/8g2g4.


Urgau:
I was able to reduce the complexity of the implementation with a narrowed regex that doesn't need checks after the fact, at the cost of a somewhat more complex regex.

Local testing shows that the runtime impact went from ~80ms to ~8ms on a normal r-l/r log. https://regex101.com/r/vCnx9Y/2

src/gha_logs.rs
```diff
@@ -11,14 +12,16 @@ use uuid::Uuid;
 pub const ANSI_UP_URL: &str = "/gha_logs/ansi_up@6.0.6.min.js";
 const MAX_CACHE_CAPACITY_BYTES: u64 = 50 * 1024 * 1024; // 50 Mb
 
+type CacheType = ((WorkflowRunJob, String), String);
```
Kobzol: I think it's time for a struct.. :D

src/github.rs
```diff
@@ -1240,6 +1262,29 @@ impl IssuesEvent {
 #[derive(Debug, serde::Deserialize)]
 struct PullRequestEventFields {}
 
+#[derive(Debug, serde::Deserialize)]
+pub struct WorkflowRunJob {
+    pub id: u64,
+    pub run_id: u64,
+    pub name: String,
+    pub workflow_name: String,
```
Kobzol: Currently we only need the SHA, right? So we can just avoid deserializing all the rest, and don't have to define the conclusion.

Kobzol (commented):
Haven't tested the regex yet.

src/gha_logs.rs
```diff
@@ -99,34 +102,64 @@ async fn process_logs(
...
+        };
+
+        let (job_and_tree_roots, logs) = futures::join!(job_and_tree_roots, logs);
+        let job_and_tree_roots_and_jobs = (job_and_tree_roots?, logs?);
 
         ctx.gha_logs
             .write()
             .await
-            .put(log_uuid.clone(), json_logs.into())
+            .put(log_uuid.clone(), job_and_tree_roots_and_jobs.into())
```
Kobzol:
```suggestion
            .put(log_uuid.clone(), job_and_tree_roots_and_logs.into())
```
? But we should really create some named struct with fields here.. e.g. `JobData`. And `ctx.gha_logs` could also be renamed e.g. to `gha_jobs`.

src/gha_logs.rs
```diff
@@ -189,11 +225,18 @@ async fn process_logs(
             `<span class="warning-marker">##[warning]</span>`
         );
 
-        // 5. Add the html to the DOM
+        // 5. Add anchors to GitHub around some paths
+        const pathRegex = /(?<boundary>[^a-zA-Z0-9.\\/])(?<inner>(?:[\\\/]?(?:checkout[\\\/])?(?<path>(?:{tree_roots})[\\\/][a-zA-Z0-9_$\-.\\\/]+))(?::(?<line>[0-9]+):(?<col>[0-9]+))?)/g;
```
Kobzol, line 259: Could you please add a comment here that shows some examples of paths that are (and aren't) parsed? And maybe also explain the `tree_roots` "magic" a bit.

Kobzol (commented):
This is seriously cool! Really great idea.

src/gha_logs.rs
Urgau: Sure, added it as `CachedLog`.

src/github.rs
Urgau: Yes, we currently only need the SHA. I have ideas for the other fields but we can add them later. Only left `head_sha`.

src/gha_logs.rs
Urgau: Sure, I can add some comments.

Regarding the examples, would it be fine if I just redirected to https://regex101.com/r/vCnx9Y/2 ? I think it would be much easier to understand the regex, than me trying to describe what it's supposed to match.

Kobzol: I think copy-pasting 2-3 examples of matched paths to the comment is enough. But feel free to also add the link.

Urgau: Sure, done.

src/gha_logs.rs
```diff
@@ -238,7 +243,19 @@ async fn process_logs(
             `<span class="warning-marker">##[warning]</span>`
         );
 
-        // 5. Add anchors to GitHub around some paths
+        // 5. Add anchors around some paths
+        //  Detailed examples of what the regex does is at https://regex101.com/r/vCnx9Y/2
+        //
+        //  But simply speaking the regex tries to find abosulte (with `/checkout` prefix) and
```
Kobzol: `abosulte` -> `absolute`

Kobzol (approved):
Not sure how well this will work on Windows/macOS (non-Docker-based CI jobs), but it looks great for Linux.

Urgau:
On our rust-lang/rust Windows jobs, the absolute paths are prefixed by `D:\a\rust\rust\` (which we don't handle) but most of the paths are relative which we do handle fine.

Similar story for apple jobs with `/Users/runner/work/rust/rust` for absolute paths, with most of them also being relative.

In any-case I don't think it's a issue for now, we can add support for them later.

PR #2122, opened 2025-07-23
Remove rustfmt.toml, update rustfmt lint to 2024 edition

apiraino:
From [comment](https://github.com/rust-lang/triagebot/pull/2092#issuecomment-3106936671).

I'm curious to see what happens if we remove the `rustfmt.toml` file that is forcing code linting to edition 2018. Running cargo fmt locally and picking the edition (2024) from `Cargo.toml` does not produce any linting errors.

Let's see here what says the CI  ...


ehuss:
I would recommend keeping this file, but updating it to 2024. Some editors call `rustfmt` directly, but that tool is unaware of `Cargo.toml`, and thus defaults to 2015.

apiraino:
ah sorry I didn't consider that! Thanks for the suggestion Eric, let me fix this

PR #2123, opened 2025-07-23
Update rustfmt config to 2024 edition

apiraino:
Fixes my mistake in #2122

Reinstates the `rustfmt.toml` file and updates its linting to edition 2024


Kobzol:
Huh, didn't know that some IDEs invoke `rustfmt` directly. Kind of annoying to keep these rustfmt config files around.

PR #2124, opened 2025-07-27
Misc changes around our GHA logs viewers

Urgau:
See the individual commits for details.

### DIFF
--- a/src/gha_logs.rs
+++ b/src/gha_logs.rs
@@ -3,7 +3,6 @@ use crate::handlers::Context;
 use anyhow::Context as _;
 use hyper::header::{CACHE_CONTROL, CONTENT_SECURITY_POLICY, CONTENT_TYPE};
 use hyper::{Body, Response, StatusCode};
-use itertools::Itertools;
 use std::collections::VecDeque;
 use std::str::FromStr;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


src/gha_logs.rs
```diff
@@ -136,22 +135,45 @@ async fn process_logs(
...
+            let roots_contents: Vec<_> = root_contents
                 .iter()
-                .filter_map(|t| (t.object_type == "tree").then_some(&t.path))
-                .join("|");
+                .filter_map(|t| (t.type_ == "dir").then_some(&t.path))
+                .map(|path| async { ctx.github.contents(&repo, &job.head_sha, path).await })
+                .collect();
+
+            // Join all futures and fail fast if one of them returns an error
+            let roots_contents = futures::future::try_join_all(roots_contents)
```
Kobzol: I'm not sure if this is a good idea, I remember hitting GH API limits when I YOLO-ed a bunch of API calls fully in parallel.

src/gha_logs.rs
```diff
@@ -136,22 +135,45 @@ async fn process_logs(
...
+                .await
+                .context("unable to fetch content details")?;
+
+            let mut tree_roots: Vec<_> = roots_contents
+                .into_iter()
+                .flatten()
+                .map(|d| d.path)
+                .collect();
+
+            // We need to sort the tree roots by decending order, otherwise `library/std` will
```
Kobzol: `descending`

src/gha_logs.rs
```diff
@@ -136,22 +135,45 @@ async fn process_logs(
...
-            // JS regex.
-            let tree_roots = trees
-                .tree
+            // Retrive all the sub-directories contents (for rust-lang/rust it's 6 API calls)
+            let roots_contents: Vec<_> = root_contents
                 .iter()
-                .filter_map(|t| (t.object_type == "tree").then_some(&t.path))
-                .join("|");
+                .filter_map(|t| (t.type_ == "dir").then_some(&t.path))
+                .map(|path| async { ctx.github.contents(&repo, &job.head_sha, path).await })
```
Kobzol: Can't we use https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28&versionId=free-pro-team%40latest&category=repos&subcategory=contents#get-a-tree again, and download the subtrees by their SHA? To avoid downloading actual contents of files in the API call (although for the second level there will hopefully be mostly directories..).

Urgau: I completely misunderstood both APIs, for some reason I though the contents API wouldn't give us the actual content unless requested via the content-type and I though the `git/trees` API couldn't give the sub-directories.

Switched back to the `git/trees` API.

src/gha_logs.rs
Urgau: I have not had any issues in my testing. I think as long as we keep the number of concurrent request reasonable I think it's fine.

Kobzol: The thing is that it doesn't run in isolation, triagebot likely does a bunch of GH requests every second. That being said, I agree that we should try it and only change it if we run into issues in production :)

Urgau:
> The last commit message is wrong now

I'm sure I had modified it. Not sure what happened. Fixed.

src/gha_logs.rs
Urgau: I don't know if the number of concurrent requests really matter for us (it's only 6 calls for r-l/r), I would be a bit more worried about our total requests per hour, but I don't know what's our limit.

PR #2125, opened 2025-07-27
Reduce workload for web browser by removing all inline style

GuillaumeGomez:
Since we now have our own version of `ansi_up`, we can iterate from there as I saw quite a few things that could be improved.

In the meantime, this first iteration removes all inline style which is quite bad for web browsers.

Just one note though: I had to tweak the `ansi_up@6.0.6.min.js` file loaded in the web browser and couldn't find where the file was generated into. Did I miss something obvious?

r? @Urgau


Kobzol:
IMO we should just switch to Rust based ANSI coloring and get rid of ansiup.

GuillaumeGomez:
As wasm or in the server source code?

Kobzol:
Normally in the backend in Rust, Urgau even implemented a Rust crate for this, it just hasn't replaced ansi-up yet. I'm sure Urgau will have some thoughts/comments here :)

GuillaumeGomez:
I would prefer this approach a lot. Waiting for @Urgau to confirm before I actually implement it.

GuillaumeGomez:
@urgau confirmed it by message so doing that instead.

GuillaumeGomez:
And CI passed, yeay. \o/

src/gha_logs.rs
```diff
@@ -234,6 +227,93 @@ async fn process_logs(
...
+
+    // 2. Tranform the ANSI escape codes to HTML
+    let anstyle_svg::HtmlParts { body, style } = anstyle_svg::Term::new()
+        .use_html5(true)
+        .render_html_parts(logs);
+
+    // 3. Add a self-referencial anchor to all timestamps at the start of the lines
+    let html = TIMESTAMP_REGEX.replace_all(&body, |ts: &regex::Captures| {
+        let ts = &ts[0][1..];
+        format!("><a id='{ts}' href='#{ts}' class='timestamp'>{ts}</a>")
```
Kobzol, line 249: Is the leading `>` intended?

src/gha_logs.rs
```diff
@@ -234,6 +227,93 @@ async fn process_logs(
         }
     };
 
+    let tree_roots = tree_roots.iter().map(|p| regex::escape(p)).join("|");
```
Kobzol, line 232: Since the logs can be quite large, doing this processing might not be super cheap (the triagebot server isn't very powerful). What about computing all of this only once, and then caching the whole HTML page output? We'd have to check how does its size compare to the size of the raw logs though, it could be much larger than the log size.

I couldn't benchmark it locally because the regex crashed for me.

src/gha_logs.rs
```diff
@@ -234,6 +227,93 @@ async fn process_logs(
...
+    let regex = format!(
+        r#"(?<boundary>[^a-zA-Z0-9.\\/])(?<inner>(?:[\\/]?(?:checkout[\\/])?(?<path>(?:{tree_roots})(?:[\\/][a-zA-Z0-9_$\\-.\\/]+)?))(?::(?<line>[0-9]+):(?<col>[0-9]+))?)"#,
+    );
+    let gha_logs_path_regex = Regex::new(&regex).unwrap();
```
Kobzol, line 236: This crashed for me locally:
```
thread 'main' panicked at /projects/personal/rust/triagebot/src/gha_logs.rs:234:50:
called `Result::unwrap()` on an `Err` value: Syntax(
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regex parse error:
    (?<boundary>[^a-zA-Z0-9.\\/])(?<inner>(?:[\\/]?(?:checkout[\\/])?(?<path>(?:tests/ui\-fulldeps|tests/ui|tests/rustdoc\-ui|tests/rustdoc\-json|tests/rustdoc\-js\-std|tests/rustdoc\-js|tests/rustdoc\-gui|tests/rustdoc|tests/run\-make|tests/pretty|tests/mir\-opt|tests/incremental|tests/debuginfo|tests/crashes|tests/coverage\-run\-rustdoc|tests/coverage|tests/codegen\-units|tests/codegen\-llvm|tests/auxiliary|tests/assembly\-llvm|tests/COMPILER_TESTS\.md|src/version|src/tools|src/stage0|src/rustdoc\-json\-types|src/rustc\-std\-workspace|src/librustdoc|src/etc|src/doc|src/ci|src/build_helper|src/bootstrap|src/README\.md|library/windows_targets|library/unwind|library/test|library/sysroot|library/stdarch|library/std_detect|library/std|library/rustc\-std\-workspace\-std|library/rustc\-std\-workspace\-core|library/rustc\-std\-workspace\-alloc|library/rtstartup|library/profiler_builtins|library/proc_macro|library/portable\-simd|library/panic_unwind|library/panic_abort|library/coretests|library/core|library/compiler\-builtins|library/alloctests|library/alloc|library/Cargo\.toml|library/Cargo\.lock|compiler/rustc_type_ir_macros|compiler/rustc_type_ir|compiler/rustc_ty_utils|compiler/rustc_transmute|compiler/rustc_traits|compiler/rustc_trait_selection|compiler/rustc_thread_pool|compiler/rustc_target|compiler/rustc_symbol_mangling|compiler/rustc_span|compiler/rustc_session|compiler/rustc_serialize|compiler/rustc_sanitizers|compiler/rustc_resolve|compiler/rustc_query_system|compiler/rustc_query_impl|compiler/rustc_public_bridge|compiler/rustc_public|compiler/rustc_proc_macro|compiler/rustc_privacy|compiler/rustc_pattern_analysis|compiler/rustc_passes|compiler/rustc_parse_format|compiler/rustc_parse|compiler/rustc_next_trait_solver|compiler/rustc_monomorphize|compiler/rustc_mir_transform|compiler/rustc_mir_dataflow|compiler/rustc_mir_build|compiler/rustc_middle|compiler/rustc_metadata|compiler/rustc_macros|compiler/rustc_log|compiler/rustc_llvm|compiler/rustc_lint_defs|compiler/rustc_lint|compiler/rustc_lexer|compiler/rustc_interface|compiler/rustc_infer|compiler/rustc_index_macros|compiler/rustc_index|compiler/rustc_incremental|compiler/rustc_hir_typeck|compiler/rustc_hir_pretty|compiler/rustc_hir_analysis|compiler/rustc_hir|compiler/rustc_hashes|compiler/rustc_graphviz|compiler/rustc_fs_util|compiler/rustc_fluent_macro|compiler/rustc_feature|compiler/rustc_expand|compiler/rustc_errors|compiler/rustc_error_messages|compiler/rustc_error_codes|compiler/rustc_driver_impl|compiler/rustc_driver|compiler/rustc_data_structures|compiler/rustc_const_eval|compiler/rustc_codegen_ssa|compiler/rustc_codegen_llvm|compiler/rustc_codegen_gcc|compiler/rustc_codegen_cranelift|compiler/rustc_builtin_macros|compiler/rustc_borrowck|compiler/rustc_baked_icu_data|compiler/rustc_attr_parsing|compiler/rustc_attr_data_structures|compiler/rustc_ast_pretty|compiler/rustc_ast_passes|compiler/rustc_ast_lowering|compiler/rustc_ast_ir|compiler/rustc_ast|compiler/rustc_arena|compiler/rustc_abi|compiler/rustc|LICENSES/Unicode\-3\.0\.txt|LICENSES/OFL\-1\.1\.txt|LICENSES/NCSA\.txt|LICENSES/MIT\.txt|LICENSES/LLVM\-exception\.txt|LICENSES/ISC\.txt|LICENSES/GPL\-3\.0\-or\-later\.txt|LICENSES/GPL\-2\.0\-only\.txt|LICENSES/GCC\-exception\-3\.1\.txt|LICENSES/CC\-BY\-SA\-4\.0\.txt|LICENSES/BSD\-2\-Clause\.txt|LICENSES/Apache\-2\.0\.txt|\.github/workflows|\.github/renovate\.json5|\.github/pull_request_template\.md|\.github/ISSUE_TEMPLATE)(?:[\\/][a-zA-Z0-9_$\\-.\\/]+)?))(?::(?<line>[0-9]+):(?<col>[0-9]+))?)
```

with `http://localhost:8000/gha-logs/rust-lang/rust/46848365843`.

Urgau: I think that due to `\\-` at the end, which I think should just be `\-`.

GuillaumeGomez: There was an extra `\`. Shame I didn't see it before. :s

GuillaumeGomez: Damn, we commented at the same time. regex101 pointed me to the wrong part of the regex.

src/gha_logs.rs
GuillaumeGomez: Yes, I use the closing of the `<br>` tag as "anchor" to ensure it's a timestamp we want to replace and not something part of the log.

Kobzol: I see, right, there's no lookbehind in `regex` yet. Please add a comment on top of `TIMESTAMP_REGEX` to explain what is it looking for.

GuillaumeGomez: Done!

GuillaumeGomez:
I'm used to having the proxy handle the compression, didn't think it wouldn't be the case. If not we should add one for sure.

So yeah, big downside is that page size is bigger, however the time needed to get the page fully rendered is shorter (as long as you don't have a slow internet connection).

GuillaumeGomez:
Not yet. Also I didn't even start benchmarking anstyle which I think can be improved.

Urgau:
> anstyle part took 800ms

That is nuts, I don't remember seeing such bottleneck when testing out anstyle for our use case. If we can improve that on anstyle part that would be great.

PR #2126, opened 2025-07-28
Add quick start instructions and allow to skip DB migrations

GuillaumeGomez:
Luckily for me, I had @Urgau to help me getting started to work on this. However, this should not be a requirement so here are instructions to make it much simpler.

Also, added an environment variable check to allow skipping DB migrations (and thus removing the need to comment some code when you don't need to use the DB).

All in all, the dev experience is still subpar but that makes it more bearable. Some more improvements would be to provide fake data so we don't need to have a github token.

r? @Kobzol

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,6 +1,13 @@
+# You can get this value from https://github.com/settings/tokens.
 # if `GITHUB_TOKEN` is not set here, the token can also be stored in `~/.gitconfig`
 GITHUB_TOKEN=MUST_BE_CONFIGURED
 DATABASE_URL=MUST_BE_CONFIGURED
+# If this variable is uncommented set to 1, the DB migrations will be skipped.
+# SKIP_DB_MIGRATIONS=0
+# If this variable is uncommented and set to 1, it will skip the workqueue
+# load (which takes ~10-15 seconds).
+# SKIP_WORKQUEUE=0
+
 GITHUB_WEBHOOK_SECRET=MUST_BE_CONFIGURED
 # for logging, refer to this document: https://rust-lang-nursery.github.io/rust-cookbook/development_tools/debugging/config_log.html
 # `RUSTC_LOG` is not required to run the application, but it makes local development easier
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -27,6 +27,25 @@ To compile the Triagebot you need OpenSSL development library to be installed (e
 
 Run `cargo build` to compile the triagebot.
 
+## Quick start
+
+For local development/debugging for the log pages, do the following steps:
+
+ 1. Create a new `.env` file with this content:
```
Kobzol: We already have `.env.sample` (https://github.com/rust-lang/triagebot/blob/master/.env.sample). Instead of duplicating the environment variables here, we should document `SKIP_WORKQUEUE` and `SKIP_DB_MIGRATIONS` in that file.

Kobzol (commented):
I wonder how does skipping migrations help you when you still have to pass `DATABASE_URL`? :thinking: Are you just counting with the fact that no SQL query will be executed?

README.md
GuillaumeGomez: Sounds good to me!

.env.sample
```diff
@@ -1,6 +1,12 @@
+# You can get this value from https://github.com/settings/tokens.
 # if `GITHUB_TOKEN` is not set here, the token can also be stored in `~/.gitconfig`
 GITHUB_TOKEN=MUST_BE_CONFIGURED
 DATABASE_URL=MUST_BE_CONFIGURED
+# If this variable is set to 1, the DB migrations will be skipped.
+SKIP_DB_MIGRATIONS=0
+# If this variable is set to 1, it will skipp the workqueue load (which takes ~10-15 seconds).
+SKIP_WORKQUEUE=0
```
Kobzol:
```suggestion
#SKIP_DB_MIGRATIONS=0
# If this variable is set to 1, it will skip the workqueue load (which takes ~10-15 seconds).
#SKIP_WORKQUEUE=0
```
They are not required, so they should be commented.

GuillaumeGomez:
Updated.

Kobzol (approved):

Issue #1372, opened 2021-05-31
Assign labels with spaces


steffahn:
[relevant Zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/A.20way.20to.20self-label.20soundness.20issues.3F) ([in the archive](https://zulip-archive.rust-lang.org/122651general/88362Awaytoselflabelsoundnessissues.html))

On my quest to find a way to self-label `I-unsound 💥` I’m getting the feeling that rustbot doesn’t offer any way at all to set labels that contain any spaces. From my limited understanding of the source code it looks like the tokenizer understands quoted strings, but the labels parser doesn’t allow such quoted strings for the label name. I would’ve expected something like `@rustbot label "I-unsound 💥"` to work, but it doesn’t currently work. There’s also the relevant question on whether `@rustbot label -"I-unsound 💥"` or `@rustbot label "-I-unsound 💥"` or both options would be the way to remove the label again.

nikomatsakis:
I'm tempted to say the problem is the label name :)

nikomatsakis:
Or, alternatively, maybe we could load the labels and strip off trailing emojis and whitespace? it'd be nice if you could just type `I-unsound`.

steffahn:
The label was already renamed ;-)

steffahn:
Might be useful to do the whitespace+emoji stripping if there is demand for having the 💥 back.

nikomatsakis:
Seems like we can close this issue, then?

steffahn:
I suppose we can.

PR #2128, opened 2025-07-29
Normalize labels before addition or removal

kailan:
The crates.io repository has some cute emoji in the label names, but this interferes with usage of label-related triagebot commands. Example:

Screenshot 2025-07-29 at 12 19 18

It seems unreasonable to expect users to exactly match the label names in these cases, and it would be a shame to remove the emoji from the label names, so I feel that it makes sense to make triagebot a bit more lax when it comes to finding labels.

This PR adds a `normalize_and_match_labels` function, which takes a set of "requested" label names and attempts to match them to the labels available in the repository. This will always return an exact match, if it exists, but otherwise will strip emoji, ignore casing, and ignore leading and trailing whitespace in an attempt to find the relevant label.

This function is used in the lower level `add_labels` and `remove_label` functions so that it can apply across any triagebot commands that handle labels. This means that the existing configuration for autolabelling in crates.io will start working:

```toml
[autolabel."A-backend"]
trigger_files = [
    "src/",
]
```

This PR is somewhat related to #1372.


src/github.rs
```diff
@@ -565,21 +565,12 @@ impl IssueRepository {
         format!("{}/{}", self.organization, self.repository)
     }
 
-    async fn has_label(&self, client: &GithubClient, label: &str) -> anyhow::Result<bool> {
-        #[allow(clippy::redundant_pattern_matching)]
-        let url = format!("{}/labels/{}", self.url(client), label);
-        match client.send_req(client.get(&url)).await {
```
kailan, line 571: No need to query each label as we have already matched them based on the result of the `labels()` function below.

src/github.rs
```diff
@@ -788,32 +844,13 @@ impl Issue {
...
-            }
-        }
-
-        if !unknown_labels.is_empty() {
-            return Err(UnknownLabels {
-                labels: unknown_labels,
-            }
-            .into());
-        }
-
```
kailan, line 807: This is all handled by `normalize_and_match_labels` now.

Urgau:
> It seems unreasonable to expect users to exactly match the label names in these cases

I feel like it's unreasonable for GitHub to allow emoji in label names but here we are.

> This PR adds a `normalize_and_match_labels` function

Unless I'm mistaken, your function would allow the same prefix to be match to multiple emoji. "label :shrug:", "label :+1:" and "label :100:" would all be matched by "label", right?

I don't think we should any of those to match "label", I feel like we should require perfect match in those cases.

-------

I didn't look too closely at the code but I saw that there wasn't any caching done, we'll need to add some, at minimum per webhook request, maybe in [`Issue`](https://github.com/rust-lang/triagebot/blob/0f75748b83675a6d9afa25f4fdac94944cbc6954/src/github.rs#L364) to start and in the long run per repository.

kailan:
> I feel like it's unreasonable for GitHub to allow emoji in label names but here we are.

Can't argue with that, lol.

> Unless I'm mistaken, your function would allow the same prefix to be match to multiple emoji. So "label 🤷", "label 👍" and "label 💯" would all be matched by "label", right?
> 
> I don't think we should allow any of those to match "label", I feel like we should require perfect match in those cases.

This is a great point, so I have added an error for this case:

`AmbiguousLabelMatch` –  "Unsure which label to use for `label` - could be one of: `label 🤷`, `label 👍`, `label 💯`"

> I didn't look too closely at the code but I saw that there wasn't any caching done

I didn't implement any caching, however this implementation will send far fewer requests to GitHub – previously a `GET /labels/label_name` would have been made for every requested label, this is now replaced by a single `GET /labels` which provides the data needed to check all labels.

src/github.rs
```diff
@@ -597,6 +588,30 @@ impl fmt::Display for UnknownLabels {
...
+impl fmt::Display for AmbiguousLabelMatch {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "Unsure which label to use for `{}` - could be one of: {}",
+            self.requested_label,
+            self.labels
+                .iter()
+                .map(|l| format!("`{}`", l))
+                .collect::<Vec<_>>()
```
Urgau: We have itertools, so collecting shouldn't be necessary.
```suggestion
```

src/github.rs
```diff
@@ -730,8 +745,71 @@ impl Issue {
         Ok(())
     }
 
+    async fn normalize_and_match_labels(
+        &self,
+        client: &GithubClient,
+        requested_labels: &[String],
+    ) -> anyhow::Result<Vec<String>> {
+        let available_labels = self.repository().labels(client).await.unwrap_or_default();
```
Urgau: Let's bubble up the error instead of ignoring it.
```suggestion
        let available_labels = self.repository().labels(client).await.context("unable to retrieve the repository labels")?;
```

src/github.rs
```diff
@@ -730,8 +745,71 @@ impl Issue {
...
+
+        let emoji_regex: Regex = Regex::new(r"[\p{Emoji}\p{Emoji_Presentation}]").unwrap();
```
Urgau: Let's move this regex into a `LazyLock<Regex>` to avoid re-creating the regex each time.

```rust
static EMOJI_REGEX: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"[\p{Emoji}\p{Emoji_Presentation}]").unwrap());
```

src/github.rs
```diff
@@ -730,8 +745,71 @@ impl Issue {
         Ok(())
     }
 
+    async fn normalize_and_match_labels(
```
Urgau, line 712: Would be great if we could add tests for this new logic. Simplest way would be to extract everything after `available_labels` into a separate synchronous function, that way we can some very simple `#[test]` in this file.

src/github.rs
```diff
@@ -730,8 +745,71 @@ impl Issue {
...
+        }
+
+        if !unknown_labels.is_empty() {
+            return Err(UnknownLabels {
+                labels: unknown_labels.into_iter().map(String::from).collect(),
+            }
+            .into());
+        }
+
+        Ok(found_labels.into_iter().map(|s| s.clone()).collect())
```
Urgau: Can't we just return `found_labels` here (without this collect)?

src/github.rs
```diff
@@ -730,8 +745,71 @@ impl Issue {
...
+                unknown_labels.push(requested_label.as_str());
+            } else if found.len() > 1 {
+                return Err(AmbiguousLabelMatch {
+                    requested_label: requested_label.clone(),
+                    labels: found.into_iter().map(|l| l.name.clone()).collect(),
+                }
+                .into());
+            } else {
+                found_labels.push(&found.first().unwrap().name);
+            }
```
Urgau: nit: you can match on a slice to avoid the manual len comparisons.

```suggestion
            match &found[..] {
                [] => {
                    unknown_labels.push(requested_label.as_str());
                },
                [label] => {
                    found_labels.push(&label.name);
                },
                [..] => {
                    return Err(AmbiguousLabelMatch {
                        requested_label: requested_label.clone(),
                        labels: found.into_iter().map(|l| l.name.clone()).collect(),
                    }
                    .into());
                },
            };
```

src/github.rs
```diff
@@ -730,8 +745,71 @@ impl Issue {
         Ok(())
     }
 
+    async fn normalize_and_match_labels(
+        &self,
+        client: &GithubClient,
+        requested_labels: &[String],
```
Urgau: Can we just take a slice of `&str`: `&[&str]` ?

src/github.rs
```diff
@@ -730,8 +745,71 @@ impl Issue {
...
+        }
+
+        Ok(found_labels.into_iter().map(|s| s.clone()).collect())
+    }
+
     pub async fn remove_label(&self, client: &GithubClient, label: &str) -> anyhow::Result<()> {
         log::info!("remove_label from {}: {:?}", self.global_id(), label);
+
+        let normalized_labels = self
+            .normalize_and_match_labels(client, &[label.to_string()])
```
Urgau:
```suggestion
            .normalize_and_match_labels(client, &[label])
```

src/github.rs
```diff
@@ -730,8 +745,71 @@ impl Issue {
...
+            .await?;
+        let label = normalized_labels.first().unwrap();
```
Urgau: Let's avoid panicking even if it should impossible.

src/github.rs
```diff
@@ -767,6 +845,19 @@ impl Issue {
         labels: Vec<Label>,
     ) -> anyhow::Result<()> {
         log::info!("add_labels: {} +{:?}", self.global_id(), labels);
+
+        let labels = self
+            .normalize_and_match_labels(
+                client,
+                &labels.into_iter().map(|l| l.name).collect::<Vec<_>>(),
```
Urgau: maybe?
```suggestion
                &labels.iter().map(|l| &l.name).collect::<Vec<_>>(),
```

src/github.rs
kailan: TIL! this is great

src/labels.rs
Urgau, line 1: Triagebot handles multiple services (mainly GitHub and Zulip) and this file while technically having being independent of GitHub has some assumptions about it, and doesn't make sense for Zulip.

Could you move it under `src/github/`?

Urgau: Otherwise this looks good.

Urgau (approved):

Urgau:
Thanks for the PR. Let's try it.

Urgau:
Let's do some quick sanity checks.

@rustbot label A-test

kailan:
> Hum, this doesn't work 100% as I would have expected. GitHub allows both Unicode emoji characters and the Markdown named ones (`:smile:`, ...), and returns the raw/unexpected version in it's API.

Oh, that's definitely an oversight that should be addressed.

> I think this may have broken standard ascii labels in some way too; libc's autolabel has started claiming everything is unknown

even worse! Okay, I will revisit this with some new test cases based on these issues.

Probably worth rolling this back if you haven't already. Apologies for the inconvenience.

Urgau:
Revert on going at https://github.com/rust-lang/triagebot/pull/2133, apologies for the inconvenience.

PR #2129, opened 2025-07-29
Reduce workload for web browser by removing all inline style

GuillaumeGomez:
Until https://github.com/rust-lang/triagebot/pull/2125 is ready, here's a nice improvement for the existing JS.

cc @Urgau @Kobzol


src/gha_logs/ansi_up@6.0.6.min.js
Urgau, line 1: This looks good to me.

However due too our aggressive caching you will need to modify `ANSI_UP_URL` to something else. Maybe like this? (and maybe rename the file as well)
```diff
- pub const ANSI_UP_URL: &str = "/gha_logs/ansi_up@6.0.6.min.js";
+ pub const ANSI_UP_URL: &str = "/gha_logs/ansi_up@6.0.6-custom.js";

Kobzol: We can just version it, v0, v1, etc.

Urgau (approved):

Urgau:
I've verified that it's still works as expected and that the caching is indeed refreshed.

Thanks for the help.

PR #2130, opened 2025-07-29
link to relevant part of rustc dev guide in modified submodule warning

lolbinarycat:
this makes finding this section when it is needed much easier for new contributors.


Kobzol (approved):
Good idea, thanks!

PR #2131, opened 2025-07-29
Also verify that the end boundary isn't path-like

Urgau:
The sorting of paths introduced in https://github.com/rust-lang/triagebot/pull/2124 for the tree roots isn't enough as `library/std` can still to be recognized for `library/stdarch`.

This is the case when `library/stdarch` is not included in the tree roots, which is the case in the lastest versions of https://github.com/rust-lang/triagebot/pull/2124 since we exclude submodules.

To fix this, we add an end boundary to the regex that checks that we don't match if there are path-like elements after the path we found.

Can be tested here https://regex101.com/r/cKwJql/1

I also the opportunity to split-up the regex in logical pieces to better see what each part does.

### DIFF
--- a/src/gha_logs.rs
+++ b/src/gha_logs.rs
@@ -335,14 +335,16 @@ body {{
         //
         //  Any other paths, in particular if prefixed by `./` or `obj/` should not taken.
         const pathRegex = new RegExp(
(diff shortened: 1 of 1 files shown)


Urgau:
I was looking at an old log, where `library/stdarch` was still printed, but the issue still remains, even if maybe more theoretical.

PR #2132, opened 2025-08-01
Allow opting out of workqueue tracking with a label

Kobzol:
This allows reviewers to add a `S-no-work-capacity-tracking` label to a PR. When this label is present, the PR will not count towards the review queue capacity of a reviewer assigned to that PR.

Requested by @BoxyUwU.

### DIFF
--- a/src/handlers/pr_tracking.rs
+++ b/src/handlers/pr_tracking.rs
@@ -324,6 +324,9 @@ fn waits_for_a_review(
     let is_rollup = labels.iter().any(|l| l.name == "rollup");
     let is_waiting_for_reviewer = labels.iter().any(|l| l.name == "S-waiting-on-review");
     let is_assigned_to_author = assignees.contains(author);
+    let has_capacity_tracking_opt_out = labels
+        .iter()
+        .any(|l| l.name == "S-no-work-capacity-tracking");
 
     is_open
         && !is_draft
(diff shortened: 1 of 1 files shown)


Urgau:
We already ignore draft, blocked, waiting on author, ... PRs. What is the usecase for an explicit label? Is there any reason it can't be handled by our standard labels?

Urgau:
> for whatever reason

That's the thing I'm trying to get to, what's the "whatever reason" that motivated this?

I feel like it's difficult to evaluate this PR without some context first.

(and to be clear, this solution seems reasonable, I just don't know if it's the right solution to the problem)

Kobzol:
Right. @BoxyUwU could you chime in? :)

BoxyUwU:
team-aware would help with the release aspect of it yeah. but it wouldn't help with the case where I've already allocated time to spend on mentoring on feature X, and in addition to that have my misc compiler reviews. "mentoring on feature X" can often wind up in spending that allocated time reviewing a T-compiler PR

Urgau:
Would it be useful to have a global/fallback capacity and then specific capacities for more specialized tasks/projects? I could imaging us having the ability to have different capacities based one the PR labels, that way you could set different capacity for `T-release` PRs, `F-feature-xxx-i-mentor` PRs and general PRs.

Urgau:
Alright, thanks for the context. It seems like the situation would quite complex to represent, and since you seems happy with a simple opt-out let's go with that.

PR #2134, opened 2025-08-02
Migrate to axum and hyper 1 :tada: 

Urgau:
This PR migrates triagebot endpoints to axum (0.8) and hyper (1.6) :tada:

In the process we clean up some things, add compression of responses, add catch panic, make use of axum types when possible (`Json`, `Path`, ...) and simplify a bunch of logic.

cc @ehuss @Mark-Simulacrum @GuillaumeGomez


src/github/webhook.rs
```diff
@@ -0,0 +1,357 @@
...
+}
+
+pub async fn webhook(
+    headers: HeaderMap,
+    State(ctx): State<Arc<crate::handlers::Context>>,
+    body: Bytes,
+) -> axum::response::Response {
+    // Extract X-GitHub-Event header
+    let event = match headers.get("X-GitHub-Event") {
+        Some(ev) => match ev.to_str() {
```
GuillaumeGomez: Better make a `let Some(ev) else { return ... }`

src/github/webhook.rs
```diff
@@ -0,0 +1,357 @@
...
+        None => {
+            return (StatusCode::BAD_REQUEST, "X-GitHub-Event header must be set").into_response();
+        }
+    };
+
+    debug!("event={event}");
+
+    // Extract X-Hub-Signature-256 header
+    let signature = match headers.get("X-Hub-Signature-256") {
+        Some(sig) => match sig.to_str() {
```
GuillaumeGomez: Same here.

src/github/webhook.rs
```diff
@@ -0,0 +1,357 @@
...
+    };
+
+    debug!("signature={signature}");
+
+    // Check signature on body
+    if let Err(_) = check_payload_signed(&signature, &body) {
+        return (StatusCode::FORBIDDEN, "Wrong signature").into_response();
+    }
+
+    let payload = match str::from_utf8(&body) {
```
GuillaumeGomez: Same here.

src/github/webhook.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+}
+
+impl std::error::Error for SignedPayloadError {}
+
+pub fn check_payload_signed(signature: &str, payload: &[u8]) -> Result<(), SignedPayloadError> {
+    let signature = signature
+        .strip_prefix("sha256=")
+        .ok_or(SignedPayloadError)?;
+    let signature = match hex::decode(&signature) {
```
GuillaumeGomez, line 327: `let Ok(signature) = .. else {}`.

Urgau: We use the `Err` variant, so can't use let-else here.

src/github/webhook.rs
```diff
@@ -0,0 +1,337 @@
...
+        return (StatusCode::BAD_REQUEST, "X-GitHub-Event header must be set").into_response();
+    };
+    let Ok(ev) = ev.to_str() else {
+        return (
+            StatusCode::BAD_REQUEST,
+            "X-GitHub-Event header must be UTF-8 encoded",
+        )
+            .into_response();
+    };
+    let Ok(event) = ev.parse::<EventName>();
```
GuillaumeGomez, line 140: I'm surprised this syntax works. O.o

GuillaumeGomez (approved):

src/github/webhook.rs
Urgau: That's because there isn't really a error type, it's a irrefutable error.

https://github.com/Urgau/triagebot/blob/28af06f1926b14a2c931e3052d2c8ed71f4571a9/src/github/webhook.rs#L63

Urgau:
> `http_body_util` seems to be unused?

Yeah, it's not used directly by triagebot, removed it.

Kobzol (approved):

PR #2135, opened 2025-08-05
add unclaim as an alias for release-assignment

lolbinarycat:


parser/src/command/assign.rs
```diff
@@ -68,7 +68,7 @@ impl AssignCommand {
             } else {
                 return Err(toks.error(ParseError::NoUser));
             }
-        } else if let Some(Token::Word("release-assignment")) = toks.peek_token()? {
+        } else if let Some(Token::Word("release-assignment" | "unclaim")) = toks.peek_token()? {
```
Urgau, line 71: Please add a test below, so we can make sure it works and won't regress.

lolbinarycat: unit test added!

Urgau, line 71: Could you also update the documentation on the enum above.

Urgau: We'll also need to update the user-documentation on the forge at https://github.com/rust-lang/rust-forge/blob/master/src/triagebot/issue-assignment.md

lolbinarycat: I plan on updating the forge once this is merged. (doc comment has been updated)

Urgau (approved):

PR #2137, opened 2025-08-07
Add `exclude-titles` option to no-mentions 

blyxyas:
Previously, users were getting warnings for including merge commits to PRs that mentioned only @rustbot ([Example](https://github.com/rust-lang/rust/pull/145056#issuecomment-3164621504)). Now it has to have at least one human mentioned in order to trigger a comment.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -515,7 +515,12 @@ pub(crate) struct IssueLinksConfig {
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
 #[serde(rename_all = "kebab-case")]
 #[serde(deny_unknown_fields)]
-pub(crate) struct NoMentionsConfig {}
+pub(crate) struct NoMentionsConfig {
+    /// The check will not be performed on titles that include any of these substrings (case
+    /// insensitive)
+    #[serde(default)]
+    pub(crate) exclude_titles: Vec<String>,
+}
 
 /// Configuration for PR behind commits checks
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -681,6 +686,7 @@ mod tests {
             trigger-files = ["posts/"]
 
             [no-mentions]
+            exclude-titles = ["subtree update"]
 
             [behind-upstream]
             days-threshold = 14
@@ -769,7 +775,9 @@ mod tests {
                 issue_links: Some(IssueLinksConfig {
                     check_commits: true,
                 }),
-                no_mentions: Some(NoMentionsConfig {}),
+                no_mentions: Some(NoMentionsConfig {
+                    exclude_titles: vec!["subtree update".into()],
+                }),
                 behind_upstream: Some(BehindUpstreamConfig {
                     days_threshold: Some(14),
                 }),
--- a/src/handlers/check_commits/no_mentions.rs
+++ b/src/handlers/check_commits/no_mentions.rs
@@ -8,7 +8,11 @@ pub(super) fn mentions_in_commits(
     conf: &NoMentionsConfig,
     commits: &[GithubCommit],
 ) -> Option<String> {
-    if conf.exclude_titles.iter().any(|s| pr_title.contains(s)) {
+    if conf
+        .exclude_titles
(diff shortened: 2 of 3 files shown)


blyxyas:
Yes I'd be willing to implement such a change, triagebot is a project I like working on.

Although I'm not sure how many titles that aren't subtree syncs (and thus, contain merge commits mentioning rustbot) would benefit from disabling the no-mentions check.

blyxyas:
Done! I've reoriented the pull request and added support for a new `exclude_titles = ["..."]` config opt.

src/config.rs
```diff
@@ -515,7 +515,12 @@ pub(crate) struct IssueLinksConfig {
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
 #[serde(rename_all = "kebab-case")]
 #[serde(deny_unknown_fields)]
-pub(crate) struct NoMentionsConfig {}
+pub(crate) struct NoMentionsConfig {
+    /// The check will not be performed on titles that include any of these substrings (case
+    /// insensitive)
```
Urgau, line 520: Unless I missed something, the check uses `contains` which isn't case insensitive, which is fine, but incoherent with this doc.

```suggestion
    /// The check will not be performed on titles that include any of these substrings
```

src/handlers/check_commits/no_mentions.rs
```diff
@@ -42,20 +57,51 @@ Signed-off-by: Foo Bar <foobar123@example.com>
...
+        "This is a body that only mentions @rustbot for a command!",
+    ));
+
+    assert_eq!(
+        mentions_in_commits("any title", &default_conf, &commits),
+        None
+    );
+
+    commits.push(dummy_commit_from_body(
+        "6565ffdd8af4ca0ec7c8faceee59c582edcd83b2",
```
Urgau: nit: let's use different sha otherwise it can confusing which sha we are referencing
```suggestion
        "4894129179b361200c9cd733ba0e906bf98747a2",
```

src/handlers/check_commits/no_mentions.rs
```diff
@@ -42,20 +57,51 @@ Signed-off-by: Foo Bar <foobar123@example.com>
...
     ));
 
     assert_eq!(
-        mentions_in_commits(&NoMentionsConfig {}, &commits),
+        mentions_in_commits("any title", &default_conf, &commits, ),
         Some(
             r"There are username mentions (such as `@user`) in the commit messages of the following commits.
 *Please remove the mentions to avoid spamming these users.*
-- d7daa17bc97df9377640b0d33cbd0bbeed703c3a
+- 6565ffdd8af4ca0ec7c8faceee59c582edcd83b2
```
Urgau:
```suggestion
- 4894129179b361200c9cd733ba0e906bf98747a2
```

src/config.rs
blyxyas: I forgot to actually make it case insensitive, fixed!

src/handlers/check_commits/no_mentions.rs
```diff
@@ -4,12 +4,24 @@
...
 pub(super) fn mentions_in_commits(
-    _conf: &NoMentionsConfig,
+    pr_title: &str,
+    conf: &NoMentionsConfig,
     commits: &[GithubCommit],
 ) -> Option<String> {
+    if conf
+        .exclude_titles
+        .iter()
+        .any(|s| pr_title.to_lowercase().contains(s))
```
Urgau: I think `s` (the excluded substring) also needs to be lowercased, otherwise an uppercased excluded title won't ever match with the always lowercased PR title.

Urgau (approved):
Thanks for working on this. Let's ship it.

Urgau:
@blyxyas Forgot to ask, would you be up sending a PR updating the documentation of `[no-mentions]` in the forge? https://github.com/rust-lang/rust-forge/blob/master/src/triagebot/no-mentions.md

blyxyas:
Yeah, of course! I'll open it ASAP

PR #2138, opened 2025-08-07
Reorder CI, make it fail fast

blyxyas:
The test suite takes about 1m 42 seconds to run, while `cargo fmt --check` takes less than a second. This PR reorders them

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,10 +25,10 @@ jobs:
       - uses: actions/checkout@v4
       - run: rustup toolchain install stable --profile minimal
       - uses: Swatinem/rust-cache@v2
-      - name: Run tests
-        run: cargo test --workspace --all-targets
       - name: Check formatting
         run: cargo fmt --all --check
+      - name: Run tests
+        run: cargo test --workspace --all-targets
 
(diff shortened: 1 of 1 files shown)


Kobzol:
IMO it makes more sense to run the easiest to fix (and run) things last. You don't want to get a formatting failure from CI if there is also a test failure.

blyxyas:
The idea behind contribution is that opening a PR and having an immediate CI error gives the contributor feedback quicker, gives them the opportunity to fix it right there and then, and making them more vigilant for the next commit.

Also, saving CI time.

Kobzol:
I think that the more important thing is to take a look at which CI steps fail more often, and how they affect one another. If you have both a formatting and a test fail, you always want to fix the test first, because that will also require another reformat. Fixing a formatting fail just to push and see a test failure is a waste of time. In general, formatting is always the *last* thing that you want to do, so it should also run last on CI, IMO. Most people also have autoformatting in their IDE, so it rarely fails, and thus should run towards the end of the CI job.

But for triagebot specifically, I don't think it really matters, so I will stop discussing this, as it's just bikeshedding and a waste of time, Imo :) Feel free to merge if you want.

Issue #2141, opened 2025-08-08
Plain enhanced logs not working on Firefox ESR 128


bjorn3:
`RegExp.escape` doesn't exist yet in this version. Firefox ESR is what Debian ships by default. Firefox started supporting `RegExp.escape` in 134.

bjorn3:
Also no error is showed anywhere in the ui. The page remains blank.

apiraino:
@bjorn3 I'm sorry I don't understand what this issue is about 😅

bjorn3:
Indeed.

PR #2144, opened 2025-08-08
Polyfill `RegExp.escape` for older web-browser

Urgau:
This PR adds a simple polyfill for the `RegExp.escape` function we use in our GHA logs viewer, as the function is quite recent on mainstream web browser: Safari (end 2024), Firefox (Jan. 2025) and Chrome/Edge (May. 2025).

Fixes https://github.com/rust-lang/triagebot/issues/2141


Kobzol (approved):
Thanks!

PR #2145, opened 2025-08-08
Report client side errors to the user in our GHA logs viewer

Urgau:
Fixes https://github.com/rust-lang/triagebot/issues/2141#issuecomment-3166857318


Urgau:
I don't know how useful that would be without the more details context in the logs. Let's keep it that way for now.

PR #2148, opened 2025-08-10
Avoid emit [behind-upstream] more than once

xizheyin:
I set the warning message for `[behind-upstream]` to be unchange when `upstream_commit_url` is the same so that it does not issue continuous warnings (i.e. `last_warnings==warnings`). 

The original `days_old` always changes, but `age_threshold` does not change. This is the smallest modification I can think of, but there may be more refined control methods.

It is suggested by [#triagebot > Outdated commit message @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Outdated.20commit.20message/near/533598281)

cc @Urgau @Kobzol

### DIFF
--- a/src/handlers/check_commits/behind_upstream.rs
+++ b/src/handlers/check_commits/behind_upstream.rs
@@ -27,8 +27,9 @@ pub(super) async fn behind_upstream(
             days_old
         );
 
+        // The message remain the same to ensure the warning is emit only once.
         Some(format!(
-            r"This PR is based on an [upstream commit]({upstream_commit_url}) that is {days_old} days old.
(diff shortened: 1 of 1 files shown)


src/handlers/check_commits/behind_upstream.rs
```diff
@@ -27,8 +27,9 @@ pub(super) async fn behind_upstream(
             days_old
         );
 
+        // The message remain the same to ensure the warning is emit only once.
```
Urgau:
```suggestion
        // The message remain the same to ensure the warning is emit only once for each base commit.
```

Urgau (approved):

PR #2149, opened 2025-08-10
Introduce a range-diff viewer for GitHub force-push 

Urgau:
This PR adds an *emulated* git range-diff viewer for use GitHub and force push.

It's a well known issue with GitHub compare UI that when a force-push changes the base commit that GitHub's compare page will show all changes from the old base to the new head, which includes enormous amount of unrelated changes.

`git range-diff` solves that by diff showing the diff of the old and new revision, and since GitHub doesn't offer this capability we can do it ourselves.

This new feature works by having a new HTML endpoint `/gh-range-diff/{owner}/{repo}/{basehead}` that takes the org, repo and `base..head`, which we then use to retrieve the old and new diff.
Those two diffs are then compared against each other, the resulting differences are then shown to the user.

| GitHub compare | range-diff |
|----------------|------------|
| image | image |
| https://github.com/rust-lang/rust/pull/132146#event-17439126441 | [localhost/range-diff](http://localhost:8000/gh-range-diff/rust-lang/rust/186956d116a191a7e2c0af816922960dbc1512e4..18980df6cd19c6e87746634d894f0fdfc679bf28) |

> To access our range-diff viewer users be able to use the bookmarklet we provide in the page, it redirects from the GitHub compare page to our viewer.


Urgau:
I just pushed a bunch a small changes which should improve the visual clarity, by reducing the highlighting parts and re-using `git range-diff` block sign to better distinguish the diff of diffs markers from the diffs markers.

I also added before and after links to each files, improved the bookmarklet help and added a small explanation for the block markers. Feedback welcome of course.

<details>

| before | after |
|----------------|------------|
| image | image | 

</details>

src/gh_range_diff.rs
```diff
@@ -0,0 +1,313 @@
...
+    header::{CONTENT_SECURITY_POLICY, CONTENT_TYPE},
+};
+use imara_diff::{
+    Algorithm, Diff, InternedInput, Interner, Token, UnifiedDiffConfig, UnifiedDiffPrinter,
+};
+use pulldown_cmark_escape::FmtWriter;
+
+use crate::{github, handlers::Context, utils::AppError};
+
+pub async fn gh_range_diff(
```
Kobzol, line 31: Could you please add some comment here to explain what it does, and explain what exactly should `basehead` contain?

Kobzol (commented):
Awesome work!

The minus/plus with red/green background is a bit distracting visually, what about just keeping the red/green bar without minus/plus? The background contrast is also a bit dim, I'd either make the background lighter or the font more visible/higher opacity/"more white".

Urgau:
Well, there is complexity, they are not always duplicated, we can have `-+`. We can see that on the exemple I gave. It means that the added line from the previous patch is removed.

Urgau:
There isn't any crap left, a range-diff does a diff of diffs, as you would expect. 

It's nevertheless quite confusing the first time you read one, as we need to show what was added and removed on the first push vs what was added and removed on the second push.

That's why there is those double markers, the first one indicates what has changed between the the first and second push, and the second marker indicates what was/is the change.

Kobzol:
I'm not saying that it's better, just that I personally find it less confusing. I think the current range diff is fine.

Kobzol (approved):

src/gh_range_diff.rs
Urgau: Sure, done.

PR #2150, opened 2025-08-11
Zulip stream type check

apiraino:
The Zulip Message API has changed a little bit since we implemented our client. This patch:
- replaces "private" with "direct" for direct messages (see [docs](https://zulip.com/api/send-message#parameter-type)), no functional change.
- fixes/improves the direct vs. stream check we do when the triagebot server receives a message from the Zulip server. The previous check was not *wrong* per se but the correct check should be on the message `type`.
- Rewrote/fixed and (hopefully) clarified the documentation about testing Zulip stuff 

tbh these changes are not *really* necessary but slightly improve the correctness of our client.

For anyone who has an opinion :)

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -52,15 +52,14 @@ struct Message {
     sender_email: String,
     /// The ID of the stream.
     ///
-    /// `None` if it is a private message.
+    /// `None` if it is a direct message.
     stream_id: Option<u64>,
     /// The topic of the incoming message. Not the stream name.
     ///
-    /// Not currently set for private messages (though Zulip may change this in
-    /// the future if it adds topics to private messages).
+    /// Not currently set for direct messages (though Zulip may change this in
+    /// the future if it adds topics to direct messages).
     subject: Option<String>,
-    /// The type of the message: stream or private.
-    #[allow(unused)]
+    /// The type of the message: stream or direct.
     #[serde(rename = "type")]
     type_: String,
 }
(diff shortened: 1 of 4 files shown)


Urgau (approved):
Looks okay to me, but I haven't tested it, so I will let others take a look otherwise I will merge in the next few days.

apiraino:
thanks! Testing this is a bit tedious, but I will do that in the next couple of days

src/zulip.rs
```diff
@@ -52,15 +52,14 @@ struct Message {
...
     stream_id: Option<u64>,
     /// The topic of the incoming message. Not the stream name.
     ///
-    /// Not currently set for private messages (though Zulip may change this in
-    /// the future if it adds topics to private messages).
+    /// Not currently set for direct messages (though Zulip may change this in
+    /// the future if it adds topics to direct messages).
     subject: Option<String>,
-    /// The type of the message: stream or private.
-    #[allow(unused)]
```
apiraino, line 63: notice that I also removed the "unused" attribute introduced in 41a793d08c2642edbf8ced3fd5b388d2b4547229

EYEHACK-M (approved):
@hassan

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2150/5c4f2b156e38c48ce758a963040fb19685e17f14..0c03e1300e21355fc5b30f1d8e614dd6d053b772)*

PR #2155, opened 2025-08-13
Accept backports from zulip

apiraino:


src/zulip.rs
```diff
@@ -302,10 +319,77 @@ async fn handle_command<'a>(
...
+
+    // TODO: factor out the Zulip "URL encoder" to make it practical to use
+    let zulip_send_req = crate::zulip::MessageApiRequest {
+        recipient: Recipient::Stream {
+            id: stream_id,
+            topic: &subject,
+        },
+        content: "",
+    };
+    let zulip_link = zulip_send_req.url(zulip_client);
```
apiraino, line 386: Note: for how the Message API works today, we can't pin exactly a single message so the link in the GitHub comment will be to the whole topic, which is a bit inconvenient but hopefully [will be solved by Zulip](https://rust-lang.zulipchat.com/#narrow/channel/122653-zulip/topic/.22near.22.20parameter.20in.20payload.20of.20send.20message.20API/near/533313688). I'd rather wait on them for a correct fix rather us add hacks to make it work.

apiraino:
I guess this patch went under the radar because I didn't request a review

r? triagebot

src/zulip/commands.rs
```diff
@@ -173,6 +176,16 @@ pub struct PingGoalsArgs {
     pub next_update: String,
 }
 
+#[derive(clap::Parser, Debug, PartialEq, Clone)]
+pub struct BackportArgs {
+    /// Release channel this backport is pointing to. Allowed: "beta" or "stable".
+    pub channel: String,
```
Kobzol: Could you please use an `enum` here for the channel and the verb, and either generate the parser with `#[derive(clap::ValueEnum)]`, or write it manually with `FromStr`? That way we won't have to check for invalid data in the backport handler, and we'll get a nice error message and a hint (in case of misspelling) for free.

src/zulip.rs
```diff
@@ -24,12 +24,29 @@ use axum::Json;
...
+use octocrab::Octocrab;
 use rust_team_data::v1::{TeamKind, TeamMember};
 use std::cmp::Reverse;
 use std::fmt::Write as _;
 use std::sync::Arc;
 use subtle::ConstantTimeEq;
-use tracing as log;
+use tracing::log;
+
+const BACKPORT_APPROVED: &str = "
```
Kobzol: Could you please turn these into functions that will receive the formatting arguments as parameters, and return a `format!`ed `String`? That way we'll be sure that we passed all the necessary parameters at compile-time.

src/zulip.rs
```diff
@@ -302,10 +319,77 @@ async fn handle_command<'a>(
...
+        content: "",
+    };
+    let zulip_link = zulip_send_req.url(zulip_client);
+
+    let message_body = message_body
+        .replace("{args.channel}", args.channel.as_str())
+        .replace("{args.verb}", verb.as_str())
+        .replace("{zulip_link}", zulip_link.as_str());
+
+    tokio::spawn(async move {
```
Kobzol: `tokio::spawn` shouldn't be needed here. We have other commands that do a lot more stuff and they work fine, it shouldn't take 10+s in most cases to send one GH comment.

src/zulip.rs
```diff
@@ -302,10 +319,77 @@ async fn handle_command<'a>(
...
+        let res = octo_client
+            .issues(repo_owner, repo_name)
+            .create_comment(args.pr_num, &message_body)
+            .await
+            .context("unable to post comment on #{args.pr_num}");
```
Kobzol: This will be formatted as `#{args.pr_num}`. You can use e.g. `with_context(|| anyhow::anyhow("unable to post comment on #{}", args.pr_num))`.

src/zulip.rs
Kobzol, line 386: This is currently unused and can be removed, right?

Kobzol (commented):
Sounds pretty useful!

src/zulip.rs
apiraino: `zulip_link` is used a bit down to format the link to the Zulip topic.

(unless I am not reading correctly where your comment is pointing to)

src/github.rs
```diff
@@ -1319,9 +1319,6 @@ impl IssuesEvent {
     }
 }
 
-#[derive(Debug, serde::Deserialize)]
-struct PullRequestEventFields {}
-
```
apiraino, line 1316: after the rebase cargo said this is not used anymore (probably a left-over from another patch)

src/zulip.rs
Kobzol: Huh, not sure how I missed that, sorry.

Kobzol: Oh, I see, it wasn't actually used! Because it was just a string placeholder in `BACKPORT_APPROVED` that was mistakenly not actually interpolated. Good thing we changed it to a function :)

apiraino: no, it is used indeed. now I moved it above the *if-then-else*

src/zulip.rs
apiraino: The task spawning was meant to "fire and forget" this task, sort of sending a message to a queue (which in principle I find it more appropriate)

src/zulip.rs
```diff
@@ -17,20 +17,47 @@ use crate::utils::pluralize;
...
-use tracing as log;
+use tracing::log;
+
+fn get_text_backport_approved(
+    channel: &BackportChannelArgs,
+    verb: &BackportVerbArgs,
+    zulip_link: &str,
+) -> String {
+    format!("
+{channel} backport {verb} as per compiler team [on Zulip]({zulip_link}). A backport PR will be authored by the release team at the end of the current development cycle. Backport labels handled by them.
```
Kobzol: "Backport labels handled by them." This sounds a bit weird, doesn't it miss "are" or something like that?

src/zulip.rs
```diff
@@ -296,10 +323,68 @@ async fn handle_command<'a>(
...
+    message_data: &Message,
+    octo_client: &Octocrab,
+    zulip_client: &ZulipClient,
+    args_data: &BackportArgs,
+) -> anyhow::Result<Option<String>> {
+    let message = message_data.clone();
+    let args = args_data.clone();
+    let stream_id = message.stream_id.unwrap();
+    let subject = message.subject.unwrap();
+    let octo_client = octo_client.clone();
```
Kobzol: I don't think we need to clone the client (not a big deal though).

src/zulip.rs
```diff
@@ -296,10 +323,68 @@ async fn handle_command<'a>(
...
+            get_text_backport_declined(&args.channel, &args.verb, &zulip_link)
+        }
+    };
+
+    let res = octo_client
+        .issues(repo_owner, repo_name)
+        .create_comment(args.pr_num, &message_body)
+        .await
+        .with_context(|| anyhow::anyhow!("unable to post comment on #{}", args.pr_num));
+    if res.is_err() {
```
Kobzol: I would rather send a reply back to the stream if we fail to post the comment, rather than just logging the error. Just using `?` after `with_context` should do the trick.

src/zulip.rs
```diff
@@ -296,10 +323,68 @@ async fn handle_command<'a>(
...
+        tracing::error!("failed to post comment: {0:?}", res.err());
+    }
+
+    Ok(Some("".to_string()))
```
Kobzol:
```suggestion
    Ok(None)
```
Otherwise triagebot might try to send an empty reply back (I think). In any case, if we don't want to reply with anything, it should be `Ok(None)`.

src/zulip.rs
apiraino: (I think both are fine)

Kobzol (approved):
I wonder if triagebot reads triagebot commands from its own comments 🤔 I guess we'll see :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2155/c6b85492bdac744d60ffe60e00943d7a886c0d6a..4c3dd4e34b22c7e1e7da745eab2d3ef0ce05d0ce)*

apiraino:
> I wonder if triagebot reads triagebot commands from its own comments 🤔 I guess we'll see :)

Do you mean if on Zulip calling `@triagebot` in a comment would trigger a circular `@triagebot` invocation? I'm not sure about what you mean

apiraino:
ah ok! Yes it will trigger a reaction applying the labels (for GH after all it's a normal comment from a GH user).

Thanks for the review!

PR #2157, opened 2025-08-14
Add `robots.txt` and block all user-agents

Urgau:
Looking at the logs I can see that a bunch of requests are being made to `/robots.txt` by crawlers, and in particular AI crawlers, probably because of the gha-logs links we post on GitHub.

Let's tell then that we do not have anything interesting for them, by adding a `robots.txt` that indicates that bots shouldn't interact with us.


Kobzol (approved):
Shoo, bots!

PR #2158, opened 2025-08-15
Post comment to our range-diff a PR was rebased

Urgau:
This PR adds a new handler `[range-diff]` that enables triagebot to post a comment to our range-diff viewer when a PR was rebased.

To detected that a rebased was done we check that if the merge base commits are the same between the "before" and "after" push.

I tested the change locally.

Context: [#t-compiler > Experimental range-diff for force-push @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Experimental.20range-diff.20for.20force-push/near/534649322)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -48,6 +48,7 @@ pub(crate) struct Config {
     pub(crate) no_mentions: Option<NoMentionsConfig>,
     pub(crate) behind_upstream: Option<BehindUpstreamConfig>,
     pub(crate) backport: Option<BackportConfig>,
+    pub(crate) range_diff: Option<RangeDiffConfig>,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -551,6 +552,12 @@ pub(crate) struct BackportRuleConfig {
     pub(crate) add_labels: Vec<String>,
 }
 
+/// Configuration for rebase range-diff comment
+#[derive(Default, PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub(crate) struct RangeDiffConfig {}
+
 fn get_cached_config(repo: &str) -> Option<Result<Arc<Config>, ConfigurationError>> {
     let cache = CONFIG_CACHE.read().unwrap();
     cache.get(repo).and_then(|(config, fetch_time)| {
@@ -689,6 +696,8 @@ mod tests {
             required-pr-labels = ["T-libs", "T-libs-api"]
             required-issue-label = "regression-from-stable-to-stable"
             add-labels = ["stable-nominated"]
+
+            [range-diff]
         "#;
         let config = toml::from_str::<Config>(&config).unwrap();
         let mut ping_teams = HashMap::new();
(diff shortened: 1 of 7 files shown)


src/config.rs
```diff
@@ -551,6 +552,11 @@ pub(crate) struct BackportRuleConfig {
     pub(crate) add_labels: Vec<String>,
 }
 
+#[derive(Default, PartialEq, Eq, Debug, serde::Deserialize)]
```
Kobzol, line 556: Could you please add some doc comment on top of this? It would be nice to have some simple programmatic documentation about the individual config options in this file.

src/handlers/check_commits/force_push_range_diff.rs
```diff
@@ -0,0 +1,58 @@
...
+    if !matches!(event.action, crate::github::IssuesAction::Synchronize) {
+        return Ok(());
+    }
+
+    let (Some(before), Some(after)) = (event.before.as_ref(), event.after.as_ref()) else {
+        tracing::warn!("synchronize event but no before or after field");
+        return Ok(());
+    };
+
+    let org = event.repository.owner();
```
Kobzol, line 26: We can check if `forced` is `true` (https://docs.github.com/en/webhooks/webhook-events-and-payloads#issues) to avoid having to call `compare` for non-force pushes.

src/handlers/check_commits/force_push_range_diff.rs
```diff
@@ -0,0 +1,58 @@
...
+    let repo = event.repository.name();
+
+    let compare_before = ctx
+        .github
+        .compare(
+            &IssueRepository {
+                organization: org.to_string(),
+                repository: repo.to_string(),
+            },
+            &event.issue.base.as_ref().context("no base ref")?.git_ref,
```
Kobzol: Why does `Issue::compare` use `base.sha`, and this uses `.git_ref`? I guess it doesn't matter that much, but it seems inconsistent/confusing.

src/handlers/check_commits/force_push_range_diff.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    let protocol = if host.starts_with("localhost:") {
+        "http"
+    } else {
+        "https"
+    };
+
+    // Rebase detected, post a comment to our range-diff.
+    event.issue.post_comment(&ctx.github,
+        &format!(r#"This PR was rebased! Check out the changes with our [`range-diff`]({protocol}://{host}/gh-range-diff/{org}/{repo}/{before}..{after})."#)
```
Kobzol:
```suggestion
        &format!(r#"This PR was rebased onto a different master commit! Check out the changes with our [`range-diff`]({protocol}://{host}/gh-range-diff/{org}/{repo}/{before}..{after})."#)
```
I often do interactive rebases just to reorder commits etc. A rebase itself doesn't mean that the merge base has changed.

Kobzol (commented):
Looks great! Left some comments.

src/handlers/check_commits/force_push_range_diff.rs
Urgau: Unfortunately we don't have that information, it's only given in for the `push` event (which only work for the current repository, not forks).

I looked at the raw payload, and I don't see anything that indicates whenever it's a force push or not.

src/handlers/check_commits/force_push_range_diff.rs
Urgau: Hum, mistake on my part, switch to `base.sha`.

src/handlers/check_commits/force_push_range_diff.rs
Kobzol: Ah, I checked the wrong webhook, nevermind.

PR #2160, opened 2025-08-16
Misc changes for our range-diff

Urgau:
This PR contains some changes regarding our range-diff feature:
 1. We indicate to the users when there aren't any differences (instead of showing nothing)
 2. We hide the previous range-diff comment when posting a new one
 3. Remove code quote around range-diff link
 4. Clarify the comment regarding PR authors and it's usefulness
    [#t-compiler/help > ✔ rustbot tells me my PR was rebased on a different master](https://rust-lang.zulipchat.com/#narrow/channel/182449-t-compiler.2Fhelp/topic/.E2.9C.94.20rustbot.20tells.20me.20my.20PR.20was.20rebased.20on.20a.20different.20master)
 5. Add tests for the comment logic

### DIFF
--- a/src/handlers/check_commits/force_push_range_diff.rs
+++ b/src/handlers/check_commits/force_push_range_diff.rs
@@ -1,10 +1,22 @@
 use anyhow::Context as _;
 
 use crate::config::RangeDiffConfig;
+use crate::db::issue_data::IssueData;
 use crate::github::GithubCompare;
 use crate::github::IssueRepository;
 use crate::github::IssuesEvent;
+use crate::github::ReportedContentClassifiers;
 use crate::handlers::Context;
+
+/// Key for the state in the database
+const RANGE_DIFF_LINK_KEY: &str = "range-diff-link";
+
+/// State stored in the database
+#[derive(Debug, Default, serde::Deserialize, serde::Serialize, Clone, PartialEq)]
+struct RangeDiffLinkState {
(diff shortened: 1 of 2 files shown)


src/handlers/check_commits/force_push_range_diff.rs
```diff
@@ -54,10 +66,49 @@ pub(super) async fn handle_event(
...
+        r#"This PR was rebased onto a different {branch} commit! Check out the changes with our [`range-diff`]({protocol}://{host}/gh-range-diff/{org}/{repo}/{oldbase}..{oldhead}/{newbase}..{newhead})."#
+    );
+
+    // Rebase detected, post a comment linking to our range-diff.
+    post_new_comment(ctx, event, message).await?;
+
+    Ok(())
+}
+
+async fn post_new_comment(
```
Urgau, line 99: I don't like that this logic is here.

I would prefer if it was shared with the other check-commits sub-handlers, but given the requirement of only hiding the previous comment when posting a new one, which is different from all the others sub-handler, it's was simpler to do it here rather than increasing the complexity in check-commits.

Kobzol: I guess that in theory we could add the logic to the shared "check commits result" comment, but for now this is fine.

Kobzol (approved):

PR #2163, opened 2025-08-19
Update review body with link to the changes since the review

Urgau:
This PR does two things:
 1. it adds a new endpoint `/gh-changes-since/{owner}/{repo}/{pr}/{oldbasehead}` which either redirect to GitHub's compare page when the base commit is the same or redirect to our range-diff when the base commit changed.
 2. it adds a new handler (`[review-changes-since]`) which automatically adds a link this newly endpoint on every "review" (see above example)

The goal is to make it easier to review the changes that happened since the review.

Asked by Ralf in [#t-compiler > Experimental range-diff for force-push @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Experimental.20range-diff.20for.20force-push/near/534963522)

<details>
image
</details>

### DIFF
--- a/src/gh_changes_since.rs
+++ b/src/gh_changes_since.rs
@@ -7,7 +7,11 @@ use axum::{
 };
 use hyper::StatusCode;
 
-use crate::{github, handlers::Context, utils::AppError};
+use crate::{
+    github,
+    handlers::Context,
+    utils::{AppError, is_repo_autorized},
+};
 
 /// Redirects to either `/gh-range-diff` (when the base changed) or to GitHub's compare
 /// page (when the base is the same).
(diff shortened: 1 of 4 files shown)


src/gh_changes_since.rs
```diff
@@ -0,0 +1,72 @@
...
+use anyhow::Context as _;
+use axum::{
+    extract::{Path, State},
+    response::{IntoResponse, Redirect, Response},
+};
+use hyper::StatusCode;
+
+use crate::{github, handlers::Context, utils::AppError};
+
+pub async fn gh_changes_since(
```
Kobzol, line 20: Could you please add a doc comment with a paragraph that describes in a few sentences what this does? Thanks!

src/gh_changes_since.rs
```diff
@@ -0,0 +1,72 @@
...
+    };
+
+    let repos = ctx
+        .team
+        .repos()
+        .await
+        .context("unable to retrieve team repos")?;
+
+    // Verify that the request org is part of the Rust project
+    let Some(repos) = repos.repos.get(&owner) else {
```
Kobzol: We might want to extract this check, together with the `repos` load, it's in a bunch of handlers at this point.

src/gh_changes_since.rs
Urgau: Sure, done.

src/gh_changes_since.rs
Urgau: Also, done.

Urgau:
> I start worrying about feature creep a bit, but this genuinely seems useful.. let's try it!

Yeah, we've recently added a bunch of features, fortunately they are all pretty straightforward and IMO simple to maintain. I wouldn't want to implement more complex features and hopefully GitHub will at some point implement them natively.

> Btw, GitHub might seriously consider paying you for doing R&D 😂

I wish. I should send them a receipt 😆.

src/handlers/review_changes_since.rs
```diff
@@ -0,0 +1,53 @@
+use anyhow::Context as _;
+
+use crate::{
+    config::ReviewChangesSinceConfig,
+    github::{Comment, Event, Issue, IssueCommentAction, IssueCommentEvent},
+    handlers::Context,
+};
+
+pub(crate) async fn handle(
```
Kobzol, line 11: Oops, I actually mainly wanted to have this commented.

PR #2166, opened 2025-08-21
Rename ShortcutCommand to ReviewCommand

apiraino:
The ShortcutCommand was born to handle one case (reviews) but I'd like to expand its scope to handle more convenience custom shortcuts for commands that will modify labels.

However just adding more code would make things a bit confusing, so I'd first start clarifying the scope of the existing code and in subsequent patches add more shortcut commands.

At a high level, I can imagine something like this:
- shortcut::ReviewCommand to handle commands like `author`, `review`, `ready`, etc.
- shortcut::RegressionCommand to handle future commands for quick switching label regressions (todo)
- shortcut:NewCommand (and so on)

Functionally this patch changes nothing, it's just renaming things.

I wrote about this on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Generic.20shorcut.20handler/near/535472560).

Thanks for having a look here. I'd like to get some opinons if this makes sense before moving forward with further work.

r?

### DIFF
--- a/parser/src/command.rs
+++ b/parser/src/command.rs
@@ -23,7 +23,7 @@ pub enum Command<'a> {
     Nominate(Result<nominate::NominateCommand, Error<'a>>),
     Prioritize(Result<prioritize::PrioritizeCommand, Error<'a>>),
     Second(Result<second::SecondCommand, Error<'a>>),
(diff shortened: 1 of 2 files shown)


apiraino:
forgot what I wanted to do here

PR #2153, opened 2025-08-12
Improve range-diff colors, contrast and mobile layout

Urgau:
image

PR #2167, opened 2025-08-22
Make `/gh-range-diff/...` coloring more similar to `git range-diff` coloring

Enselic:
I love the [new](https://github.com/rust-lang/rust/pull/144499#issuecomment-3214780025) range-diff feature of rustbot! Thanks!

I think the current color scheme is confusing though, since it is very different from the regular `git range-diff` tool, which I am used to. This PR makes the coloring the same as `git range-diff`. It really is just a diff of a diff, with different styles for "before", "after" and "top-level diff".

Since we make the coloring the same, remove the explicit explanation, since now I think it becomes easy to see what the range-diff shows. This is just a quick evening hack, I'm happy to make further adjustments.

## New look
Using [this](https://github.com/rust-lang/rust/pull/144499#issuecomment-3214780025) as an example, here is the coloring with this PR for both dark and light mode:

image

image

## Old look

For reference, here is how the same range-diff is rendered before this PR for both dark and light mode. I find it very confusing that a removed line in the "before" version is completely uncolored, and that an added line in the "before" version is marked as red (which indicates removal):

image

image

## CLI `git range-diff` look
And finally, here is how the CLI tool `git range-diff` renders the same diff (only dark mode shown). As you can see it is much more similar to the new coloring than the old:

image

cc @Urgau @Kobzol since you seems to be the one working and reviewing this new very useful feature.

### DIFF
--- a/src/gh_range_diff.rs
+++ b/src/gh_range_diff.rs
@@ -242,18 +242,24 @@ fn process_old_new(
       overflow-x: auto;
     }}
     .removed-block {{
-      background-color: rgb(255, 206, 203);
+      background-color: rgba(255, 150, 150, 1);
       white-space: pre;
     }}
     .added-block {{
-      background-color: rgb(172, 238, 187);
+      background-color: rgba(150, 255, 150, 1);
       white-space: pre;
     }}
-    .removed-line {{
-      color: #DE0000;
+    .removed-line-after {{
+      color: rgb(220, 0, 0)
     }}
-    .added-line {{
-      color: #2F6500;
+    .added-line-after {{
+      color: rgb(0, 73, 0)
+    }}
+    .removed-line-before {{
+      color: rgb(192, 78, 76)
+    }}
+    .added-line-before {{
+      color: rgb(63, 128, 94)
     }}
     @media (prefers-color-scheme: dark) {{
       body {{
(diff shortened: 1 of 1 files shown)


Urgau:
I'm not opposed to changing the colors, but as I wrote [in the announcement](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Experimental.20range-diff.20for.20force-push/near/534646415), having different colors from `git range-diff` was a deliberate choice.

I have found when implementing the feature that `git range-diff` would more often than not unnecessarily highlight some parts, and dim other parts. Kobzol was (and still is I believe) very confused about the colors. See https://github.com/rust-lang/triagebot/pull/2153 and https://github.com/rust-lang/triagebot/pull/2156 for the adjustments we made.

To take an [example](https://triage.rust-lang.org/gh-range-diff/rust-lang/rust/0c8f814fa3d35c1aa7f8ce933e88c82d94a6cd39..304c51d7eae38698ae90e3b40afc6300aea76f25). I find all the green colors to be overwhelming. Triagebot on the other hand shows (IMO) a much cleaner and direct way to see what actually changed.

| `git range-diff` | triagebot |
|--------|---------|
| image | image |

> I find it very confusing that a removed line in the "before" version is completely uncolored, and that an added line in the "before" version is marked as red (which indicates removal)

That's because we use the colors to (try to) indicate the final state (like a normal diff would) compared to `git range-diff` which use the colors the show the differences between the two diffs.

Although I have to agree that it's imperfect and a bit weird in your example.

Urgau:
Looking at the PR I see you haven't touch the "context" lines, which is the majority of the green colors in the `git range-diff` output and my major complain about it.

For the example I gave above, it would look like this. Which is honestly fine to me.

image

Curious to hear @Kobzol opinion on it.

src/gh_range_diff.rs
```diff
@@ -391,13 +393,19 @@ enum HunkTokenStatus {
     Removed,
 }
 
+enum BeforeOrAfter {
+    Before,
+    After,
+}
```
Urgau: Is seems to me that `BeforeOrAfter` is equivalent to `HunkTokenStatus`? (`Added` == `After` and `Before` == `Removed`)

If my interpretation is right, could you just keep one. No need to have two identical enums.

src/gh_range_diff.rs
```diff
@@ -264,23 +270,19 @@ fn process_old_new(
...
     }}
     </style>
 </head>
 <body>
 <h3>range-diff of {oldbase}<wbr>...{oldhead} {newbase}<wbr>...{newhead}</h3>
-<p>Bookmarklet: <a href="{bookmarklet}" title="Drag-and-drop me on the bookmarks bar, and use me on GitHub compare page.">range-diff</a> <span title="This javascript bookmark can be used to access this page with the right URL. To use it drag-on-drop the range-diff link to your bookmarks bar and click on it when you are on GitHub's compare page to use range-diff compare.">&#128712;</span> | {ADDED_BLOCK_SIGN}&nbsp;<span class="added-line">+</span> adds a line | {ADDED_BLOCK_SIGN}&nbsp;<span class="removed-line">-</span> removes a line | {REMOVED_BLOCK_SIGN}&nbsp;<span class="removed-line">+</span> removes the added line | {REMOVED_BLOCK_SIGN}&nbsp;- cancel the removal</p>
+<p>Bookmarklet: <a href="{bookmarklet}" title="Drag-and-drop me on the bookmarks bar, and use me on GitHub compare page.">range-diff</a> <span title="This javascript bookmark can be used to access this page with the right URL. To use it drag-on-drop the range-diff link to your bookmarks bar and click on it when you are on GitHub's compare page to use range-diff compare.">&#128712;</span></p>
```
Urgau: I think we may still want to keep the explanation, to help people not familiar with `range-diff`, but reduced to two options:
 - `[-] before` (with the faded color)
 - `[+] after`

src/gh_range_diff.rs
```diff
@@ -242,18 +242,24 @@ fn process_old_new(
...
-    .added-line {{
-      color: #2F6500;
+    .added-line-after {{
+      color: rgba(0, 255, 0, 1);
+    }}
+    .removed-line-before {{
+      color: rgba(100, 0, 0, 1);
+    }}
+    .added-line-before {{
+      color: rgba(0, 100, 0, 1);
```
Urgau: These colors look great in dark mode, unfortunately in light they are IMO not great (none of them respect the WCAG accessibility contrast guidelines).

Could you move the current `before` and `after` rules below in the dark mode section.

As for the light mode, I've experimented locally and I think those colors looks better in light mode (they also respect WCAG accessibility contrast guidelines):
 - `removed-line-before: rgb(192, 78, 76)`
 - `added-line-before: rgb(63, 128, 94)`
 - `removed-line-after: rgb(220, 0, 0)`
 - `added-line-after: rgb(0, 73, 0)`

| Before | After |
| ------- | ---- |
| Image | Image |

(Feel free to use different ones, as long as they are at least >= AA)

Enselic: I'm happy with your pick of colors so I used them verbatim. Here are how things look with the latest code (5db7ee8):

Dark mode:
image

Light mode:
image

src/gh_range_diff.rs
Enselic: Seems completely reasonable. Done. (Hopefully I didn't misunderstand.)

See screenshots [here](https://github.com/rust-lang/triagebot/pull/2167#discussion_r2295704976).

src/gh_range_diff.rs
Enselic: Oops, you're right. Good point. Fixed.

PR #2170, opened 2025-09-02
Exclude drafts from auto-backport nomination

apiraino:
We have a mechanism (see https://github.com/rust-lang/triagebot/commit/2556583bc5acb3461f1a1c4caa085b59c58ff81a) that auto-nominates pull requests for a backport if they fix a P-high/P-critical regression.

@jieyouxu suggested in this [Zulip comment](https://rust-lang.zulipchat.com/#narrow/channel/474880-t-compiler.2Fbackports/topic/.23146098.3A.20stable-nominated/near/537190990) to exclude pull requests drafts from this mechanism (which makes sense), so this patch implements this suggestion.

r? @jieyouxu 

thanks for a review :)


jieyouxu (approved):
Thanks, this makes sense, though I'm not super familiar with triagebot's codebase so rolling a triagebot reviewer :3

src/handlers/backport.rs
```diff
@@ -50,13 +50,14 @@ pub(super) async fn parse_input(
         None => return Ok(None),
     };
 
-    // Only handle events when the PR is opened or the first comment is edited
+    // Only handle events when the PR is opened (and it's not a draft) or when the first comment is edited
     let should_check = matches!(event.action, IssuesAction::Opened | IssuesAction::Edited);
-    if !should_check || !event.issue.is_pr() {
+    if !should_check || !event.issue.is_pr() || !event.issue.draft {
```
Kobzol: This should be `event.issue.draft`, right? Otherwise you would exclude non-draft PRs :sweat_smile:

Kobzol (changes requested):

src/handlers/backport.rs
apiraino: :sob:

apiraino:
@Kobzol back to this old small patch (had forgot about yr comment). I inverted the check boolean logic and added your suggestions. Feel to have a look when you have a sec.

thanks

Kobzol (approved):
Looks good, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2170/30acfc2a3f6affab301b31882910e7f893474ac2..d4cd168ffc6633117f36320655d4afc34f054a15)*

Issue #1851, opened 2024-11-01
Automatic ping when certain patterns occur in the diff?


RalfJung:
The `#[rustc_allow_const_fn_unstable]` attribute should only be added with wg-const-eval approval. However, it hard to ensure that all reviewers know about this, and it is easy for a reviewer to miss this even if they are generally aware of this policy.

So could we get some bot help for that? If a PR adds new occurrences of this attribute, can we get an automatic ping from rustbot? Ideally, PRs removing the attribute wouldn't trigger the ping. So only lines with a `+` in the diff should be considered. Also, it might be worth restricting this to the part of the diff that is inside the `library/` folder.

PR #2171, opened 2025-09-02
Add way to do mentions of content changes

Urgau:
This PR adds a way to do mentions of content changes (instead of just filename changes).

```
[mentions."#[rustc_attr]"]
type = "content"
message = "This is a message."
cc = ["@someone"]
```

Fixes https://github.com/rust-lang/triagebot/issues/1851

### DIFF
--- a/src/handlers/mentions.rs
+++ b/src/handlers/mentions.rs
@@ -3,7 +3,7 @@
 //! interested people.
 
 use crate::{
-    config::{MentionsConfig, MentionsPathConfig},
+    config::{MentionsConfig, MentionsEntryConfig, MentionsEntryType},
     db::issue_data::IssueData,
     github::{IssuesAction, IssuesEvent},
     handlers::Context,
@@ -17,12 +17,13 @@ use tracing as log;
 const MENTIONS_KEY: &str = "mentions";
 
 pub(super) struct MentionsInput {
-    paths: Vec<String>,
+    to_mention: Vec<String>,
 }
 
 #[derive(Debug, Default, Deserialize, Serialize, Clone, PartialEq)]
 struct MentionState {
-    paths: Vec<String>,
+    #[serde(alias = "paths")]
+    entries: Vec<String>,
 }
 
 pub(super) async fn parse_input(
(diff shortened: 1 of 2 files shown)


src/handlers/mentions.rs
```diff
@@ -61,23 +62,31 @@ pub(super) async fn parse_input(
...
         let to_mention: Vec<_> = config
-            .paths
+            .entries
             .iter()
-            .filter(|(path, MentionsPathConfig { cc, .. })| {
-                let path = Path::new(path);
-                // Only mention matching paths.
-                let touches_relevant_files = file_paths.iter().any(|p| p.starts_with(path));
+            .filter(|(entry, MentionsEntryConfig { cc, type_, .. })| {
+                let relevent = match type_ {
```
Kobzol:
```suggestion
                let relevant = match type_ {
```
:)

src/handlers/mentions.rs
```diff
@@ -94,23 +103,30 @@ pub(super) async fn handle_input(
...
         }
         match message {
             Some(m) => result.push_str(m),
-            None => write!(result, "Some changes occurred in {to_mention}").unwrap(),
+            None => match type_ {
+                MentionsEntryType::Filename => {
+                    write!(result, "Some changes occurred in {to_mention}").unwrap()
+                }
+                MentionsEntryType::Content => {
+                    write!(result, "Some changes regarding `{to_mention}` occurred").unwrap()
```
Kobzol: We could maybe also print the filenames where the mention was found? In case it is a large diff. But I'm not sure if it's necessary, most of the time it should be obvious.

Kobzol (approved):
Freel free to merge with/without including the filepaths

src/handlers/mentions.rs
```diff
@@ -61,23 +63,42 @@ pub(super) async fn parse_input(
...
+                            .iter()
+                            .filter(|p| p.starts_with(path))
+                            .map(|p| PathBuf::from(p))
+                            .collect()
+                    }
+                    MentionsEntryType::Content => {
+                        // Only mentions byte-for-byte matching content inside the patch.
+                        files
+                            .iter()
+                            .filter(|f| f.patch.contains(&**entry))
```
fmease: Didn't Ralf wish for checking *additions* only in #1851 (i.e., exclude deletions and context lines)? Tho I guess also flagging *deletions* might be quite important in some cases. That leaves us with "context lines"? Does `f.patch` contain these or are we good?

Urgau: > Didn't Ralf wish for checking _additions_ only in #1851 (i.e., exclude deletions and context lines)? Tho I guess also flagging _deletions_ might be quite important in some cases.

I missed that, but as you said flagging _deletions_ might also be important. If someone really want that we can IMO see later.

> That leaves us with "context lines"? Does `f.patch` contain these or are we good?

They do contain "context lines". I guess we'll want to restrict to only additions and removal lines.

Urgau: That's done.

src/handlers/mentions.rs
```diff
@@ -94,23 +115,36 @@ pub(super) async fn handle_input(
...
+        let MentionsEntryConfig { message, cc, type_ } = &config.entries[&entry];
         if !result.is_empty() {
             result.push_str("\n\n");
         }
         match message {
             Some(m) => result.push_str(m),
-            None => write!(result, "Some changes occurred in {to_mention}").unwrap(),
+            None => match type_ {
+                MentionsEntryType::Filename => {
+                    write!(result, "Some changes occurred in {entry}").unwrap()
```
Kobzol, line 131: Hmm, we don't actually use the relevant file paths for the filename. But I guess it doesn't hurt to collect them.

src/handlers/mentions.rs
```diff
@@ -122,3 +156,64 @@ pub(super) async fn handle_input(
     }
     Ok(())
 }
+
+fn patch_contains(patch: &str, needle: &str) -> bool {
+    for line in patch.lines() {
+        if (!line.starts_with("+++") && line.starts_with('+'))
+            || (!line.starts_with("---") && line.starts_with('-'))
```
RalfJung, line 163: I don't think we should trigger a mention when the text is removed.

Urgau: We discussed that above in https://github.com/rust-lang/triagebot/pull/2171#discussion_r2317002450 and considered that removal might also be important. We can change that if you prefer since you're the only user.

RalfJung: I don't know when you'd care about deletions. Additions is what I proposed originally. At least for the attributes I have in mind for this, removing them requires no extra care (beyond the things CI already ensures).

Kobzol: Won't changes also count as a removal + addition?

RalfJung: Yeah so they would still cause a ping even if we just scan `+` lines, which seems like reasonable behavior to me.

Urgau: Opened https://github.com/rust-lang/triagebot/pull/2176 to change the behavior.

Issue #2002, opened 2025-05-21
Shortcuts for regressions moving to the next release channel


apiraino:
Here's a TODO for me:

Add a `@rustbot to-beta` shortcut for:
`@rustbot label -regression-from-stable-to-nightly +regression-from-stable-to-beta`

Add a `@rustbot to-stable` shortcut for:
`@rustbot label -regression-from-stable-to-beta +regression-from-stable-to-stable`

They would be very useful for me (more concise to write and no need to go to the label UI widget)

I *was* sure there was something like that already but I can't find them in the forge so it was probably just my imagination ...

@rustbot claim

apiraino:
I was (almost) right, it's already in place but only for pull requests and the mapping is hardcoded
https://forge.rust-lang.org/triagebot/shortcuts.html

The handler was implemented in #1381

- Why was this handler restricted to issues?
- I think it could make sense reworking this handler and have it parse the configuration from the triagebot.toml

PR #2172, opened 2025-09-03
Add alias for compound labels

### LINKED ISSUE #1634: label names handled by the shortcut command should be configurable
Currently the label names handled by the shortcut command are hardcoded:

https://github.com/rust-lang/triagebot/blob/a2ae093f52e39cd051df4340b68a752f77464c9c/src/handlers/shortcut.rs#L29-L31

That would enforce other repos to use the exact same names, but that restriction shouldn't be necessary and we should be able to change them as we want.

apiraino:
Discussion on [#triagebot > Generic shorcut handler @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Generic.20shorcut.20handler/near/519696621)).

WIP, ~~currently blocked on a lifetime issue~~

Closes #1634 and #2002

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,5 +1,6 @@
 use crate::changelogs::ChangelogFormat;
 use crate::github::{GithubClient, Repository};
+use parser::command::relabel::{Label, LabelDelta, RelabelCommand};
 use std::collections::{HashMap, HashSet};
 use std::fmt;
 use std::sync::{Arc, LazyLock, RwLock};
@@ -250,10 +251,64 @@ pub(crate) struct MentionsEntryConfig {
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
 #[serde(rename_all = "kebab-case")]
-#[serde(deny_unknown_fields)]
 pub(crate) struct RelabelConfig {
     #[serde(default)]
     pub(crate) allow_unauthenticated: Vec<String>,
+    // alias identifier -> labels
+    #[serde(flatten)]
+    pub(crate) aliases: HashMap<String, RelabelAliasConfig>,
+}
+
+impl RelabelConfig {
+    pub(crate) fn retrieve_command_from_alias(&self, input: RelabelCommand) -> RelabelCommand {
+        let mut deltas = vec![];
+        if !self.aliases.is_empty() {
+            // parse all tokens: if one matches an alias, extract the labels
+            // else, it will assumed to be a valid label
+            for tk in input.0.into_iter() {
+                let name = tk.label() as &str;
+                if let Some(alias) = self.aliases.get(name) {
+                    let cmd = alias.to_command(matches!(tk, LabelDelta::Remove(_)));
+                    deltas.extend(cmd.0);
+                } else {
+                    deltas.push(tk);
+                }
+            }
+        }
+        RelabelCommand(deltas)
+    }
+}
+
+#[derive(Default, PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub(crate) struct RelabelAliasConfig {
+    /// Labels to be added
+    pub(crate) add_labels: Vec<String>,
+    /// Labels to be removed
+    pub(crate) rem_labels: Vec<String>,
+}
+
+impl RelabelAliasConfig {
+    /// Translate a RelabelAliasConfig into a RelabelCommand for GitHub consumption
+    fn to_command(&self, inverted: bool) -> RelabelCommand {
+        let mut deltas = Vec::new();
+        let mut add_labels = &self.add_labels;
+        let mut rem_labels = &self.rem_labels;
+
+        // if the polarity of the alias is inverted, swap labels before parsing the command
+        if inverted {
+            std::mem::swap(&mut add_labels, &mut rem_labels);
+        }
+
+        for l in add_labels.iter() {
+            deltas.push(LabelDelta::Add(Label(l.into())));
+        }
+        for l in rem_labels.iter() {
+            deltas.push(LabelDelta::Remove(Label(l.into())));
+        }
+        RelabelCommand(deltas)
+    }
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -761,11 +816,11 @@ mod tests {
 
             [mentions."src/"]
             cc = ["@someone"]
-            
+
             [mentions."target/"]
             message = "This is a message."
             cc = ["@someone"]
-            
+
             [mentions."#[rustc_attr]"]
             type = "content"
             message = "This is a message."
@@ -835,6 +890,7 @@ mod tests {
             Config {
                 relabel: Some(RelabelConfig {
                     allow_unauthenticated: vec!["C-*".into()],
+                    aliases: HashMap::new()
                 }),
                 assign: Some(AssignConfig {
                     warn_non_default_branch: WarnNonDefaultBranchConfig::Simple(false),
@@ -1034,6 +1090,76 @@ mod tests {
     }
 
     #[test]
+    fn relabel_alias_config() {
+        let config = r#"
+            [relabel.to-stable]
+            add-labels = ["regression-from-stable-to-stable"]
+            rem-labels = ["regression-from-stable-to-beta", "regression-from-stable-to-nightly"]
+        "#;
+        let config = toml::from_str::<Config>(&config).unwrap();
+
+        let mut relabel_configs = HashMap::new();
+        relabel_configs.insert(
+            "to-stable".into(),
+            RelabelAliasConfig {
+                add_labels: vec!["regression-from-stable-to-stable".to_string()],
+                rem_labels: vec![
+                    "regression-from-stable-to-beta".to_string(),
+                    "regression-from-stable-to-nightly".to_string(),
+                ],
+            },
+        );
+
+        let expected_cfg = RelabelConfig {
+            allow_unauthenticated: vec![],
+            aliases: relabel_configs,
+        };
+
+        assert_eq!(config.relabel, Some(expected_cfg));
+    }
+
+    #[test]
+    fn relabel_alias() {
+        // [relabel.my-alias]
+        // add-labels = ["Alpha"]
+        // rem-labels = ["Bravo", "Charlie"]
+        let relabel_cfg = RelabelConfig {
+            allow_unauthenticated: vec![],
+            aliases: HashMap::from([(
+                "my-alias".to_string(),
+                RelabelAliasConfig {
+                    add_labels: vec!["Alpha".to_string()],
+                    rem_labels: vec!["Bravo".to_string(), "Charlie".to_string()],
+                },
+            )]),
+        };
+
+        // @triagebot label my-alias
+        let deltas = vec![LabelDelta::Add(Label("my-alias".into()))];
+        let new_input = relabel_cfg.retrieve_command_from_alias(RelabelCommand(deltas));
+        assert_eq!(
+            new_input,
+            RelabelCommand(vec![
+                LabelDelta::Add(Label("Alpha".into())),
+                LabelDelta::Remove(Label("Bravo".into())),
+                LabelDelta::Remove(Label("Charlie".into())),
+            ])
+        );
+
+        // @triagebot label -my-alias
+        let deltas = vec![LabelDelta::Remove(Label("my-alias".into()))];
+        let new_input = relabel_cfg.retrieve_command_from_alias(RelabelCommand(deltas));
+        assert_eq!(
+            new_input,
+            RelabelCommand(vec![
+                LabelDelta::Add(Label("Bravo".into())),
+                LabelDelta::Add(Label("Charlie".into())),
+                LabelDelta::Remove(Label("Alpha".into())),
+            ])
+        );
+    }
+
+    #[test]
     fn issue_links_uncanonicalized() {
         let config = r#"
             [issue-links]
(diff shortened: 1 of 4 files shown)


src/handlers/relabel.rs
```diff
@@ -53,14 +60,12 @@ pub(super) async fn handle_command(
                 });
             }
             LabelDelta::Remove(label) => {
-                results.push((
-                    label,
-                    event.issue().unwrap().remove_label(&ctx.github, &label),
-                ));
+                to_rem.push(label);
             }
```
apiraino: here I changed a bit the logic: instead of storing the futures I just store the labels to be removed and then execute the futures later.  Functionally I *think* it changes nothing but the original version (introduced in b6ccaa0fb46f815d3dbf56c9a8e9a10b6d796171) created a lifetime error after my my patch.

src/handlers/relabel.rs
```diff
@@ -24,9 +24,16 @@ pub(super) async fn handle_command(
...
     input: RelabelCommand,
 ) -> anyhow::Result<()> {
-    let mut results = vec![];
+    let mut to_rem = vec![];
     let mut to_add = vec![];
-    for delta in &input.0 {
+
+    // If the input matches a valid alias, read the [relabel] config.
+    // if any alias matches, extract the alias config (RelabelRuleConfig) and build a new RelabelCommand.
+    // Discard anything after the alias.
```
Urgau: I don't think we should discard anything, at least not silently.

In my opinion we should simply expand all the aliases, and if someone does `@rustbot label +label +alias +other-label` I would expect both `label` and `other-label` to be added + whatever `alias` expands to.

src/config.rs
```diff
@@ -238,10 +239,62 @@ pub(crate) struct MentionsPathConfig {
...
+    pub(crate) configs: Option<HashMap<String, RelabelRuleConfig>>,
+}
+
+impl RelabelConfig {
+    pub(crate) fn retrieve_command_from_alias(&self, input: RelabelCommand) -> RelabelCommand {
+        match &self.configs {
+            Some(configs) => {
+                dbg!(&configs);
+                // get only the first token from the command
+                // extract the "alias" from the RelabelCommand
```
Urgau: What's the reason for only considering the first token as an alias?

I would expect aliases to work at any position.

src/handlers/relabel.rs
apiraino: When I implemented this patch I thought about allowing both labels and aliases but ultimately discarded the option. My reasoning was that it would open up to contradicting labelling cases:
```
[config.alias]
add-label = ["P-high"]
rem-label = ["I-prioritize"]

# should I-prioritize stay or not?
@rustbot label alias I-prioritize
```

I think a user should *either* send ONE command alias *or* a set of labels, not both. If a user wants to send many labels, they can configure an alias.

src/config.rs
apiraino: see previous comment

src/handlers/relabel.rs
Urgau: We already have the issue when someone does `@rustbot labels I-prioritize -I-prioritize I-prioritize`, currently we adds all the labels first and then removes the other one (resulting in always no label, no matter the position).

IMO the behavior should be made consistent by using the precedence, from left to right, so in either case (your example and mine) `I-prioritize` would be added to the issue.

But whatever the behavior I don't this issue as being big enough compare to not allow regular labels and labels to be mixed together, which would just looked like an arbitrary limitation. I could very well see my-self wanting to do `@rustbot labels +diagnostic-alias +F-my-feature` (where `diagnostic-alias = +T-compiler +A-diagnostics`) and expecting it to work.

*We could also error out if the alias and labels conflict, it's less pretty but still an option.*

apiraino: This is now fixed by you in #2189 

(note to self: after merge of this PR, I will update the documentation. I would like to also mention the new left-to-right ordering of labels parsing - might be not obvious)

src/config.rs
```diff
@@ -249,10 +250,68 @@ pub(crate) struct MentionsEntryConfig {
...
 pub(crate) struct RelabelConfig {
     #[serde(default)]
     pub(crate) allow_unauthenticated: Vec<String>,
+    // alias identifier -> labels
+    #[serde(flatten)]
+    pub(crate) configs: Option<HashMap<String, RelabelRuleConfig>>,
+}
+
+impl RelabelConfig {
+    pub(crate) fn retrieve_command_from_alias(&self, input: RelabelCommand) -> RelabelCommand {
```
apiraino, line 263: @Urgau this should now implement the token parsing as you suggested: we parse each token, if it's an alias it is "exploded" into labels (according to the triagebot configuration). A new command is returned to the handler

src/config.rs
```diff
@@ -250,10 +250,68 @@ pub(crate) struct MentionsEntryConfig {
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
 #[serde(rename_all = "kebab-case")]
-#[serde(deny_unknown_fields)]
 pub(crate) struct RelabelConfig {
     #[serde(default)]
     pub(crate) allow_unauthenticated: Vec<String>,
+    // alias identifier -> labels
+    #[serde(flatten)]
+    pub(crate) configs: Option<HashMap<String, RelabelRuleConfig>>,
```
Urgau: I'm pretty sure we don't need it for to be an `Option`, we can just use `.is_empty()`.

Urgau: Let's maybe call it `aliases`, since that's what we are calling it everywhere else.
```suggestion
    pub(crate) aliases: HashMap<String, RelabelRuleConfig>,
```

src/config.rs
```diff
@@ -250,10 +250,68 @@ pub(crate) struct MentionsEntryConfig {
...
+                            deltas.push(tk.clone());
+                        }
+                    }
+                }
+            }
+            None => {
+                // nothing to do, return the original command
+                return input;
+            }
+        };
```
Urgau: I think we can simplify all of this logic a bit and avoid some clones.

```suggestion
    for tk in input.0.into_iter() {
        let name = tk.label().as_str();
        if let Some(alias) = self.aliases.get(name) {
            let cmd = cfg.to_command();
            deltas.extend(cmd.0);
        } else {
            deltas.push(tk);
        }
    }
```

src/config.rs
```diff
@@ -250,10 +250,68 @@ pub(crate) struct MentionsEntryConfig {
...
+        RelabelCommand(deltas)
+    }
+}
+
+#[derive(Default, PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub(crate) struct RelabelRuleConfig {
```
Urgau: `Rule` is a bit vague, `Alias` maybe?

```suggestion
pub(crate) struct RelabelAliasConfig {
```

src/github.rs
```diff
@@ -1338,9 +1340,6 @@ impl IssuesEvent {
     }
 }
 
-#[derive(Debug, serde::Deserialize)]
-struct PullRequestEventFields {}
-
 #[derive(Debug, serde::Deserialize)]
 pub struct WorkflowRunJob {
```
Urgau, line 1345: 👍

src/config.rs
```diff
@@ -250,10 +250,68 @@ pub(crate) struct MentionsEntryConfig {
...
+impl RelabelRuleConfig {
+    /// Translate a RelabelRuleConfig into a RelabelCommand for GitHub consumption
+    pub fn to_command(&self) -> RelabelCommand {
+        let mut deltas = Vec::new();
+        for l in self.add_labels.iter() {
+            deltas.push(LabelDelta::Add(Label(l.into())));
+        }
+        for l in self.rem_labels.iter() {
+            deltas.push(LabelDelta::Remove(Label(l.into())));
+        }
```
Urgau: This doesn't take into account the polarity of the alias it-self.

If someone does `@rustbot label -alias` then all the `add_labels` should be removed and all the `rem_labels` should be added.

We can maybe do it like that.

```suggestion
        let mut add_labels = self.add_labels;
        let mut rem_labels = self.rem_labels;
        
        // if the polarity is inverted swap the add and removed labels
        if inverted {
            std::mem::swap(&mut add_labels, &mut rem_labels);
        }
        
        for l in self.add_labels.iter() {
            deltas.push(LabelDelta::Add(Label(l.into())));
        }
        for l in self.rem_labels.iter() {
            deltas.push(LabelDelta::Remove(Label(l.into())));
        }
```
with `tk.to_command(matches!(tk, LabelDelta::Remove(_)))`

Urgau (commented):
Looks pretty good to me. Mainly some small nits.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2172/820288bf738424050e13c455bd3915a077eec711..167fdcbdc19f7c32773886e5e3720ade4000db30)*

src/config.rs
apiraino: I *think* my impl should be correct but the test fail. Can you please have a look at commit c00f4f867fce5b003fa147f2c5e6430e841abdf9? I'm not completely sure I understood how you wanted to impl this.

Urgau: My suggestion still used `self.add_labels` and `self.rem_labels` in the for loops instead of the local variables `add_labels` and `rem_labels` (ie without `self.`).

```diff
diff --git a/src/config.rs b/src/config.rs
index d5e3a57..1d2c830 100644
--- a/src/config.rs
+++ b/src/config.rs
@@ -301,10 +301,10 @@ impl RelabelAliasConfig {
             std::mem::swap(&mut add_labels, &mut rem_labels);
         }

-        for l in self.add_labels.iter() {
+        for l in add_labels.iter() {
             deltas.push(LabelDelta::Add(Label(l.into())));
         }
-        for l in self.rem_labels.iter() {
+        for l in rem_labels.iter() {
             deltas.push(LabelDelta::Remove(Label(l.into())));
         }
         RelabelCommand(deltas)
@@ -1160,9 +1160,9 @@ mod tests {
         assert_eq!(
             new_input,
             RelabelCommand(vec![
-                LabelDelta::Remove(Label("Alpha".into())),
                 LabelDelta::Add(Label("Bravo".into())),
                 LabelDelta::Add(Label("Charlie".into())),
+                LabelDelta::Remove(Label("Alpha".into())),
             ])
         );
     }
```

Btw, while testing I also saw some weird thing with unstable feature, I fixed them with `let name = tk.label() as &str;` locally. (Don't forget to rebase since there is a confict)

src/config.rs
```diff
@@ -725,6 +780,8 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
+    use parser::error::Error;
+    use parser::token::Tokenizer;
```
apiraino: I had to make the `token` module public to access it in these tests. let me know if you don't agree with this change.

src/config.rs
```diff
@@ -1033,6 +1091,82 @@ mod tests {
...
+        };
+
+        assert_eq!(config.relabel, Some(expected_cfg));
+    }
+
+    #[cfg(test)]
+    fn parse<'a>(input: &'a str) -> Result<Option<Vec<LabelDelta>>, Error<'a>> {
+        let mut toks = Tokenizer::new(input);
+        Ok(RelabelCommand::parse(&mut toks)?.map(|c| c.0))
+    }
```
Urgau: Instead of parsing the relabel command from the parser, could we just have create the deltas our-selves?

That would avoid the need to expose the internal of the parser (avoid exposing `Tokenizer`).


```suggestion
```

src/config.rs
```diff
@@ -1033,6 +1091,82 @@ mod tests {
...
+                "my-alias".to_string(),
+                RelabelAliasConfig {
+                    add_labels: vec!["Alpha".to_string()],
+                    rem_labels: vec!["Bravo".to_string(), "Charlie".to_string()],
+                },
+            )]),
+        };
+
+        // @triagebot label my-alias
+        let deltas = parse("label my-alias").unwrap().unwrap();
```
Urgau:
```suggestion
        let deltas = vec[LabelDelta::Add(Label("my-alias".into()))];
```

src/config.rs
```diff
@@ -1033,6 +1091,82 @@ mod tests {
...
+            new_input,
+            RelabelCommand(vec![
+                LabelDelta::Add(Label("Alpha".into())),
+                LabelDelta::Remove(Label("Bravo".into())),
+                LabelDelta::Remove(Label("Charlie".into())),
+            ])
+        );
+
+        // @triagebot label -my-alias
+        let deltas = parse("label -my-alias").unwrap().unwrap();
```
Urgau:
```suggestion
        let deltas = vec[LabelDelta::Remove(Label("my-alias".into()))];
```

parser/src/lib.rs
```diff
@@ -2,7 +2,7 @@ pub mod command;
 pub mod error;
 mod ignore_block;
 mod mentions;
-mod token;
+pub mod token;
```
Urgau:
```suggestion
mod token;
```

src/config.rs
```diff
@@ -250,10 +251,64 @@ pub(crate) struct MentionsEntryConfig {
...
+pub(crate) struct RelabelAliasConfig {
+    /// Labels to be added
+    pub(crate) add_labels: Vec<String>,
+    /// Labels to be removed
+    pub(crate) rem_labels: Vec<String>,
+}
+
+impl RelabelAliasConfig {
+    /// Translate a RelabelAliasConfig into a RelabelCommand for GitHub consumption
+    pub fn to_command(&self, inverted: bool) -> RelabelCommand {
```
Urgau: nit:
```suggestion
    fn to_command(&self, inverted: bool) -> RelabelCommand {
```

Urgau (commented):
Looks good to me, modulo the unnecessary exposure of the parser internals.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2172/692de911b7655b2dc5aa799245be8bcb18916b01..c82efd6d810349f888383217f44175bfa3cf9d7c)*

Issue #2173, opened 2025-09-03
Make warning of issue links in commit message configurable


weihanglo:
`[issue-links]` warns you when you have a issue link in a commit message, regardless it is canonicalized link like `rust-lang/rust#12345` or just `#12345`. 

There is an option `issue-links.check-commits` but it disables the entire check. We'd like to have an option to allow canonicalized link in commit messages and stop discouraging issue links.

PR #2175, opened 2025-09-04
Add support for only checking uncanonicalized issue links in commits

Urgau:
This PR adds support for only checking uncanonicalized issue links in commits.

```toml
[issue-links]
check-commits = "uncanonicalized"
```

Fixes https://github.com/rust-lang/triagebot/issues/2173

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -521,8 +521,62 @@ pub(crate) struct RenderedLinkConfig {
 #[serde(rename_all = "kebab-case")]
 #[serde(deny_unknown_fields)]
 pub(crate) struct IssueLinksConfig {
-    #[serde(default = "default_true")]
-    pub(crate) check_commits: bool,
+    #[serde(default)]
+    pub(crate) check_commits: IssueLinksCheckCommitsConfig,
+}
+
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -521,8 +521,54 @@ pub(crate) struct RenderedLinkConfig {
...
 pub(crate) struct IssueLinksConfig {
-    #[serde(default = "default_true")]
-    pub(crate) check_commits: bool,
+    #[serde(default)]
+    pub(crate) check_commits: IssueLinksCheckCommitsConfig,
+}
+
+#[derive(PartialEq, Eq, Debug)]
+pub(crate) enum IssueLinksCheckCommitsConfig {
+    Activated(bool),
```
Kobzol: Maybe `Off` / `Uncanonicalized` / `All` would be a bit easier to understand? It took me a while to grok what this enum means. But not a big deal.

Kobzol (approved):
Feel free to merge with/without the nit.

src/config.rs
Urgau: Adjusted the code as suggested, and added some comments to improve the meaning.

Issue #1950, opened 2025-04-22
Bring clippy into ci for cleaner code


xizheyin:
I was modifying the code and realized that there are a lot of places where style could be improved, I scanned it with `clippy` and found a very large number of warnings. is there a need to bring `clippy` into CI?

apiraino:
Someone already sent a while ago a patch to fix clippy lints #1758

IMHO (I am not a maintainer of this repo) I think the first wave of fixes should be manual and carefully reviewed because things may break. Also we probably don't want to apply blindly *any* clippy lint.

xizheyin:
Make sense, and it would be better if each lint was divided into different commits. I do think there is a need to do some of that. For example, I had some unused code in my PR submission that would not be detected, and I discovered this manually. I don't know if it's a problem with `rust-analyzer`.

PR #2181, opened 2025-09-14
Run Clippy

ada4a:
For #1950

This turned out rather big... I'd like to make this easier to review, but I'm not sure how -- maybe split each lint's changes into its own commit?

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -225,7 +225,10 @@ pub async fn schedule_job(
         anyhow::bail!("Job {} does not exist in the current job list.", job_name);
     }
 
-    if let Err(_) = get_job_by_name_and_scheduled_at(db, job_name, &when).await {
+    if get_job_by_name_and_scheduled_at(db, job_name, &when)
+        .await
+        .is_err()
+    {
         // mean there's no job already in the db with that name and scheduled_at
         insert_job(db, job_name, &when, &job_metadata).await?;
     }


Urgau:
> This turned out rather big... I'd like to make this easier to review, but I'm not sure how -- maybe split each lint's changes into its own commit?

If you can split into smaller commits, that would be great. Not necessarily each lint into it's own commit, bundling multiple lints changes is fine.

ada4a:
I did try to group related lints, but there were many small commits left from lints that I couldn't find a good group for -- hopefully, having the lint name on those commits will at least make their intent clear.

Urgau:
Thanks for splitting into many commits, that will help a lot.

I will take a look at the PR in the next new weeks *(I'm on vacation for the next two weeks)*.

flake.nix
Urgau, line 1: Could you remove the Nix changes from this PR, we are discussing those at #2179.

src/gh_range_diff.rs
```diff
@@ -536,9 +537,7 @@ impl UnifiedDiffPrinter for HtmlDiffPrinter<'_> {
             // Same number of lines before and after, can do word-hightling.
 
             // Diff the individual lines together.
-            let diffs_and_inputs: Vec<_> = before
-                .into_iter()
-                .zip(after.into_iter())
+            let diffs_and_inputs: Vec<_> = iter::zip(before, after)
```
Urgau, line 541: This is very pedantic :smile:

src/handlers/major_change.rs
```diff
@@ -382,29 +382,29 @@ async fn handle(
...
+        @rustbot resolve reason-for-concern
+        ```
+        See documentation at [https://forge.rust-lang.org](https://forge.rust-lang.org/compiler/proposals-and-stabilization.html#what-kinds-of-comments-should-go-on-a-mcp-in-the-compiler-team-repo)
+
+        </p>
+        </details>
+        {}
+
+        [stream]: {topic_url}",
             config.open_extra_text.as_deref().unwrap_or_default(),
```
Urgau, line 395: Is this actually the same thing? The other ones were not indented.

src/handlers/check_commits/issue_links.rs
```diff
@@ -30,8 +31,10 @@ pub(super) fn issue_links_in_commits(
                 .any(|i| c.commit.message.starts_with(i))
         })
         .filter(|c| does_match(&c.commit.message))
-        .map(|c| format!("- {}\n", c.sha))
-        .collect::<String>();
+        .fold(String::new(), |mut commits, c| {
+            _ = writeln!(commits, "- {}", c.sha);
+            commits
+        });
```
Urgau: Unsure about `clippy::format_collect`, I'm not finding it particularly more readable compared to before, and it's not like this code is perf sensitive anyway.

src/github.rs
```diff
@@ -1533,7 +1528,7 @@ impl Repository {
             .map(|(key, val)| format!("{key}:{val}"))
             .chain(include_labels.iter().map(|label| format!("label:{label}")))
             .chain(exclude_labels.iter().map(|label| format!("-label:{label}")))
-            .chain(std::iter::once(format!("repo:{}", self.full_name)))
+            .chain([format!("repo:{}", self.full_name)])
```
Urgau: Nice trick.

src/github.rs
```diff
@@ -618,7 +619,7 @@ pub(crate) struct UnknownLabels {
 // NOTE: This is used to post the Github comment; make sure it's valid markdown.
 impl fmt::Display for UnknownLabels {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Unknown labels: {}", &self.labels.join(", "))
+        write!(f, "Unknown labels: {}", self.labels.iter().format(", "))
```
Urgau: Hum, I'm not finding this `iter().format(...)` dance better than `.join(...)`. I'm not sure it's worth it.

src/rfcbot.rs
```diff
@@ -3,15 +3,15 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct FCP {
+pub struct Fcp {
```
Urgau: `FCP` is the correct name, it stands for Final Comment Period, it should not be lower-cased.

src/rfcbot.rs
```diff
@@ -30,7 +30,7 @@ pub struct Concern {
     pub reviewer: Reviewer,
 }
 #[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct FCPIssue {
+pub struct FcpIssue {
```
Urgau: Same here.

src/rfcbot.rs
```diff
@@ -60,18 +60,18 @@ pub struct StatusComment {
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct FullFCP {
-    pub fcp: FCP,
+pub struct FullFcp {
```
Urgau: And here.

src/gh_range_diff.rs
```diff
@@ -446,7 +446,6 @@ struct HtmlDiffPrinter<'a>(pub &'a Interner<&'a str>);
 
 impl HtmlDiffPrinter<'_> {
     fn handle_hunk_line<'a>(
-        &self,
```
Urgau, line 448: Can we just allow `clippy::unused_self` here, we may use `&self` here in the future.

Urgau (commented):
This looks pretty good. Thanks for doing that.

Left some changes and a few nits.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2181/885b6ed6e7c624bdd376e1dc1b7640a14d988352..45d88f87d0fd444d9b287e190ac753080ded6863)*

src/rfcbot.rs
```diff
@@ -3,15 +3,15 @@ use serde::{Deserialize, Serialize};
...
     pub id: u32,
     pub fk_issue: u32,
     pub fk_initiator: u32,
     pub fk_initiating_comment: u64,
     pub disposition: Option<String>,
     pub fk_bot_tracking_comment: u64,
-    pub fcp_start: Option<String>,
-    pub fcp_closed: bool,
+    pub start: Option<String>,
+    pub closed: bool,
```
Urgau: Those field name changes needs to be reverted, the are coming from a API, they need to stay the same as the api.

src/gh_range_diff.rs
ada4a: I must admit this one isn't even a Clippy lint, but rather a personal preference:
- I've often found the formatting of `.zip` to be suboptimal, where, especially when the to-be zipped iterators were built as a method chain of iterator adapters, they would get split up across many lines, making it hard to see what exactly we're zipping. For example, something like this:
    ```rs
    left
        .iter()
        .enumerate()
        .zip(right.into_iter().map(|x| x + 2))
    ```
    Is imo much more understandable when written like this:
    ```rs
    iter::zip(
        left.iter().enumerate()
        right.into_iter().map(|x| x + 2)
    )
    ```
- And as shown in this particular case, `iter::zip` allows for replacing the explicit `x.iter()` with `&x`, and removing `.into_iter()` entirely, which I also find pretty nice.

But again, this change in particular doesn't bring us closer to long-term goal behind this PR (have Clippy run as part of CI), so I could just remove it..

src/handlers/major_change.rs
ada4a: Oh, no, sorry, that's an annoying bug in the Git merge driver that I'm co-developing – I really should get to fixing it...

src/gh_range_diff.rs
Urgau: I have no preference for either, so I'm fine with keeping your version.

src/rfcbot.rs
ada4a: I'm aware^^ This is from [`clippy::upper_case_acronyms`](https://rust-lang.github.io/rust-clippy/master/index.html#upper_case_acronyms), which specifically targets acronyms

src/handlers/check_commits/issue_links.rs
ada4a: Yeah, you're right -- not to mention that `clippy::format_collect` is `pedantic`.

I could theoretically use `Itertools::format_with` instead:
```rs
.format_with("\n", |c, f| f(&format_args!("- {}", c.sha)))
.to_string();
```
But it's also not the most readable imo. WDYT?

src/github.rs
ada4a: Fair enough. I removed all the instances where the `.iter()`/`.into_iter()` was needed, leaving only those where we already were working with an iterator.

src/handlers/check_commits/issue_links.rs
Urgau: Yeah, I also don't find it more readable IMO.

src/rfcbot.rs
ada4a: @Urgau friendly ping in case this slipped through the cracks -- should I revert this change or leave it in?

Urgau: Sorry, I didn't realized this was waiting for me. Yes, please do revert this change, `FCP` is the canonical name and it should stay that way.

PR #2184, opened 2025-09-21
Improve range-diff color contrast in dark mode

### LINKED ISSUE #2183: The dimmed colors of our range-diff are too dimed in dark mode
Reported at [#t-compiler > Experimental range-diff for force-push @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/Experimental.20range-diff.20for.20force-push/near/540672308) by @saethlin.

We should make less dimmed, probably by "stealing" `git range-diff` colors instead of using our own colors.

Image

Noratrieb:
The current colors are very low contrast, and therefore very hard to read. This replaces the colors with significantly brighter ones to improve contrast a lot.

The red drifted a lot into the orange, because when making the red brighter, it just got *harder* to read for me because red on black is bad combination. So i just made it slightly orange, which looks a lot better. I did not pick the colors from `git range-diff` itself because the red there is still pretty low contrast.

The colors now have a comfortable ~4.55 (AA accessibility) contrast according to developer tools.

image

fixes #2183

### DIFF
--- a/src/gh_range_diff.rs
+++ b/src/gh_range_diff.rs
@@ -303,7 +303,7 @@ fn process_old_new(
         color: rgba(0, 255, 0, 1);
       }}
       .line-removed-before {{
-        color: rgb(206, 48, 0);
+        color: rgb(255, 159, 131);
       }}
       .line-added-before {{
         color: rgba(11, 142, 0, 1);


Noratrieb:
I don't think we have much if any acceptable range of brightness for red on black.

Urgau:
I tried playing with some red colors but the contrast checker wasn't very happy. The only thing I can think of would be to go for more brown-ish colors (`rgb(185, 98, 72)` in the example below), there the distinction is more pronounced.

image

Noratrieb:
Yeah I think changing the hue is a good strategy, I also went a bit into the orange originally. I think I'd make that brown a bit brighter, I don't find the contrast very good. But something along those lines should work probably and give the visible contrast between before and after.

Urgau:
We experimented in private and found that pastel colors provided more contrast with the bright red and black background, so let's go with that. `rgb(255, 159, 131)`

image

src/gh_range_diff.rs
```diff
@@ -303,10 +303,10 @@ fn process_old_new(
         color: rgba(0, 255, 0, 1);
       }}
       .line-removed-before {{
-        color: rgba(100, 0, 0, 1);
+        color: rgb(206, 48, 0);
```
Urgau:
```suggestion
        color: rgb(255, 159, 131);
```

PR #2188, opened 2025-09-29
Fix GitHub base change de-serialization

Urgau:
Looks like our `base` field for the "changes" in an issue/PR has always been wrong, it has two additional fields. Found while looking at the logs.

https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=edited#pull_request


Kobzol (approved):
Oops :D

PR #2189, opened 2025-09-30
Apply labels from left to right in relabel

Urgau:
As we discussed in #2172 and in [#t-compiler/contrib-private > RFC: triagebot parsing command aliases](https://rust-lang.zulipchat.com/#narrow/channel/244344-t-compiler.2Fcontrib-private/topic/RFC.3A.20triagebot.20parsing.20command.20aliases/with/541247373) the way labels are applied in relabel is weird when the same label is used multiple times.

This PR makes the behavior more user friendly by pre-computing a "final state" where all the label deltas are applied from left to right, meaning that `+label -label`/`-label +label` would cancel each other instead of using the issue state.

### DIFF
--- a/src/handlers/relabel.rs
+++ b/src/handlers/relabel.rs
@@ -27,6 +27,11 @@ pub(super) async fn handle_command(
     event: &Event,
     input: RelabelCommand,
 ) -> anyhow::Result<()> {
+    let Some(issue) = event.issue() else {
+        anyhow::bail!("event is not an issue");
+    };
+
+    // Check label authorization for the current user
     for delta in &input.0 {
         let name = delta.label().as_str();
         let err = match check_filter(name, config, is_member(&event.user(), &ctx.team).await) {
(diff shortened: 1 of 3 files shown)


src/handlers/relabel.rs
```diff
@@ -182,10 +177,41 @@ fn match_pattern(pattern: &str, label: &str) -> anyhow::Result<MatchPatternResul
     })
 }
 
+fn compute_label_deltas(deltas: &[LabelDelta]) -> (Vec<Label>, Vec<Label>) {
+    let mut add = BTreeSet::new();
+    let mut remove = BTreeSet::new();
+
+    for delta in deltas {
```
Kobzol, line 178: Just a note to myself: this implementation means that doing `+foo -foo -foo` will actually remove `foo` from the issue. This is probably useful for something like `alias -foo`, where `alias` does `+foo -foo`.

Urgau: Yeah, and it also means that `+foo +foo -foo` does nothing (`+foo +foo` -> `+foo` and `+foo -foo` cancels each other out).

src/handlers/relabel.rs
```diff
@@ -24,8 +27,11 @@ pub(super) async fn handle_command(
     event: &Event,
     input: RelabelCommand,
 ) -> anyhow::Result<()> {
-    let mut results = vec![];
-    let mut to_add = vec![];
+    let Some(issue) = event.issue() else {
+        anyhow::bail!("event is not an issue");
+    };
```
apiraino, line 32: @Urgau why did you introduce this block?

I don't know for other people but I need to be able to edit labels on both pull requests *and* issues.

Kobzol: PRs are also issues, so this should work fine. The code before did an unwrap on event.issue() anyway.

apiraino: ah wait maybe I am misunderstanding: `Issue` here is meant as both a "ticket" and a pull request so [Event::Issue](https://github.com/rust-lang/triagebot/blob/master/src/github.rs#L2413) covers both.

But I still don't understand why adding this check: what other event are you preventing to happen? 

The error message at least is confusing, I think.

Urgau: > why did you introduce this block?

Because the code below was asserting the presence of it. It's better return a message than panic and block all the other handlers.

We should indeed probably change the message to be explicitly about PRs since it's now a user message.

apiraino: > We should indeed probably change the message to be explicitly about PRs since it's now a user message.

I'm working on #2172 I will change that as well

PR #2190, opened 2025-10-02
Improve removal of multiple GitHub labels at the same time

Urgau:
GitHub doesn't offer an API to remove multiple labels at the same time, this leads us to have to manually have for-loops to remove multiple of them.

Instead I propose that we paper over GitHub's API and do the multiple API calls (in parallel) in our helper, that will make the function consistent with `add_labels` and allow us to process the removal in parallel.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -785,33 +785,47 @@ impl Issue {
         Ok(())
     }
 
-    pub async fn remove_label(&self, client: &GithubClient, label: &str) -> anyhow::Result<()> {
-        log::info!("remove_label from {}: {:?}", self.global_id(), label);
-        // DELETE /repos/:owner/:repo/issues/:number/labels/{name}
-        let url = format!(
-            "{repo_url}/issues/{number}/labels/{name}",
-            repo_url = self.repository().url(client),
-            number = self.number,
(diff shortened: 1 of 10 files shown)


src/github.rs
```diff
@@ -785,33 +785,47 @@ impl Issue {
...
-        // DELETE /repos/:owner/:repo/issues/:number/labels/{name}
-        let url = format!(
-            "{repo_url}/issues/{number}/labels/{name}",
-            repo_url = self.repository().url(client),
-            number = self.number,
-            name = label,
+    pub async fn remove_labels(
+        &self,
+        client: &GithubClient,
+        labels: Vec<Label>,
```
Urgau, line 791: It's unclear to me if taking a `Vec<Label>` is the right choice, in most cases we need to convert `Vec<&str>`/`Vec<String>` to it. Maybe we should take an `impl IntoIterator<Item = &str>` or something like that.

Kobzol: I think it's fine, although I'd use `&[Label]`, but that doesn't matter that much (it would probably require more generic annotations).

Kobzol (approved):

PR #2191, opened 2025-10-02
Turn automatic backport nominations into suggestions

apiraino:
Beta backport nominations were confusing so we decided to turn them into suggestions for reviewers and author of a PR fixing a regression. The idea is to raise awareness on patches fixing P-high/P-critical regressions but at the same time not make it look like a *mandatory* decision for T-compiler to take.

After this patch, I will modify the template that creates Zulip topics for backport nominations. For example it could look like this:
```
PR #123456 "Fix insulation in flux capacitor" fixes a regression.
Cc: @user1, @user2, @user3 please evaluate nominating this PR for backport, thanks!
The following poll is a vibe-check and not binding.

/poll Should #123456 be beta backported?
approve
decline
don't know
```

<details><summary>(visual representation)</summary>
<p>

grafik

</p>
</details> 

(`@user1, @user2, @user3`) are PR author and reviewers.
 
As per [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/238009-t-compiler.2Fmeetings/topic/.5Bweekly.5D.202025-10-02/near/542758900) discussion.

### DIFF
--- a/src/handlers/notify_zulip.rs
+++ b/src/handlers/notify_zulip.rs
@@ -1,9 +1,12 @@
+use crate::github::User;
 use crate::zulip::api::Recipient;
+use crate::zulip::render_zulip_username;
 use crate::{
     config::{NotifyZulipConfig, NotifyZulipLabelConfig, NotifyZulipTablesConfig},
     github::{Issue, IssuesAction, IssuesEvent, Label},
     handlers::Context,
 };
+use futures::future::join_all;
 use tracing as log;
 
 pub(super) struct NotifyZulipInput {
@@ -214,10 +217,15 @@ pub(super) async fn handle_input<'a>(
                 topic: &topic,
             };
 
+            // Issue/PR authors/reviewers will receive a mention if the template has `{recipients}`.
+            let recipients = &mut event.issue.assignees.clone();
+            recipients.push(event.issue.user.clone());
+
             for msg in msgs {
                 let msg = msg.replace("{number}", &event.issue.number.to_string());
                 let msg = msg.replace("{title}", &event.issue.title);
                 let msg = replace_team_to_be_nominated(&event.issue.labels, msg);
+                let msg = msg.replace("{recipients}", &get_zulip_ids(ctx, &recipients).await);
 
                 let req = crate::zulip::MessageApiRequest {
                     recipient,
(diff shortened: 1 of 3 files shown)


src/handlers/notify_zulip.rs
```diff
@@ -214,10 +217,15 @@ pub(super) async fn handle_input<'a>(
                 topic: &topic,
             };
 
+            // Issue/PR authors/reviewers will receive a mention if the template has `{recipients}`.
+            let recipients = &mut event.issue.assignees.clone();
+            recipients.push(event.issue.user.clone());
+
```
apiraino, line 223: This is different from before: if the template has a `{recipients}` placeholder, author and reviewers of the PR/issue will receive a mention.

(this change is very specific to my use-case but I think it can be used in other contexts)

src/handlers/notify_zulip.rs
```diff
@@ -236,6 +244,23 @@ pub(super) async fn handle_input<'a>(
     Ok(())
 }
 
+async fn get_zulip_ids(ctx: &Context, recipients: &[User]) -> String {
```
apiraino, line 247: this code here is duplicating what we have in `src/zulip.rs` so a better way would be to refactor that module and expose functions. Maybe another time.

apiraino: to generate the mention it's using the trick to write `@|12345` where 12345 is the Zulip ID of the user.

Kobzol:
Let me know if you want a review.

src/handlers/notify_zulip.rs
```diff
@@ -236,6 +244,23 @@ pub(super) async fn handle_input<'a>(
...
+    let gh_ids_fut = recipients
+        .iter()
+        .map(|recipient| async move { ctx.team.github_to_zulip_id(recipient.id).await });
+    let zulip_ids = join_all(gh_ids_fut).await;
+
+    zulip_ids
+        .iter()
+        .map(|x| {
+            let id = x.as_ref().unwrap().unwrap();
```
Kobzol: This will crash the handler if either the PR author (more likely) or the assignee (less likely) do not have a Zulip ID in team. Could you please filter out users for which we could not determine the Zulip ID, instead of crashing?

apiraino: ah sorry that was a leftover I forgot to remove! Thanks for spotting it

src/handlers/notify_zulip.rs
apiraino: fixed (with not so good-looking code, open to suggestions)

src/handlers/notify_zulip.rs
apiraino: fixed (with not so good-looking code, open to suggestions)

Kobzol (approved):
Thanks, let's try.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2191/5e11c7bb1e27b878ac103f1c5d231e730074d392..63ae02e7e42e135c84dc2c7e15695ff727c6987b)*

PR #2192, opened 2025-10-03
Improve and simplify commands errors

Urgau:
I had to fought today to find the error in https://github.com/rust-lang/rustup/issues/2876#issuecomment-3365786430, mainly because CloudWatch is really bad, but also because we didn't post any error to the user.

This PR improves the way we deal with user facing error by creating a `inform!` macro that wraps `anyhow::bail!`+`UserError`. This allows two things, first that we always post an error message to the user and second that we log it (correctly).

This has also the side-effect that don't return 500 when we only have user facing errors. It also reduce the boilerplate of doing `ErrorComment`+`post()`+`return Ok(())` each time.

To put it in another way, let's use our the `HandlerError` infra the way it should be used.

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -10,9 +10,15 @@ use std::fmt;
 use std::sync::Arc;
 use tracing as log;
 
-macro_rules! inform {
+/// Creates a [`UserError`] with message.
+///
+/// Should be used when an handler is in error due to the user action's (not a PR,
+/// not a issue, not authorized, ...).
+///
+/// Should be used like this `return user_error!("My error message.");`.
+macro_rules! user_error {
     ($err:expr $(,)?) => {
-        anyhow::bail!(crate::handlers::UserError($err.into()))
+        anyhow::Result::Err(anyhow::anyhow!(crate::handlers::UserError($err.into())))
     };
 }
 
@@ -439,6 +445,9 @@ impl fmt::Display for HandlerError {
     }
 }
 
+/// Represent a user error.
+///
+/// The message will be shown to the user via comment posted by this bot.
 #[derive(Debug)]
 pub struct UserError(String);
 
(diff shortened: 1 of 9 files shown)


src/handlers.rs
```diff
@@ -10,21 +10,13 @@ use std::fmt;
...
-
-impl fmt::Display for HandlerError {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match self {
-            HandlerError::Message(msg) => write!(f, "{}", msg),
-            HandlerError::Other(_) => write!(f, "An internal error occurred."),
-        }
-    }
+macro_rules! inform {
+    (msg:literal $(,)?) => {
```
fmease: This arm is never taken because it expects `msg:literal` literally as you didn't prefix it with `$`. In any case, this whole arm is not really necessary because `expr` includes `literal` basically.

Urgau: Indeed, fixed.

src/handlers.rs
```diff
@@ -27,6 +27,26 @@ impl fmt::Display for HandlerError {
...
+
+impl std::error::Error for UserError {}
+
+impl fmt::Display for UserError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str(&self.0)
+    }
+}
+
+macro_rules! inform {
```
Kobzol: Could you add a comment on top of this that explains what it does, or rather how/when it should be used?

src/handlers.rs
```diff
@@ -27,6 +27,26 @@ impl fmt::Display for HandlerError {
     }
 }
 
+#[derive(Debug)]
+pub struct UserError(String);
```
Kobzol: Same here, let's add a comment that describes what is this for.

Kobzol (commented):
I'm not a big fan of control-flow being hidden inside macros, so I'd rather do `return user_error!(...);`. But I'm also fine with the current version.

src/handlers.rs
Urgau: Sure, comment added and renamed to `user_error!`.

Urgau:
Agree with you that the hidden control-flow was a bit weird. Changed to our proposed `user_error!` macro.

PR #2199, opened 2025-10-11
Make timestamps visually uncopyable and simplify the impl

Urgau:
This PR improves https://github.com/rust-lang/triagebot/pull/2198 by making the timestamps visually unselect-able/uncopyable.

This is done by putting the timestamps in a "floating" `::before` element, which has the consequence of making the element actually uncopyable.

This also simplifies the implementation of our copy event by removing the timestamps removal as they are no longer select-able. This also as the advantage that partial selection of a timestamp does not leave it in the copy.

image

### DIFF
--- a/src/gha_logs.rs
+++ b/src/gha_logs.rs
@@ -212,18 +212,29 @@ pub async fn gha_logs(
     <title>{job_name} - {owner}/{repo}@{short_sha}</title>
     {icon_status}
     <style>
+:root {{
+    --timestamp-length: 28ch;
+}}
 body {{
   font: 14px SFMono-Regular, Consolas, Liberation Mono, Menlo, monospace;
   background: #0C0C0C;
   color: #CCC;
   white-space: pre;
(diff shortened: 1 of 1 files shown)


src/gha_logs.rs
```diff
@@ -217,14 +217,22 @@ body {{
...
+  padding-left: 29ch; // timestamp length + one space
+  margin: 0;
 }}
 .timestamp {{
   color: #848484;
   text-decoration: none;
+  position: absolute;
+  left: 0;
+  width: 28ch; // timestamp length
 }}
```
fmease: You could write this more robustly (i.e., not encoding invariants in comments) like this:

```css
:root {
    --timestamp-length: 28ch;
}
body {
    /* … */
    padding-left: calc(var(--timestamp-length) + 1ch);
}
.timestamp {
    /* … */
    width: var(--timestamp-length);
}
```

(I haven't checked if you could simplify the added CSS rules overall; I'm just drive-by reviewing)

PR #2200, opened 2025-10-13
Prefer using an HTML table with `user-select: none` in GHA logs

Urgau:
When I tested #2199 I was a EuroRust on a not so great internet connection so I only tested on relatively "small" GHA logs (~30000 lines) and the perf was okay.

But when I went yesterday evening on https://triage.rust-lang.org/gha-logs/rust-lang/rust/52553156046 (which is ~80000 lines) the load of the page was very long, which isn't too surprising since the logs it-self is ~5Mio it-self, so I looked more in details as to why it is taking ~50s on my PC with Firefox to render the logs and realized that #2199 destroyed the perf by making the layout calculation in Firefox do much much more work. (Chrome is better but still slower than before)

To fix this perf regression and still make the timestamps not part of the copy, I switched the DOM to use an HTML table whose first column is the timestamps (which have `user-select: none` to prevent copy) and the second column is the log line.

For context, with #2199 the layout calculation was 39s and with this PR it's only ~2.5s on my PC.

Best review commit by commit.

image

### DIFF
--- a/src/gha_logs.rs
+++ b/src/gha_logs.rs
@@ -219,6 +219,7 @@ body {{
 }}
 .timestamp {{
   color: #848484;
+  user-select: none;
   text-decoration: none;
 }}
 .timestamp:hover {{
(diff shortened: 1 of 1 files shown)


src/gha_logs.rs
```diff
@@ -270,14 +271,19 @@ body {{
...
-            html = html.substr(1);
-        }}
-
-        // 3. Add a self-referencial anchor to all timestamps at the start of the lines
-        const dateRegex = /^(\d{{4}}-\d{{2}}-\d{{2}}T\d{{2}}:\d{{2}}:\d{{2}}\.\d+Z)/gm;
-        html = html.replace(dateRegex, (ts) => 
-            `<a id="${{ts}}" href="#${{ts}}" class="timestamp">${{ts}}</a>`
+        html = html.replace(/^\uFEFF/gm, "");
+        
+        // 3. Transform each log lines that doesn't start with a timestamp into a row where everything is in the second column
```
Kobzol, line 276: Now that I think about it, why don't we just generate the HTML structure on the backend, instead of doing it in this super hacky way on the frontend? Can `ansi_to_html` deal with pre-existing HTML tags?

src/gha_logs.rs
```diff
@@ -270,14 +271,19 @@ body {{
...
+        const untsRegex = /^(?!\d{{4}}-\d{{2}}-\d{{2}}T\d{{2}}:\d{{2}}:\d{{2}}\.\d+Z)(.*)$/gm;
+        html = html.replace(untsRegex, (match, log) => 
+            `<tr><td></td><td>${{log}}</td></tr>`
```
Kobzol, line 279: This could be `<td colspan="2">`, I think?

src/gha_logs.rs
Urgau: We tried in https://github.com/rust-lang/triagebot/pull/2125 with anstyle but the processing done by anstyle was an order of magnitude slower than what we currently use in JS.

src/gha_logs.rs
Urgau: I don't think so. According to https://www.w3schools.com/TAGS/tryit.asp?filename=tryhtml_td_colspan, `colspan` make a column span multiple column, which is not what I want here.

src/gha_logs.rs
Kobzol: I know, I meant to still do ANSI in JS, but generate the tables and the rest of the HTML structure on the backend.

src/gha_logs.rs
Kobzol: I think that is exactly what we want =D But an empty `<td>` works anyway.

src/gha_logs.rs
Urgau: Won't that cause problems with styles? like if a line is red but there is no reset at the end of it, the next line would still be red, but if we processed line by line the ANSI codes (as we would need to do if we pre-processed into a table) we would lose the context from the first line and so the second line wouldn't be red anymore

Kobzol: Yeah, sadly AnsiUp can't really deal with HTML input in the way I envisioned it. Nevermind.

src/gha_logs.rs
Urgau: > I think that is exactly what we want

There are multi-lines in the logs, and I don't think we wouldn't want the log line to be under the timestamps. We want the rest of the log lines to be under the other log lines in the second column only.

Like this:
image

Not like this:
image

Kobzol: Didn't think of that! Thanks for clarifying.

Issue #2067, opened 2025-06-07
Feature request: Auto-create a tracking issue for accepted MCPs


fmease:
Uplifted from https://github.com/rust-lang/triagebot/pull/2066/files#r2133761555.
This proposal is blocked on PR #2066.

I propose that triagebot should automatically create a T-compiler tracking issue over at rust-lang/rust for accepted (compiler) MCPs once their waiting period has passed (right before or after automatically closing the MCP, cc #2066).

**Motivation**: It would make it less likely for accepted MCPs to fall into oblivion which is a confirmed issue that's been discussed in the past (like maybe a year ago). For example, I don't even know if we have an accurate and up to date list of accepted but still unimplemented MCPs (I remember that there were motions to compile such a lists some time ago).

---

How should the generated tracking issue look like, roughly? Here's a quick mockup:

* **Title**: `Tracking issue for MCP <NUMBER>: <TITLE>`
  * `NUMBER`: The issue number in the rust-lang/compiler-team repo.
  * `TITLE`:  The title of the MCP.
* **Labels**: <kbd>C-tracking-issue</kbd> + <kbd>T-compiler</kbd> + [<kbd>B-MCP-approved</kbd>](https://github.com/rust-lang/rust/labels/B-MCP-approved) + [<kbd>S-tracking-unimplemented</kbd>](https://github.com/rust-lang/rust/labels/S-tracking-unimplemented)
* **Assignee**: (Maybe) the author of the MCP.

**Description**:

```md
This is a tracking issue for [MCP {number}](https://github.com/rust-lang/compiler-team/issues/{number}).

### About tracking issues

Tracking issues are used to record the overall progress of implementation.
They are also used as hubs connecting to other relevant issues, e.g., bugs or open design questions.
A tracking issue is however *not* meant for large scale discussion, questions, or bug reports about a feature.
Instead, open a dedicated issue for the specific matter and add the relevant feature gate label.
Discussion comments will get marked as off-topic or deleted.
Repeated discussions on the tracking issue may lead to the tracking issue getting locked.

### Steps

- [ ] Implement the MCP
- [ ] Adjust documentation

### Unresolved questions

*None so far.*

### Implementation history

*Empty so far.*
```

fmease:
Does this itself need an MCP 🤔😆

Kobzol:
[...]

PR #2202, opened 2025-10-17
Automatically create tracking issues when major change are accepted

Urgau:
This PR adds the required infrastructure to automatically create a tracking issues when a major change is accepted.

```toml
[major-change.tracking-issue-template]
repository = "triagebot"
title = "Tracking issue for MCP#${mcp_number}"
body = """
Multi line body with ${mcp_number} and ${mcp_title}
"""
labels = ["C-tracking-issue", "T-compiler"]
```

Fixes https://github.com/rust-lang/triagebot/issues/2067
cc @fmease

### DIFF
--- a/src/handlers/major_change.rs
+++ b/src/handlers/major_change.rs
@@ -519,7 +519,7 @@ impl Job for MajorChangeAcceptenceJob {
 
         let now = Utc::now();
 
-        match process_seconded(ctx, &major_change, now).await {
+        match try_accept_mcp(ctx, &major_change, now).await {
             Ok(()) => {
                 tracing::info!(
                     "{}: major change ({:?}) as been accepted",
@@ -549,7 +549,7 @@ impl Job for MajorChangeAcceptenceJob {
     }
 }
 
-async fn process_seconded(
+async fn try_accept_mcp(
     ctx: &super::Context,
     major_change: &MajorChangeSeconded,
     now: DateTime<Utc>,


src/handlers/major_change.rs
Kobzol: This function name confused me while I was reviewing this PR. I'd suggest renaming it to something like `try_accept_mcp`.

Kobzol (approved):
Implementation looks good!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2202/820288bf738424050e13c455bd3915a077eec711..07e16dea437d3add2287cf3d2aa2c47d39b476a2)*

src/handlers/major_change.rs
Urgau: Sure, done.

fmease, line 1: This looks great, thanks!

I don't want to smother you with extra work, so these things can be done as follow-ups (or not all :P): In my issue & MCP I also proposed to

* have some `suppress-auto-tracking` label to suppress the creation of the tracking issue.
  
  So it would be pretty great if one was able to register such a label in `[major-change.tracking-issue-template]` assuming you agree that such a mechanism would be useful (IIRC @Kobzol doubted the usefulness).
  
  That could come in clutch in situations like open MCP https://github.com/rust-lang/compiler-team/issues/924 where we already have a tracking issue: https://github.com/rust-lang/rust/issues/132181 (sort of). And IIRC my MCP / issue had some other vague examples?
* have a way to ping the MCP author (and registered mentors if technically feasible) since the link to the MCP can be easily missed; however, it's probably not worth the effort

Urgau: > * have some `suppress-auto-tracking` label to suppress the creation of the tracking issue

While I don't doubt the usefulness in some scenario I think they are going to be extremely limited. Even in your example I would still want to have the tracking issue created as they wouldn't track the same thing.

If we see the need for it we can always add it later.

> * have a way to ping the MCP author

Done, added the `${mcp_author}` variable to get the handle of the MCP issue author.

> * and registered mentors if technically feasible

That would require some sort of easily-parseble format for triagebot, it's currently a free for all. Let's defer that for now.

fmease (approved):

PR #2207, opened 2025-10-24
Only consider the line as a Zulip command in a stream

Urgau:
This PR updates our Zulip command handling logic to only consider the line where we were pinged to be part of the command and not everything else as we currently do.

This is necessary when the commands are invoked from a stream where the command may be interlaced between regular comments.

Spotted in [#**t-release>meeting 2025 10 24@546940454**](https://rust-lang.zulipchat.com/#narrow/channel/241545-t-release/topic/meeting.202025.2010.2024/near/546940454)

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -177,14 +177,15 @@ async fn process_zulip_request(ctx: Arc<Context>, req: Request) -> anyhow::Resul
 async fn handle_command<'a>(
     ctx: Arc<Context>,
     mut gh_id: u64,
-    command: &'a str,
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2207/842e2bd2d992353cfb61ebd9703406bf5c225eed..ee4022b496d1f76e8dd985d7f263a4ce0e7f2f4f)*

PR #2210, opened 2025-10-25
Improve our handling of user facing errors

Urgau:
Instead of having each user error (like `UnknownLabels`) be it's own error type, group all the user errors under the `UserError` type, which now becomes a enum.

This simplifies the handling as it means that we only have to catch one type `UserError`, and post a comment for it on the issue. It also avoids the need for each handler to catch all the user errors and do a manual handling (`issue_handlers!` and `command_handlers!` do that for them).

This also avoid accidentally eating the error instead of posting it to the user.

I have tested the changes on my own test repo with success.

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -22,7 +22,7 @@
 
 use crate::db::issue_data::IssueData;
 use crate::db::review_prefs::{RotationMode, get_review_prefs_batch};
-use crate::errors::{self, user_error};
+use crate::errors::{self, AssignmentError, user_error};
 use crate::github::UserId;
 use crate::handlers::pr_tracking::ReviewerWorkqueue;
 use crate::{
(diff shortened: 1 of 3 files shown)


src/handlers/assign.rs
```diff
@@ -656,11 +656,8 @@ pub(super) async fn handle_command(
                 e.apply(&ctx.github, String::new()).await?;
                 return Ok(());
             } // we are done
-            Err(github::AssignmentError::InvalidAssignee) => {
-                issue
-                    .set_assignee(&ctx.github, &ctx.username)
-                    .await
-                    .context("self-assignment failed")?;
+            Err(AssignmentError::InvalidAssignee) => {
+                issue.set_assignee(&ctx.github, &ctx.username).await?;
```
Kobzol, line 660: The `.context("self-assignment failed")?;` part could be left there, I suppose?

Kobzol (approved):
Thanks! This is quite nice.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2210/5424b88a23ca7ee938442a3d7b699d7f2d9107b7..87d1c444cce3fcdc1f69c0a9325dafcceee53843)*

src/handlers/assign.rs
Urgau: No, it cannot stay. `.context()` calls requires `AssignmentError` to implement the `Error` but implementing it conflicts with out `From<AssignmentError> for anyhow::Error` as anyhow implements `From` for every type that implements the `Error` trait.

Urgau: We could of course call `.into()` before, but that felt unnecessary IMO.

Kobzol: Yeah, it would have to be converted to `anyhow::Error` first. Nevermind.

PR #2211, opened 2025-10-28
Switch review labels for users with write perm when requesting changes

### LINKED ISSUE #2169: For `review-submitted` labels, don't require that the reviewer is assigned
Leaving a "request changes" review only applies the `S-waiting-on-review` label if the review comes from somebody who is assigned to the PR. However, this means we don't get that benefit on repos where auto-assign isn't enabled. Could this be loosened so a relabel happens when done by anybody with write access, rather than restricted to the assignee? I think removing the assignment check should suffice here since users without write access can't change leave the red/green reviews.

This could also be made configurable. IMO this probably isn't necessary since in practice leaving a "request changes" review usually always indicates action required, regardless of whether the assignee is up to date, and "comment" is used for things that aren't blocking.

Example PR: https://github.com/rust-lang/libc/pull/4615

Urgau:
This PR adjusts the `[review-submitted]` issue handler to also switch review labels for users with write permission when requesting changes.

Fixes #2169
cc @tgross35

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -554,6 +554,11 @@ pub enum Selection<'a, T: ?Sized> {
     Except(&'a T),
 }
 
+#[derive(Debug, serde::Deserialize)]
+pub struct CollaboratorPermission {
+    pub permission: String,
+}
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct IssueRepository {
     pub organization: String,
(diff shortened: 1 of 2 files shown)


src/github.rs
```diff
@@ -554,6 +554,11 @@ pub enum Selection<'a, T: ?Sized> {
     Except(&'a T),
 }
 
+#[derive(Debug, serde::Deserialize)]
+pub struct CollaboratorPermission {
+    pub permission: String,
```
Kobzol, line 559: We could use an enum here to make it a bit nicer to deal with.

Urgau: I also wanted use an enum but GitHub's own [documentation](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28&versionId=free-pro-team%40latest&productId=rest#get-repository-permissions-for-a-user)/schema says it's just a string, but on the other hand they say it can only be `admin`, `write`, `read`, and `none`. Unsure what to do.

Kobzol: Let's leave it. If we use the collaborator permissions on more places, then we can refactor it.

PR #2215, opened 2025-10-29
Add beta backport shortcuts to t-compiler agenda

apiraino:
Adding hidden shortcuts as a convenience for us to accept backport during the meeting

r? whoever has time

thanks

### DIFF
--- a/templates/_issue.tt
+++ b/templates/_issue.tt
@@ -2,6 +2,11 @@
 {%- if backport_branch != "" %}
   - Authored by {{ issue.author }}
   - Voting [Zulip topic](#)
+<!--
+Triagebot commands for handling backports on GitHub
+If approved: @triagebot backport {{ backport_branch }} accept {{ issue.number }}
+If declined: @triagebot backport {{ backport_branch }} decline {{ issue.number }}
+-->
 {%- endif -%}
 {% if issue.mcp_details.concerns %}{%- for concern in issue.mcp_details.concerns %}
     - concern: [{{- concern.0 -}}]({{- concern.1 -}})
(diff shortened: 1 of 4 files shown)


templates/prioritization_agenda.tt
```diff
@@ -87,7 +87,7 @@ note_id: xxx
 
 ## Performance logs
 
-> [triage logs](https://github.com/rust-lang/rustc-perf/tree/master/triage#triage-logs)
+> [triage logs](https://github.com/rust-lang/rustc-perf/tree/master/triage/2025)
```
Kobzol: Can you interpolate the current year? Just to avoid having to update it in a few months :)

templates/_issue.tt
```diff
@@ -2,6 +2,11 @@
 {%- if backport_branch != "" %}
   - Authored by {{ issue.author }}
   - Voting [Zulip topic](#)
+<!--
```
Kobzol, line 5: Why is the HTML comment used here?

templates/prioritization_agenda.tt
apiraino: I did the lazy thing and enabled the "builtins" feature for `zola`, this brings in some dependendencies. If you prefer I can figure out another hand-made solution which doesn't increase the crate count

templates/_issue.tt
apiraino: it's a suggestion text for the person driving the meeting to copy and paste the message directly to approve/decline the backport. Saves some typing to Wesley :-)

templates/prioritization_agenda.tt
Kobzol: Fine by me, might be useful elsewhere.

templates/_issue.tt
Kobzol: I see, cool.

PR #2216, opened 2025-10-29
Add alias for compound labels (take 2)

apiraino:
> [!NOTE]
> This is the fixed version of https://github.com/rust-lang/triagebot/commit/154a7e764c3d9b9c405006053e67e4f6b3d75bfe

Configure relabel command aliases from the `triagebot.toml`. When a valid alias is parsed, it will be replaced with the labels configured.

Example configuration:
```
[relabel.cmd-alias]
add-labels = ["Foo", "Bar"]
rem-labels = ["Baz"]
```

The command `@rustbot label cmd-alias` translates to:
```
@rustbot label +Foo +Bar -Baz
```

The command `@rustbot label -cmd-alias` translates to:
```
@rustbot label +Baz -Foo -Bar
```

Note: self-canceling labels will be omitted.
The command `@rustbot label cmd-alias +Baz` translates to:
```
@rustbot label +Foo +Bar
```

r? @Urgau

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,5 +1,6 @@
 use crate::changelogs::ChangelogFormat;
 use crate::github::{GithubClient, Repository};
+use parser::command::relabel::{Label, LabelDelta, RelabelCommand};
 use std::collections::{HashMap, HashSet};
 use std::fmt;
 use std::sync::{Arc, LazyLock, RwLock};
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -250,10 +251,66 @@ pub(crate) struct MentionsEntryConfig {
...
+                } else {
+                    deltas.push(tk);
+                }
+            } else {
+                deltas.push(tk);
+            }
+        }
+        RelabelCommand(deltas)
+    }
+}
```
apiraino, line 278: The fix is here, I changed the logic to consider also the case where aliases are not configured.

(not a big fan of this double *else* though)

Urgau: What about removing the if entirely, like this?
```suggestion
    pub(crate) fn retrieve_command_from_alias(&self, input: RelabelCommand) -> RelabelCommand {
        let mut deltas = vec![];
        // parse all tokens: if one matches an alias, extract the labels
        // else, it will assumed to be a label
        for tk in input.0.into_iter() {
            let name = tk.label() as &str;
            if let Some(alias) = self.aliases.get(name) {
                let cmd = alias.to_command(matches!(tk, LabelDelta::Remove(_)));
                deltas.extend(cmd.0);
            } else {
                deltas.push(tk);
            }
        }
        RelabelCommand(deltas)
    }
}
```

apiraino: good point it's probably not really necessary

PR #2218, opened 2025-10-29
Add `r? me` as an alias for self-assignment

Urgau:
Suggested at [#t-compiler/contrib-private > Encouraing more compiler reviewer on rotation? @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/244344-t-compiler.2Fcontrib-private/topic/Encouraing.20more.20compiler.20reviewer.20on.20rotation.3F/near/547714257)

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -803,6 +803,13 @@ async fn find_reviewer_from_names(
         return Ok(ReviewerSelection::from_name(name.clone()));
     }
 
+    // Allow `me` as an alias for self-assign, which is always allowed.
+    if let [name] = names
+        && name == "me"
+    {
+        return Ok(ReviewerSelection::from_name(requested_by.to_string()));
+    }
+
     let candidates =
         candidate_reviewers_from_names(db, workqueue, teams, config, issue, names).await?;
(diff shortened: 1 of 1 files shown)


src/handlers/assign.rs
```diff
@@ -803,6 +803,13 @@ async fn find_reviewer_from_names(
         return Ok(ReviewerSelection::from_name(name.clone()));
     }
 
+    // Allow `me` as an alias for self-assign, which is always allowed.
+    if let [name] = names
+        && name == "me"
```
Kobzol, line 808: It's a bit disqualifying for https://github.com/me.. 😆 But until they become a Rust team member, I guess it's fine.

Kobzol (approved):

src/handlers/assign.rs
Urgau: That's unexpected. :smile:

If they ever become part of the project we could make it so `r? @me` does assign them.

fmease: That already applies to users called like adhoc groups like `compiler` (https://github.com/compiler) :D (or user `try` (https://github.com/try) in the case of bors (`@bors2 delegate=try`))

Issue #2217, opened 2025-10-29
Add command that rerolls assignment based on PR contents (`[assign.owners]`)


RalfJung:
The initial PR assignment takes into account all the data in triagebot.toml about who reviews which part of the codebase. There's no way to re-sample that set of reviewers though; if I want to reroll, I need to do e.g. `r? compiler` and just hope that I find someone who knows the right area.

Would be good to have a command that does a re-roll taking into account the `assign.owners` table.

PR #2219, opened 2025-10-29
Add `@rustbot reroll` command

Kobzol:
Currently if no reviewer can be found, we just output a generic string, the real reason is just logged. I don't *think* that this should happen often, but if it does, we could refactor `determine_assignee` to return the error instead of just logging it.

Fixes: https://github.com/rust-lang/triagebot/issues/1762
Fixes: https://github.com/rust-lang/triagebot/issues/2217

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -579,8 +579,9 @@ pub(super) async fn handle_command(
                 let pr = ctx
                     .github
                     .pull_request(issue.repository(), issue.number)
-                    .await?;
-                let Some(diff) = pr.diff(&ctx.github).await? else {
+                    .await
+                    .context("Cannot load pull request from GitHub")?;
+                let Some(diff) = pr.diff(&ctx.github).await.context("Cannot load PR diff")? else {
                     bail!(
                         "expected issue {} to be a PR, but the diff could not be determined",
                         issue.number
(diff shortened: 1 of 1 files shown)


src/handlers/assign.rs
```diff
@@ -564,6 +571,33 @@ pub(super) async fn handle_command(
...
+                };
+
+                let (assignee, _) =
+                    determine_assignee(ctx, None, issue, &event.user().login, config, diff).await?;
+                if let Some(assignee) = assignee {
+                    set_assignee(ctx, issue, &ctx.github, &assignee).await?;
+                } else {
+                    issue
+                        .post_comment(&ctx.github, "Cannot determine a new reviewer. Use `r? <username or team>` to request a specific reviewer or a team.")
+                        .await?;
```
Urgau: Any reason not to use `user_error!` or `ErrorComment`?

src/handlers/assign.rs
```diff
@@ -564,6 +571,33 @@ pub(super) async fn handle_command(
...
             }
+            AssignCommand::Reroll => {
+                // We need to compute the PR diff here, but the IssuesEvent created from a
+                // comment webhook doesn't contain the required `base` and `head` fields.
+                // So we have to load the information about the pull request from the GitHub API
+                // explicitly.
+                let pr = ctx
+                    .github
+                    .pull_request(issue.repository(), issue.number)
+                    .await?;
```
Urgau: Could you add some `.context("...")` to all the newly added `?`, so we can more easily detect where the error came from.

src/handlers/assign.rs
Kobzol: Nope, just forgot that we can use it :) Changed it.

src/handlers/assign.rs
Kobzol: Done!

Urgau (commented):
Looks good to me, but I would like to hear from @jackh726 as well.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2219/ca35b2f88e69b53640aa4c7054a3fb48616f8dab..9393844abf4d94421e89868b6d4d9e0bfcb184e6)*

Issue #2225, opened 2025-11-07
Remove the "no mentions" handler


emilyalbini:
[GitHub is removing notifications for @-mentions in commit messages on December 8th 2025.](https://github.blog/changelog/2025-11-07-removing-notifications-for-mentions-in-commit-messages/) We will then be able to remove the no-mentions handler.

Urgau:
Removed in https://github.com/rust-lang/triagebot/pull/2230 and https://github.com/rust-lang/rust-forge/pull/987. 🎉

PR #2230, opened 2025-12-08
Remove `[no-mentions]` handler as GitHub removed mentions in commits

Urgau:
cf. https://github.com/rust-lang/triagebot/issues/2225


Urgau:
Actually that's not right, I missed https://github.com/rust-lang/rustc_codegen_cranelift/blob/069016eef896febeb302a4ca95c4d7008276300f/triagebot.toml#L7

PR #2231, opened 2025-12-10
Store tokens as secret strings

Kobzol:
To avoid leaking them into logs by accident.

@apiraino found out about tokens being leaked into logs.

r? @apiraino


apiraino (approved):
I like this approach

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2231/c40db643d7e01b84de6231f1dc876d82bfc4858f..b9d5eaa468decb8e056f86b9cfe1191694f628ca)*

PR #2242, opened 2025-12-31
docs_update: Skip if a PR is already open

ehuss:
This fixes an issue where triagebot would get stuck in a loop trying to update a PR every few hours if a PR is already open for a docs update.

The solution here is to just skip the update if a pull request is already open. This uses a GraphQL search query because it is relatively cheap (costs 1 graphql point), and there wasn't anything that already did anything similar.

### DIFF
--- a/src/handlers/docs_update.rs
+++ b/src/handlers/docs_update.rs
@@ -66,7 +66,10 @@ impl Job for DocsUpdateJob {
 pub async fn docs_update() -> Result<Option<Issue>> {
     let gh = GithubClient::new_from_env();
 
-    if is_pr_already_open(&gh).await? {
+    if is_pr_already_open(&gh)
+        .await
+        .context("failed to determine if a PR is already open for updating the docs")?
+    {
         tracing::trace!("pr is already open");
         return Ok(None);
     }


src/handlers/docs_update.rs
```diff
@@ -65,6 +65,12 @@ impl Job for DocsUpdateJob {
 
 pub async fn docs_update() -> Result<Option<Issue>> {
     let gh = GithubClient::new_from_env();
+
+    if is_pr_already_open(&gh).await? {
```
Urgau: Let's add some context to the error.
```suggestion
    if is_pr_already_open(&gh).await.context("fail to determine if a PR is already open for updating the docs")? {
```

Urgau (commented):
I tested the GraphQL query with curl, works as expected.

Just a small nit.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2242/54054a99413a9e7a1535f338889ce7cc1ce2f68d..7d275ff3db60578d204fb4f300ac504015990858)*

Urgau (approved):

PR #2244, opened 2026-01-03
Enhance mentions with glob patterns matching

### LINKED ISSUE #1855: Support listing multiple files/folders for a single "mention" configuration
Sometimes one wants to group together a bunch of files/folders and have them all watched by a team. A particularly common pattern for this is `folder/name.rs`+`folder/name/`. It would be nice if triagebot support somehow configuring this in one go, instead of having to repeat the configuration for both patterns. Maybe it could even support `folder/name*`, or some special syntax to specifically cover both `name.rs` and `name/` (since that's how Rust modules work)?

Urgau:
This PR adds supports for glob patterns matching in the `[mentions]` file path matching.

We currently use a "starts with" matching, but users have expressed the desire for more expressiveness #1855.

In order to *not* break the world, we add an implicit `*` at the end to make the glob always a "starts with" pattern.

I have also switch library from `glob` to `globset` as alternates (`{foo,bar}`) are not supported by the former.

Fixes #1855
cc @tgross35

### DIFF
--- a/src/handlers/check_commits/validate_config.rs
+++ b/src/handlers/check_commits/validate_config.rs
@@ -2,7 +2,7 @@
 //! changes are a valid configuration file.
 
 use crate::{
-    config::CONFIG_FILE_NAME,
+    config::{CONFIG_FILE_NAME, MentionsEntryConfig, MentionsEntryType},
     github::FileDiff,
     handlers::{Context, IssuesEvent},
 };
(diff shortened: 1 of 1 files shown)


Urgau:
> > In order to break the world
> 
> I assume this was supposed to say "In order to _not_ break the world" :D

🤦, yes, to *not* break the world.

PR #2245, opened 2026-01-07
Fix backport label handler

apiraino:
Do not allow nominating for backport a PR that is already backport approved.
Revert the `-nominated` label that was added on GitHub.

r? @Kobzol 

thanks

### DIFF
--- a/src/handlers/backport.rs
+++ b/src/handlers/backport.rs
@@ -15,13 +15,6 @@ static CLOSES_ISSUE_REGEXP: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new("(?i)(?P<action>close[sd]*|fix([e]*[sd]*)?|resolve[sd]*)(?P<spaces>:? +)(?P<org_repo>[a-zA-Z0-9_-]*/[a-zA-Z0-9_-]*)?#(?P<issue_num>[0-9]+)").unwrap()
 });
 
-const BACKPORT_LABELS: [&str; 4] = [
-    "beta-nominated",
-    "beta-accepted",
-    "stable-nominated",
-    "stable-accepted",
-];
-
 const REGRESSION_LABELS: [&str; 3] = [
     "regression-from-stable-to-nightly",
     "regression-from-stable-to-beta",
@@ -41,7 +34,7 @@ pub(crate) struct BackportInput {
 }
 
 pub(super) async fn parse_input(
-    _ctx: &Context,
+    ctx: &Context,
     event: &IssuesEvent,
     config: Option<&BackportConfig>,
 ) -> Result<Option<BackportInput>, String> {
@@ -69,9 +62,26 @@ pub(super) async fn parse_input(
     let pr = &event.issue;
 
     let pr_labels: Vec<&str> = pr.labels.iter().map(|l| l.name.as_str()).collect();
-    if contains_any(&pr_labels, &BACKPORT_LABELS) {
-        log::debug!("PR #{} already has a backport label", pr.number);
-        return Ok(None);
+
+    if let IssuesAction::Labeled { label } = &event.action {
+        if (label.name == "beta-nominated" && contains_any(&pr_labels, &["beta-accepted"]))
+            || (label.name == "stable-nominated" && contains_any(&pr_labels, &["stable-accepted"]))
+        {
+            log::debug!(
+                "Will not nominate for backport, PR #{} is already backport accepted (found labels: {:?})",
+                pr.number,
+                pr_labels
+            );
+            let label_to_remove = vec![Label {
+                name: label.name.clone(),
+            }];
+            let _ = event
+                .issue
+                .remove_labels(&ctx.github, label_to_remove)
+                .await
+                .context("failed to remove labels from the issue");
+            return Ok(None);
+        }
     }
 
     // Retrieve backport config for this PR, based on its team label(s)
@@ -86,7 +96,7 @@ pub(super) async fn parse_input(
                 cfg.required_pr_labels.iter().map(String::as_str).collect();
             if !contains_any(&pr_labels, &required_pr_labels) {
                 log::warn!(
-                    "Skipping backport nomination: PR is missing one required label: {:?}",
+                    "Skipping backport nomination: PR is missing the required labels. Labels found: {:?}",
                     pr_labels
                 );
                 return false;


src/handlers/backport.rs
```diff
@@ -53,13 +58,17 @@ pub(super) async fn parse_input(
...
     // - is converted from draft to ready for review
     // - when the first comment is edited
+    // - when a label is added (later we check which one)
     let skip_check = !matches!(
         event.action,
-        IssuesAction::Opened | IssuesAction::Edited | IssuesAction::ReadyForReview
+        IssuesAction::Opened
+            | IssuesAction::Edited
+            | IssuesAction::ReadyForReview
+            | IssuesAction::Labeled { label: _ }
```
apiraino: @Kobzol Did you tell me a while ago about adding the `Labeled` case - or do I remember wrong?

src/handlers/backport.rs
```diff
@@ -69,9 +78,35 @@ pub(super) async fn parse_input(
     let pr = &event.issue;
 
     let pr_labels: Vec<&str> = pr.labels.iter().map(|l| l.name.as_str()).collect();
-    if contains_any(&pr_labels, &BACKPORT_LABELS) {
-        log::debug!("PR #{} already has a backport label", pr.number);
-        return Ok(None);
```
apiraino, line 74: the check I put here was wrong. with the previous version if a PR had *any* backport label, that  would block the handler.

more correctly, the handler should instead understand how labels work.

Again: unsure why it works without this fix :thinking:

src/handlers/backport.rs
Kobzol: I don't remember, to be honest :) But I think it makes sense.

src/handlers/backport.rs
Kobzol: Maybe because it didn't listen for the label event, and when the issue was opened, it didn't have the label yet?

src/handlers/backport.rs
```diff
@@ -69,9 +78,35 @@ pub(super) async fn parse_input(
...
+
+    // if adding a "-nominated" label to a PR, proceed if PR does not have it already
+    // (we don't check for "-accepted" labels)
+    let re = Regex::new("(beta|stable)-nominated").unwrap();
+    if matches!(&event.action, IssuesAction::Labeled { label } if re.is_match(&label.name)) {
```
Kobzol:
```suggestion
  if let IssuesAction::Labeled { label } = &event.action
        && BACKPORT_NOMINATED_LABELS.contains(&label.name.as_str())
```
To avoid the regex? Doesn't seem necessary. Same below with the accepted labels.

src/handlers/backport.rs
apiraino: :man_shrugging:

src/handlers/backport.rs
```diff
@@ -69,9 +78,37 @@ pub(super) async fn parse_input(
...
+            return Ok(None);
+        }
+    }
+
+    // If adding an "-accepted" label to a PR, proceed if PR does not have it already
+    // (we skip checking the "-nominated" labels, sometimes T-release straight backport accepts PRs)
+    if let IssuesAction::Labeled { label } = &event.action
+        && BACKPORT_ACCEPTED_LABELS.contains(&label.name.as_str())
+    {
+        if contains_any(&pr_labels, &BACKPORT_ACCEPTED_LABELS) {
```
Kobzol: This will have the same bug as before, right? If the PR is already `beta-accepted`, it will stop the handler here once `stable-accepted` is added.

You'll need to separate the labels on the axis of beta/stable, rather than nominated/accepted.

apiraino: Hmm ... you're right but the solution is more complex that separating on another axis. I'm afraid I'll need to map all possible cases (4 labels = 16 cases) and then see if I can reduce to a couple of *ifs*.

In any case, I want to keep it simple: if this does not work, I have another [patch](https://github.com/apiraino/triagebot/commit/35d1a53c2b4b98266ba3c21f2ba9d7431530e081) to add "negated labels" when sending zulip notification, that would solve the issue from another point of view.

I'll be back on this.

Kobzol: Can't just remove this:
```rust
if contains_any(&pr_labels, &BACKPORT_LABELS) {
    log::debug!("PR #{} already has a backport label", pr.number);
    return Ok(None);
}
```
? Seems like the bug is just caused by this.

Btw, I tested that once you receive the `Labeled { label: "foo" }` event, then `pr.labels()` already contains `foo`.

apiraino: sorry, I am confused by your comment: isn't that chunk of code removed in this patch? Or did you mean to just remove that code *without* adding anything else?

In the latter case, I think that not checking labels would open up to confusion when adding regression labels (which was the point of this patch).

Kobzol: Yeah, I meant remove without adding anything else. Say that we remove the original check - what will be the problematic situations that could be caused in that case? Adding `-nominated` once there is already `-accepted`? Or something else? I don't know what is a "regression label", it wasn't mentioned in this PR so far :sweat_smile:

src/handlers/backport.rs
```diff
@@ -69,9 +75,28 @@ pub(super) async fn parse_input(
     let pr = &event.issue;
 
     let pr_labels: Vec<&str> = pr.labels.iter().map(|l| l.name.as_str()).collect();
-    if contains_any(&pr_labels, &BACKPORT_LABELS) {
-        log::debug!("PR #{} already has a backport label", pr.number);
-        return Ok(None);
+
+    // If a "-nominated" label is added to an already "-accepted" PR, remove the label added by github
+    if let IssuesAction::Labeled { label } = &event.action
```
Kobzol: Just to make sure that it is intended, this logic means that if the PR was e.g. beta accepted previously, it then cannot be also stable nominated.

apiraino: OK @Kobzol I give up in trying to make this check "smart", I thought I will just hardcode the only two cases I was actually interested in preventing.

wdty?

src/handlers/backport.rs
```diff
@@ -69,9 +62,27 @@ pub(super) async fn parse_input(
...
+            );
+            let label_to_remove = vec![Label {
+                name: label.name.clone(),
+            }];
+            let _ = &event
+                .issue
+                .remove_labels(&ctx.github, label_to_remove)
+                .await
+                .context("failed to remove labels from the issue")
+                .unwrap();
```
apiraino: Removing a label after it was added using the GH UI will produce a strange "add label / remove label" events one after the other. This will be solved after we have negated labels rule (as mentioned elsewhere here). That will prevent triggering this handler.

~~(removing the `.unwrap()`)~~

src/handlers/backport.rs
```diff
@@ -86,7 +97,7 @@ pub(super) async fn parse_input(
                 cfg.required_pr_labels.iter().map(String::as_str).collect();
             if !contains_any(&pr_labels, &required_pr_labels) {
```
apiraino, line 97: I discovered that this check (`contains_any` on `required_pr_labels`) has a little bug: If the triagebot configuration is something like:
```
required-pr-labels = ["t-libs", "t-libs-api"]
```
this handler will run twice. This is fixable in different ways (in another patch), I just wanted to leave a note about it.

Kobzol (approved):
I actually wanted to propose this, I think it's the best solution for now 😆

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2245/6be41b94b5a946fc9fc78faa9ca7e390b57d1f12..19f869bb941bc9dc80389b6515eeac5f261fc576)*

PR #2248, opened 2026-01-09
Fix `[mentions]` glob matching to not match partial directories

Urgau:
This PR fixes our `[mentions]` glob matching to not match on partial directories.

This is particularly relevant if an entry is `src/tools/cargo` and the modified paths contains `src/tools/cargotest` which shouldn't be considered to be matching (as per `Path::starts_with` rules).

Reported at [#triagebot > Mentions glob matching](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Mentions.20glob.20matching/with/567093226)

cc @jieyouxu @tgross35

### DIFF
--- a/src/handlers/mentions.rs
+++ b/src/handlers/mentions.rs
@@ -430,6 +430,36 @@ mod tests {
     }
 
     #[test]
+    fn entry_submodule_and_normal_dir_modified() {
+        assert_eq!(
+            modified_paths_matches(
+                &[
+                    Path::new("src/tools/cargo"),
+                    Path::new("src/tools/cargotest")
+                ],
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Tricky.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2248/219a76b53bdb6f06711dc3714aab16b8e00d1413..f5fa24ab9932cce093d35472bc5d3136bd566977)*

Kobzol:
Maybe a test for a case where someone wanted to actually match both `cargo` and `cargotest` (a non-path-like prefix?).

PR #2250, opened 2026-01-12
Don't mention the author in CC

JonathanBrouwer:
There is an existing check but this only works if only one person is in the CC list
Got annoyed by getting pings for my own PRs, so decided to do something about it :P

Code is completely untested, not sure if there is a good way to test it?

### DIFF
--- a/src/handlers/mentions.rs
+++ b/src/handlers/mentions.rs
@@ -17,7 +17,13 @@ use std::{fmt::Write, path::PathBuf};
 const MENTIONS_KEY: &str = "mentions";
 
 pub(super) struct MentionsInput {
-    to_mention: Vec<(String, Vec<PathBuf>)>,
+    to_mention: Vec<ToMention>,
+}
+
+struct ToMention {
+    entry: String,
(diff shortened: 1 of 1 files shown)


src/handlers/mentions.rs
```diff
@@ -85,13 +91,19 @@ pub(super) async fn parse_input(
...
-            let pings_non_author = match &cc[..] {
-                [only_cc] => only_cc.trim_start_matches('@') != event.issue.user.login,
-                _ => true,
-            };
-            if !relevant_file_paths.is_empty() && pings_non_author {
-                Some((entry.to_string(), relevant_file_paths))
+            // Filter author from the cc list
+            let relevant_ccs = cc
+                .iter()
+                .filter(|cc| cc.trim_start_matches('@') != event.issue.user.login)
```
Kobzol: Let's also lowercase both. It happened to me recently a few times that even the single author check didn't work, because `triagebot.toml` lowercases my username (`kobzol`), but from the GitHub API it's actually `Kobzol`.

JonathanBrouwer: Done :)

PR #2251, opened 2026-01-13
Add GitHub issue comments viewer for long issue

Urgau:
This PR adds a GitHub issue comments viewer.

It's primarily designed for long/very long issue (not PR) where GitHub is really unhelpful with it's "Load More".

It can be accessed at `/gh-comments/{owner}/{repo}/{issue}`.

Technically it's using a GraphQl query (up to 100 comments at a time) to get the issue details and the comments (including the already HTML-ed markdown body). As far as I could see that query only cost **1** per call.

Regarding the styling of comment bodies, this PR uses the [github-markdown-css](https://github.com/sindresorhus/github-markdown-css) project (under MIT license).

For a big issue like [#99301](https:://www.github.com/rust-lang/rust/issues/99301), with 150 comments, the loading time is ~3s on my machine.

<details>
<summary>
Example of a not very long issue (but with many states)
</summary>

image

</details>

### DIFF
--- a/src/gh_comments.rs
+++ b/src/gh_comments.rs
@@ -1,3 +1,4 @@
+use std::collections::VecDeque;
 use std::fmt::Write;
 use std::sync::Arc;
 use std::time::Instant;
@@ -14,6 +15,7 @@ use hyper::{
     header::{CACHE_CONTROL, CONTENT_SECURITY_POLICY, CONTENT_TYPE},
 };
 
+use crate::github::{GitHubGraphQlComment, GitHubIssueWithComments};
 use crate::{
     errors::AppError,
     github::GitHubSimplifiedAuthor,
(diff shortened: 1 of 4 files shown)


Urgau:
> A simpler thing to avoid us being spammed might be to have a rate limiter with back-off period per IP

I've implemented the rate limiter in #2252.

SpriteOvO:
Would it also be possible to display reactions?

src/github.rs
```diff
@@ -2914,6 +2914,144 @@ impl GithubClient {
...
+    ) -> anyhow::Result<GitHubIssueWithComments> {
+        let mut cursor: Option<String> = None;
+        let mut all_comments = Vec::new();
+        let mut issue_json;
+
+        loop {
+            let mut data = self
+                .graphql_query(
+                    "
+query ($owner: String!, $repo: String!, $issueNumber: Int!, $cursor: String) {
```
Kobzol, line 2932: Just out of curiosity, any reason why use GraphQL and not https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#list-issue-comments? Is there some information not provided by the REST API? I vaguely remember some cases where the same thing being fetched by the REST API was several times faster than through the GraphQL. Would be nice to check the perf. difference on some big PR.

src/gh_comments.rs
```diff
@@ -0,0 +1,173 @@
...
+              <a href="https://github.com/{author_login}" target="_blank">{author_login}</a>
+              <span>commented on Jan 11, 2026</span>
+            </div>
+          </div>
+
+          <a href="{comment_url}" target="_blank" class="github-link">View on GitHub</a>
+        </div>
+
+        <div class="comment-body markdown-body">
+          {body_html}
```
Kobzol, line 337: I was worried about XSS here, but it seems like GitHub already pre-escapes the HTML body of the comment. I hope that we can trust that...

src/github.rs
apiraino: As a datapoint I stopped adding gql github APIs in triagebot because they're too flaky, [the usage reporting](https://docs.github.com/en/graphql/overview/rate-limits-and-query-limits-for-the-graphql-api?versionId=free-pro-team%40latest&productId=graphql) is not very reliable and I don't see GH improving them.

We already have high-level functions to retrieve comments in triagebot, don't we?

Urgau:
> 🤔 Does it only work for issues and not PRs?

Unfortunately it only works for issues for now. Supporting PRs requires more handling, in particular regarding review comments, and code highlighting; which I didn't wanted to handle in this first version.

Urgau:
> Would it also be possible to display reactions?

I've found that it most cases they are not that useful, but if people want to see them, I can add them in a follow-up PR.

src/gh_comments.rs
Urgau: Yeah, not only does GitHub do the escaping, they also gives us the rendered markdown as HTML, greatly reducing the complexity here.

src/github.rs
Urgau: Unfortunately neither the minimized state nor the minimized reason are present in the REST API, forcing us to use the GraphQl API.

Another reason is that it's not possible to retrieve the issue body/title and comments in the same request, we would have to do two requests minimum each time, while we only have to do one GraphQl request.

The GraphQl request is also consistently costing only 1 (checked with `rateLimit { cost }`).

Urgau:
I've added a small 35Mb cache with auto pruning as soon as we get an webhook event from that issue.

src/gh_comments.rs
```diff
@@ -0,0 +1,346 @@
...
+        } else if self.capacity + value.estimated_size as u64 >= MAX_CACHE_CAPACITY_BYTES {
+            self.entries.pop_back()
+        } else {
+            None
+        };
+        if let Some(removed) = removed {
+            self.capacity -= removed.1.estimated_size as u64;
+        }
+
+        // Add entry the front of the list ane return it
```
Kobzol:
```suggestion
        // Add entry the front of the list and return it
```

PR #2252, opened 2026-01-13
Add rate limiter to our most exposed and costly endpoints

Urgau:
This PR adds rate limiter to our most exposed and costly endpoints.

I used the [tower-governor](https://github.com/benwis/tower-governor) crate with a burst size of 3 requests per IP and a replenishment of one element every 15 seconds.

I don't know if those parameters are good, but we can adjust them later if we find out that they are too light or too tight.

Those limits are NOT applied to the `/github-hook` and `/zulip-hook` endpoints.

Related to https://github.com/rust-lang/triagebot/pull/2251#issuecomment-3746370870
cc @Mark-Simulacrum

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -15,6 +15,9 @@ use tokio::{task, time};
 use tower::ServiceBuilder;
 use tower::buffer::BufferLayer;
 use tower::limit::RateLimitLayer;
+use tower_governor::GovernorLayer;
+use tower_governor::governor::GovernorConfigBuilder;
+use tower_governor::key_extractor::SmartIpKeyExtractor;
 use tower_http::catch_panic::CatchPanicLayer;
 use tower_http::compression::CompressionLayer;
 use tower_http::request_id::{MakeRequestUuid, PropagateRequestIdLayer, SetRequestIdLayer};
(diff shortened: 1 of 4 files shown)


Kobzol (commented):
Makes sense. Also unsure of the numbers, I think that we could go higher, but let's see how this works. Can we avoid using the governor on localhost though? It would get annoying for developing triagebot :) Maybe by not applying it if `debug_assertions` is set, or hide it behind a CLI parameter? Or filtering out requests from localhost? (although not sure if the requests on the deployment through Docker aren't also from localhost, we'd have to test that I guess).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2252/82f9d735e202a15249db86d76df2f3febd615fdf..8fd61fdecfe681551abe4344608c58a01e186daf)*

src/main.rs
```diff
@@ -102,6 +105,16 @@ async fn run_server(addr: SocketAddr) -> anyhow::Result<()> {
         spawn_job_runner(ctx.clone());
     }
 
+    // Allow bursts with up to 3 requests per IP address
+    // and replenishes one element every 15 seconds
+    let config = GovernorConfigBuilder::default()
+        .per_second(15)
+        .burst_size(3)
```
apiraino: Who (or what) hits these `/gha-logs/*` endpoints? Do we have stats about their usage?

3 req/s with 15s cooldown look a bit restrictive - but it really depends on the answer to the above questions.

Urgau: The `/gha-logs` endpoint is the `(plain enhanced)` link when CI fails[^1].

It's used to display the raw-enhanced logs from GitHub Actions. It can take ~3/4s to fetch and display our logs.

I'm not expecting anyone to try open 4 of those links in <15s. So I think a 15s cool down is good enough.

[^1]: https://github.com/rust-lang/rust/pull/151123#issuecomment-3750603675

Kobzol:
Ok, fair enough for the rebuild, it still seems a bit annoying and unexpected if someone would work on it locally though. Maybe just disabling it in debug mode would do the trick, to avoid complicating things?

Urgau:
Disabling it when `debug_assertions` is set seems kind-of dangerous to me.

I prefer to follow our existing convention of using enthronement variables, so I added `DISABLE_RATE_LIMIT` with a warning on usage.

Kobzol (approved):
Fair enough, gating it on compilation options could likely be confusing, you're right.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2252/82f9d735e202a15249db86d76df2f3febd615fdf..b97d1db8959a0ab7b16ec6650beba361f2ce7232)*

PR #2255, opened 2026-01-18
Add reactions to our GitHub comments viewer

Urgau:
This PR adds reactions to our GitHub comments viewer, as asked in https://github.com/rust-lang/triagebot/pull/2251#issuecomment-3747954392

image

cc @SpriteOvO

### DIFF
--- a/src/gh_comments/style.css
+++ b/src/gh_comments/style.css
@@ -1,3 +1,11 @@
+/* SPDX-License-Identifier: MIT */
+
+/*
+ * Style for the /gh-comments endpoints in Triagebot.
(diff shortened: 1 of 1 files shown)


src/gh_comments/style.css
```diff
@@ -210,6 +210,31 @@ details:not([open]) > .review-thread-header {
     fill: currentColor;
 }
 
+/* === Reactions === */
```
Kobzol, line 221: Might be nice to add a comment on top of this file that says that the CSS version filepath should be bumped after every chang to this file.

Issue #2257, opened 2026-01-21
gha-logs: copied text is indented (sometimes excemtping the first line)


RalfJung:
To reproduce, I went to https://triage.rust-lang.org/gha-logs/rust-lang/rust/60798934705 and copied the linkcheck error. pasting the results looks like this:
```
	core/marker/trait.UnsafeUnpin.html:124: broken link fragment `#associatedtype.NonZeroInner` pointing to `core/num/trait.ZeroablePrimitive.html`
	core/num/struct.NonZero.html:3880: broken link fragment `#associatedtype.NonZeroInner` pointing to `core/num/trait.ZeroablePrimitive.html`
	std/marker/trait.UnsafeUnpin.html:195: broken link fragment `#associatedtype.NonZeroInner` pointing to `std/num/trait.ZeroablePrimitive.html`
	std/num/struct.NonZero.html:4108: broken link fragment `#associatedtype.NonZeroInner` pointing to `std/num/trait.ZeroablePrimitive.html`
```
If I am slightly less careful in how I select text, and only select the first character of the first line but don't drag the mouse further up towards the next line, I get:
```
core/marker/trait.UnsafeUnpin.html:124: broken link fragment `#associatedtype.NonZeroInner` pointing to `core/num/trait.ZeroablePrimitive.html`
	core/num/struct.NonZero.html:3880: broken link fragment `#associatedtype.NonZeroInner` pointing to `core/num/trait.ZeroablePrimitive.html`
	std/marker/trait.UnsafeUnpin.html:195: broken link fragment `#associatedtype.NonZeroInner` pointing to `std/num/trait.ZeroablePrimitive.html`
	std/num/struct.NonZero.html:4108: broken link fragment `#associatedtype.NonZeroInner` pointing to `std/num/trait.ZeroablePrimitive.html`
```
It seems like there is invisible indentation in front of each line, and for the first line it gets copied or not depending on how exactly one selects the text (there is no difference visible in the selection).

This is with Firefox.

PR #2259, opened 2026-01-24
Strip added tabulations when copying logs lines

Urgau:
This is a targeted fix/hack to prevent Firefox (and others) from automatically adding tabulations when coping logs lines.

Fixes https://github.com/rust-lang/triagebot/issues/2257

### DIFF
--- a/src/gha_logs/gha_logs.js
+++ b/src/gha_logs/gha_logs.js
@@ -114,8 +114,17 @@ if (location.hash !== "") {
 
 // 9. Add a copy handler that force plain/text copy
 logsEl.addEventListener("copy", function(e) {
-    var text = window.getSelection().toString();
-    e.clipboardData.setData('text/plain', text);
+    // Get the browser selection as string
+    const text = window.getSelection().toString();
+
+    // Let's strip the first tabulation on each line as
+    // WebBrowsers tend to add tabulations to the copy between
+    // each column in a table.
+    //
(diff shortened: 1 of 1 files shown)


src/gha_logs/gha_logs.js
```diff
@@ -114,8 +114,9 @@ if (location.hash !== "") {
 
 // 9. Add a copy handler that force plain/text copy
 logsEl.addEventListener("copy", function(e) {
-    var text = window.getSelection().toString();
-    e.clipboardData.setData('text/plain', text);
+    const text = window.getSelection().toString();
+    const stripped = text.replace(/^\t/gm, "");
```
Kobzol, line 125: Could you maybe just add a comment here to explain what this does and why we need it? Thanks.

Urgau: Sure, done.

src/gha_logs/gha_logs.js
```diff
@@ -114,8 +114,17 @@ if (location.hash !== "") {
 
 // 9. Add a copy handler that force plain/text copy
 logsEl.addEventListener("copy", function(e) {
-    var text = window.getSelection().toString();
-    e.clipboardData.setData('text/plain', text);
+    // Get the browser selection as string
+    const text = window.getSelection().toString();
+
+    // Let's stripped the first tabulation on each lines as
```
Kobzol:
```suggestion
    // Let's strip the first tabulation on each line as
```

PR #2260, opened 2026-01-26
repo-aware team-stats queries

davidtwco:
`team-stats` isn't especially useful as it doesn't consider whether any particular reviewer is in the `adhoc-groups` for a team for a given repository. This means that in practice, most of a team can be considered on rotation even when the majority aren't actually on rotation.

Adds a `team-stats <team> <repo>` variant of the `team-stats` Zulip command that considers the `adhoc-groups` of the relevant team in the repository requested.

I haven't tested this because it's tricky to receive Zulip webhooks on my work machine.

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -268,7 +268,9 @@ async fn handle_command<'a>(
                 ping_goals_cmd(ctx.clone(), gh_id, message_data, args).await
             }
             ChatCommand::DocsUpdate => trigger_docs_update(message_data, &ctx.zulip),
-            ChatCommand::TeamStats { name } => team_status_cmd(&ctx, name).await,
+            ChatCommand::TeamStats { name, repo } => {
+                team_status_cmd(&ctx, name, repo.as_deref()).await
+            }
         };
 
         let output = output?;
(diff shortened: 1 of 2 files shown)


Urgau:
I was more looking at it not crashing, but yes, all the members off rotation doesn't seems right.

src/zulip.rs
```diff
@@ -511,25 +548,42 @@ async fn team_status_cmd(ctx: &Context, team_name: &str) -> anyhow::Result<Optio
...
     let (mut on_rotation, mut off_rotation): (Vec<&TeamMember>, Vec<&TeamMember>) =
         members.iter().partition(|member| {
             let rotation_mode = review_prefs
                 .get(member.github.as_str())
                 .map(|prefs| prefs.rotation_mode)
                 .unwrap_or_default();
-            matches!(rotation_mode, RotationMode::OnRotation)
+            let in_review_queue_for_repository = reviewers_for_repository
+                .as_ref()
+                .map(|reviewers| reviewers.contains(&member.github))
```
Urgau, line 573: I think this part is not right. It's comparing `adhoc_group` reviewers:
https://github.com/rust-lang/rust/blob/55407b8cdb1457c62e0c852f5b53a9cf63ec4e1b/triagebot.toml#L1452-L1454

with team names:
https://github.com/rust-lang/team/blob/c2df2caa271926fdd5d8d1c6632167a331292e89/people/Aaron1011.toml#L3

but one starts with a `@` while the other one doesn't, creating a mismatch.

PR #2264, opened 2026-02-01
Handle missing author in gh-comments by using the ghost account

Urgau:
This PR change our de-serialization of the author field for the `/gh-comments` endpoint as GitHub GraphQl api can return a `null` author when the account/user no longer exist.

In that case we should just do what GitHub does and link it to the ghost account.

Fun fact: I discovered this bug while at FOSDEM, I was looking at some of our most commented issues and rust-lang/rust#32838 was not loading at all due to having deleted accounts.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -3187,8 +3187,7 @@ pub struct GitHubIssueWithComments {
     pub body_html: String,
     pub state: GitHubIssueState,
     pub url: String,
-    #[serde(deserialize_with = "deserialize_author")]
-    pub author: GitHubSimplifiedAuthor,
+    pub author: Option<GitHubSimplifiedAuthor>,
     #[serde(rename = "createdAt")]
     pub created_at: chrono::DateTime<chrono::Utc>,
     #[serde(rename = "updatedAt")]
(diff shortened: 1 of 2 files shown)


src/github.rs
```diff
@@ -3187,6 +3187,7 @@ pub struct GitHubIssueWithComments {
     pub body_html: String,
     pub state: GitHubIssueState,
     pub url: String,
+    #[serde(deserialize_with = "deserialize_author")]
```
Kobzol: Isn't this just:
```suggestion
    #[serde(default)]
```
?

Urgau: `#[serde(default)]` only use the default value when it's not provided but the response contains `null` which can't de-serialized into `GitHubSimplifiedAuthor`.

Kobzol: Right, I see. Well, in that case I'd personally make that explicit (`Option<GitHubSimplifiedAuthor>`), and do the `.unwrap_or_default()` call in other code, rather than in deserialization.

But I'm fine with both, feel free to keep the current version.

Urgau: I was thinking the same thing but it seemed more annoying than just hacking the de-serialization.

I push a variant of what you wanted, unsure if it's better, I will let you be the judge of it. ;-)

PR #2266, opened 2026-02-03
Add `pr_merged` option in `[autolabel]` to label PRs after merged

Urgau:
As requested in https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/On-merge.20label.3F, this PR adds a new option for `[autolabel]`: `pr_merged` label PRs after merged.

```toml
[autolabel."needs-relnotes-triage"]
pr_merged = true
```

*Note that I haven't tested it but I don't see why it wouldn't work.*

cc @jieyouxu @ytmimi

### DIFF
--- a/src/handlers/autolabel.rs
+++ b/src/handlers/autolabel.rs
@@ -100,6 +100,7 @@ pub(super) async fn parse_input(
 
                 let is_opened =
                     matches!(event.action, IssuesAction::Opened | IssuesAction::Reopened);
+                let is_closed = matches!(event.action, IssuesAction::Closed);
 
                 // Treat the following situations as a "new PR":
                 // 1) PRs that were (re)opened and are not draft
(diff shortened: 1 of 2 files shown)


src/handlers/autolabel.rs
```diff
@@ -145,6 +154,10 @@ pub(super) async fn parse_input(
                     to_remove.push(Label {
                         name: label.to_owned(),
                     });
+                } else if cfg.new_merged && matches!(event.action, IssuesAction::Reopened) {
```
Kobzol: I don't think that a PR can be unmerged :thinking:. In any case, I think we don't need this.

Urgau: For some reason I though it was possible to re-open a merged PR, but apparently it's only possible for closed PR.

PR #2268, opened 2026-02-04
Add basic explanation for reviewer selection

Kobzol:
The review selection algorithm is a pretty much a black box. This aims to provide some basic explanation to let reviewers know why they were selected.

It only adds the explanation when a PR is opened (without explicit `r?`), because at that point we actually send some comment. After `@rustbot reroll` or `r?` we don't send a comment, so adding it would cause new spam.

Maybe we could add some `@rustbot review-explain` command or something for these cases? But then we'd have to remember the explanation in the DB, I suppose. Or add `r? foo --explain` :)

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -246,8 +246,13 @@ pub(super) async fn handle_input(
                         SelectionStep::RandomlySelectedFrom(candidates) => {
                             format!("Random selection from {}", format_candidates(&candidates))
                         }
-                        SelectionStep::Fallback => "Fallback group".to_string(),
-                        SelectionStep::FileDiff => "Files modified in the PR".to_string(),
+                        SelectionStep::Fallback(group) => {
+                            format!("Fallback group: {}", format_candidates(&group))
+                        }
+                        SelectionStep::FileDiff(candidates) => format!(
+                            "People who recently interacted with files modified in this PR: {}",
+                            format_candidates(&candidates)
+                        ),
                         SelectionStep::Expansion { from, to } => {
                             format!(
                                 "{} expanded to {}",
@@ -432,7 +437,7 @@ async fn determine_assignee(
             {
                 Ok(assignee) => {
                     return Ok((
-                        Some(assignee.prepend_selection_step(SelectionStep::FileDiff)),
+                        Some(assignee.prepend_selection_step(SelectionStep::FileDiff(candidates))),
                         false,
                     ));
                 }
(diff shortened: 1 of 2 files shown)


src/handlers/assign.rs
```diff
@@ -813,22 +879,58 @@ Please select a different reviewer.",
...
+enum SelectionStep {
+    /// The user assigned themselves.
+    SelfAssign,
+    /// Random selection amongst N candidates.
+    RandomlySelectedFrom(Vec<String>),
+    /// The reviewer was selected as a last resort from the adhoc fallback group.
+    Fallback,
+    /// The reviewer was selected based on the PR diff.
+    FileDiff,
+    ///
```
ehuss: The enums are pretty obvious, but this empty doc string seemed a little strange.

Kobzol: Oops, that was not intended. Thanks!

src/handlers/assign/tests/tests_candidates.rs
```diff
@@ -192,6 +196,7 @@ async fn at_max_capacity() {
                     suppressed_error: Some(FindReviewerError::ReviewerAtMaxCapacity {
                         username: "martin".to_string(),
                     }),
+                    selection_steps: vec![],
```
Urgau, line 209: Is this possible to have a test where the `selection_steps` are fulfilled.

src/handlers/assign.rs
```diff
@@ -890,11 +994,21 @@ async fn find_reviewer_from_names(
         candidates
     );
 
-    // Select a random reviewer from the filtered list
-    Ok(candidates
-        .into_iter()
-        .choose(&mut rand::thread_rng())
-        .expect("candidate_reviewers_from_names should return at least one entry"))
+    if candidates.len() == 1 {
+        Ok(candidates.into_iter().next().unwrap())
```
Urgau, line 1005:
```suggestion
    if candidates.len() == 1 {
        Ok(candidate.remove(0))
```

Kobzol: Not sure what you meant by this, `candidates` is a `HashSet` :thinking:

src/handlers/assign/tests/tests_candidates.rs
Kobzol: I was wondering why none of the tests were failing - I forgot to set the expansion selection steps in the returned reviewer :laughing:

Kobzol: I'm afraid that's the only thing we can easily test today though, the rest of the selection steps are in code paths that are not currently tested, as they are not very easy to mock.

Kobzol:
Updated tests.

src/handlers/assign.rs
Urgau: Nevermind, I thought it was a `Vec`.

PR #2269, opened 2026-02-04
Report banned users to moderator stream on Zulip

Kobzol:
Context: https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Feature.20request.3A.20automatic.20topic.20post.20on.20ban/with/571946320

The triagebot account will need to be added to this channel for this to work.

### DIFF
--- a/src/github.rs
+++ b/src/github.rs
@@ -3202,6 +3202,125 @@ query ($owner: String!, $repo: String!, $issueNumber: Int!, $commentsCursor: Str
 
         serde_json::from_value(final_issue).context("fail to deserialize final response")
     }
+
+    /// Fetches recent comments made by a user in a GitHub organization.
+    ///
+    /// Returns up to `limit` comments, sorted by creation date (most recent first).
+    /// Each comment includes the URL, body snippet, and the issue/PR title it was made on.
+    pub async fn user_comments_in_org(
+        &self,
+        username: &str,
+        org: &str,
+        limit: usize,
+    ) -> anyhow::Result<Vec<UserComment>> {
+        // GitHub's GraphQL API doesn't seem to support filtering comments by author directly.
+        // We can use two endpoints here - either use the search query and filter by commented and
+        // organization, or use the user query and access its issueComments connection.
+        // The user endpoint would be more efficient, in theory. However, if the user makes a lot of
+        // comments in different organizations, we might load a lot of data before we get to their
+        // comments in the given organization. So instead we use the search endpoint.
+
+        // The endpoint loads issues (and PRs), not comments.
+        // The `commenter:` filter guarantees each returned issue has at least one comment
+        // from the user. So we fetch `limit` issues and a small number of recent comments
+        // per issue, then filter to only the user's comments.
+        let search_query = format!("commenter:{username} org:{org} sort:updated-desc");
+        let issues_to_fetch = limit;
+        let comments_per_issue = 100;
+
+        let data = self
+            .graphql_query(
+                r#"
+query($query: String!, $issueLimit: Int!, $commentLimit: Int!) {
+  search(query: $query, type: ISSUE, first: $issueLimit) {
+    nodes {
+      ... on Issue {
+        url
+        title
+        comments(first: $commentLimit, orderBy: {field: UPDATED_AT, direction: DESC}) {
+          nodes {
+            author { login }
+            body
+            url
+            createdAt
+          }
+        }
+      }
+      ... on PullRequest {
+        url
+        title
+        comments(first: $commentLimit, orderBy: {field: UPDATED_AT, direction: DESC}) {
+          nodes {
+            author { login }
+            body
+            url
+            createdAt
+          }
+        }
+      }
+    }
+  }
+}
+                "#,
+                serde_json::json!({
+                    "query": search_query,
+                    "issueLimit": issues_to_fetch,
+                    "commentLimit": comments_per_issue,
+                }),
+            )
+            .await
+            .context("failed to search for user comments")?;
+
+        let mut all_comments: Vec<UserComment> = Vec::new();
+
+        if let Some(nodes) = data["data"]["search"]["nodes"].as_array() {
+            for node in nodes {
+                let issue_title = node["title"].as_str().unwrap_or("Unknown");
+                let issue_url = node["url"].as_str().unwrap_or("");
+
+                if let Some(comments) = node["comments"]["nodes"].as_array() {
+                    for comment in comments {
+                        // Filter to only comments by the target user
+                        let author = comment["author"]["login"].as_str().unwrap_or("");
+                        if !author.eq_ignore_ascii_case(username) {
+                            continue;
+                        }
+
+                        let body = comment["body"].as_str().unwrap_or("");
+                        let url = comment["url"].as_str().unwrap_or("");
+                        let created_at = comment["createdAt"]
+                            .as_str()
+                            .and_then(|s| chrono::DateTime::parse_from_rfc3339(s).ok())
+                            .map(|dt| dt.with_timezone(&chrono::Utc));
+
+                        all_comments.push(UserComment {
+                            issue_title: issue_title.to_string(),
+                            issue_url: issue_url.to_string(),
+                            comment_url: url.to_string(),
+                            body: body.to_string(),
+                            created_at,
+                        });
+                    }
+                }
+            }
+        }
+
+        // Sort by creation date (most recent first) and take the limit
+        all_comments.sort_by(|a, b| b.created_at.cmp(&a.created_at));
+        all_comments.truncate(limit);
+
+        Ok(all_comments)
+    }
+}
+
+/// A comment made by a user on an issue or PR.
+#[derive(Debug, Clone)]
+pub struct UserComment {
+    pub issue_title: String,
+    pub issue_url: String,
+    pub comment_url: String,
+    pub body: String,
+    pub created_at: Option<DateTime<Utc>>,
 }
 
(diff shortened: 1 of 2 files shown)


src/handlers/report_user_bans.rs
```diff
@@ -0,0 +1,114 @@
+//! Handler that reports GitHub bans and unbans to a Zulip channel.
+
+use crate::github::{OrgBlockAction, OrgBlockEvent, UserComment};
+use crate::handlers::Context;
+use crate::zulip::MessageApiRequest;
+use crate::zulip::api::Recipient;
+use tracing as log;
+
+// The #mods channel on Zulip
+const ZULIP_STREAM_ID: u64 = 464799;
```
Urgau, line 10: I would be better if this was configured somewhere, same with the one below, but they are less important.

Maybe r-l/team could store the stream ID.

We would really benefit from having an organization-wide config.

--------------

Looking at Zulip docs, we may not need the ID at all, we can apparently just use the stream name (https://zulip.com/api/send-message#parameter-to), which we could get from the team database.

src/github/webhook.rs
```diff
@@ -251,6 +259,20 @@ async fn process_payload(
 
             Event::Create(payload)
         }
+        EventName::OrgBlock => {
```
Urgau, line 262: Receiving those events requires "at least read-level access for the "Administration" organization permission". Let's make sure we add those permissions.

cf. [#t-infra > Triagebot org permissions](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Triagebot.20org.20permissions/with/571974309)

src/github.rs
```diff
@@ -3176,6 +3202,125 @@ query ($owner: String!, $repo: String!, $issueNumber: Int!, $commentsCursor: Str
...
+        // The endpoint loads issues (and PRs), not comments.
+        // The `commenter:` filter guarantees each returned issue has at least one comment
+        // from the user. So we fetch `limit` issues and a small number of recent comments
+        // per issue, then filter to only the user's comments.
+        let search_query = format!("commenter:{username} org:{org} sort:updated-desc");
+        let issues_to_fetch = limit;
+        let comments_per_issue = 100;
+
+        let data = self
+            .graphql_query(
```
Urgau, line 3232: Instead of a GraphQl query I think we could do it with the REST `events` endpoint: [List public events for a user](https://docs.github.com/en/rest/activity/events#list-public-events-for-a-user), and filter by `IssueCommentEvent`, or maybe just print a summary of all their actions?

src/handlers/report_user_bans.rs
```diff
@@ -0,0 +1,114 @@
...
+        id: ZULIP_STREAM_ID,
+        topic: &topic,
+    };
+
+    let req = MessageApiRequest {
+        recipient,
+        content: &message,
+    }
+    .send(&ctx.zulip)
+    .await;
```
apiraino, line 73: I am unclear if this Zulip API ([docs](https://zulip.com/api/send-message#parameter-topic)) just creates a topic or appends a comment to an existing topic with the same name.

Urgau: I think in Zulip world, topic names are just metadata attached to the message, so there isn't really a creation or append to a topic. Just sending a message with a topic name is enough to "create/update it".

Kobzol: Yeah, it's essentially an upsert - creates if it doesn't exist or adds a new message if it already did.

src/github/webhook.rs
Kobzol: Seems like we should have it.

src/handlers/report_user_bans.rs
Kobzol: Hmm, we don't have this Zulip stream synchronized in `team` today. In a way I'd rather hardcode it, to ensure that the messages won't be sent anywhere else by accident, the ID should make sure of that.

src/github.rs
Kobzol: image

This note doesn't instill much confidence in me :laughing: When someone is banned, it's often for their recent interactions, I guess, would be nice to show those. Documentation for this event is also pretty terrible, I couldn't even find the data about the comment in it, and I scrolled for 3 minutes.. :laughing:

src/handlers/report_user_bans.rs
Urgau: Thinking more about this, I think we could have in the team database a triagebot configuration for each organization, and/or even for the organization we managed. We could store Zulip stream IDs, the team database url, the main repo name, ... without having to hard code it.

But that's a much larger change, should be discussed separately.

One this I would still prefer to have, is to filter the organization where the ban happend based on the one registered in the team database, since as we discovered we get webhooks [from other organizations](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Triagebot.20GitHub.20App.20migration/near/572170728).

src/github.rs
Urgau: > Documentation for this event is also pretty terrible, I couldn't even find the data about the comment in it

It's under `payload.comment.body`. ;-)

That endpoints seems to basically return the same thing as a webhook input.

> This note doesn't instill much confidence in me 😆

That's fair, then let's not use it.

src/handlers/report_user_bans.rs
Kobzol: Hmm, I mean, worst case we just get a notification that someone was banned in `rust-embedded`? It's still under `team`, so in theory our moderation actions also reach there.

I hardcoded the stream because it is special; we need to manually add the triagebot account to it for this to work. The whole logic is only for the mods, I don't think we'd use it anywhere else.

apiraino: I agree that we probably don't need to overengineer this at this time. Besides this case - don't we already use the `triagebot.toml` as the "source of truth" for Zulip IDs?

Kobzol: I'm not aware of any usage of Zulip IDs in `triagebot.toml` specifically. I think it's not worth automating this, because we need to add triagebot to the channel manually anyway, so even if the config would change the channel, we need to intervent manually.

Urgau: > Hmm, I mean, worst case we just get a notification that someone was banned in `rust-embedded`?

Mark mentioned [serde-rs in Zulip](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Triagebot.20GitHub.20App.20migration/near/572138904), which IMO is a bit more problematic, but I also understand your argument.

Urgau: > I' specifically. I think it's not worth automating this, because we need to add triagebot to the channel manually anyway, so even if the config would change the channel, we need to intervent manually.

We do have the Zulip ID for major change configured in `triagebot.toml`: https://github.com/rust-lang/compiler-team/blob/d6155a9a4f3b373c5190381281de89c56c31f3dc/triagebot.toml#L7

Kobzol: Ah, I see. That is a public channel, so it doesn't need the manual configuration though, the bans are expected to be sent to a private channel.

The configuration is also not per-repo, but kind of per-org, or globally. Do we have some global triagebot configuration today? (unless we want to configure it with an environment variable :cold_face:)

apiraino: > > I' specifically. I think it's not worth automating this, because we need to add triagebot to the channel manually anyway, so even if the config would change the channel, we need to intervent manually.
> 
> We do have the Zulip ID for major change configured in `triagebot.toml`: [rust-lang/compiler-team@`d6155a9`/triagebot.toml#L7](https://github.com/rust-lang/compiler-team/blob/d6155a9a4f3b373c5190381281de89c56c31f3dc/triagebot.toml#L7)

Yes, that's what I meant.

But to be clear: only as a FIY. I was not suggesting changing anything here.

Urgau: > The configuration is also not per-repo, but kind of per-org, or globally. Do we have some global triagebot configuration today? (unless we want to configure it with an environment variable 🥶)

That's what I mentioned [above](https://github.com/rust-lang/triagebot/pull/2269#discussion_r2770327027) about having a `triagebot.toml` file for each organization in the team repo, but it's not necessary for this PR.

Kobzol: Ah, I understand now, I misunderstood what you meant by that :facepalm: I guess that might be useful for.. some things, yeah.

src/github.rs
```diff
@@ -3176,6 +3202,125 @@ query ($owner: String!, $repo: String!, $issueNumber: Int!, $commentsCursor: Str
...
+                r#"
+query($query: String!, $issueLimit: Int!, $commentLimit: Int!) {
+  search(query: $query, type: ISSUE, first: $issueLimit) {
```
Urgau, line 3235: Just as a FWIW, I've tested this query on my-self and the results are quite mixed, it works okay for issues, but it doesn't handle review comments, so for my case some PR comment would be missing.

Not the end of world, but something to be aware of.

Kobzol: Right, I think that would need a different GraphQL node. I think that people mostly write cursed stuff in normal issue/PR comments, but if someone starts behaving funky in review comments, we can extend it :D The comment list is mostly just a nice-to-have reporting anyway.

Urgau (approved):
Let's try it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2269/dd8b73a8eaa1c449be3f9b74544e2295b482c4aa..8227e4f0692d53c88305e67d3cafa1e47b33d1c0)*

PR #2271, opened 2026-02-06
Add Zulip command for showing recent GitHub comments of a user

Kobzol:
Useful for mods (https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Feature.20request.3A.20get.20recent.20interactions.20of.20a.20given.20user/with/572441902).

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -28,7 +28,7 @@ use axum::response::IntoResponse;
 use commands::BackportArgs;
 use itertools::Itertools;
 use octocrab::Octocrab;
-use rust_team_data::v1::{Team, TeamKind, TeamMember, Teams};
+use rust_team_data::v1::{TeamKind, TeamMember};
 use secrecy::{ExposeSecret, SecretString};
 use std::cmp::Reverse;
 use std::fmt::Write as _;
@@ -369,6 +369,12 @@ async fn handle_command<'a>(
                 StreamCommand::Backport(args) => {
                     accept_decline_backport(message_data, &ctx.octocrab, &ctx.zulip, &args).await
                 }
+                StreamCommand::Comments {
+                    username,
+                    organization,
+                } => recent_comments_cmd(&ctx, gh_id, &username, &organization)
+                    .await
+                    .map(Some),
             };
         }
 
--- a/src/zulip/commands.rs
+++ b/src/zulip/commands.rs
@@ -179,6 +179,14 @@ pub enum StreamCommand {
     DocsUpdate,
     /// Accept or decline a backport.
     Backport(BackportArgs),
+    /// Show recent GitHub comments of a user in the rust-lang organization.
+    Comments {
+        /// GitHub username to look up.
+        username: String,
+        /// Organization where to find the comments.
+        #[arg(long = "org", default_value = "rust-lang")]
+        organization: String,
+    },
 }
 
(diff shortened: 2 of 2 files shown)


src/zulip.rs
```diff
@@ -465,6 +478,54 @@ async fn ping_goals_cmd(
     }
 }
 
+fn get_mod_teams(teams: &Teams) -> Vec<&Team> {
+    let mut mod_teams = vec![];
+    for name in ["mods", "mods-venut"] {
```
fmease:
```suggestion
    for name in ["mods", "mods-venue"] {
```

Kobzol: venut, rofl. Thanks :laughing:

src/zulip.rs
```diff
@@ -465,6 +478,54 @@ async fn ping_goals_cmd(
     }
 }
 
+fn get_mod_teams(teams: &Teams) -> Vec<&Team> {
+    let mut mod_teams = vec![];
+    for name in ["mods", "mods-venue"] {
```
Urgau: Should we also add the triagebot members?

I'm thinking if there are errors/issues, having access to it might help for debugging.

We could also not gate it on a specific team. We already only allow project members. [#triagebot > Feature request: get recent interactions of a given user @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Feature.20request.3A.20get.20recent.20interactions.20of.20a.20given.20user/near/572460302)

Urgau: Based on [Oli's](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Feature.20request.3A.20get.20recent.20interactions.20of.20a.20given.20user/near/572467726) response, I am fine with either adding the triagebot team or verifying whether it is a team member, as we already do for GitHub user.
https://github.com/rust-lang/triagebot/blob/1ca5cd03b0e15bb15f17d0ea1795b25d65261ecb/src/github.rs#L307-L326

src/zulip.rs
```diff
@@ -465,6 +478,54 @@ async fn ping_goals_cmd(
...
+    mod_teams
+}
+
+/// Output recent GitHub comments made by a given user in a given organization.
+/// This command can only be used by mods.
+async fn recent_comments_cmd(
+    ctx: &Context,
+    gh_id: u64,
+    username: &str,
+    organization: &str,
```
Urgau, line 493: I suppose it doesn't matter, but we could verify that the organization is one that we know of, with [`utils::is_repo_autorized`](https://github.com/rust-lang/triagebot/blob/1ca5cd03b0e15bb15f17d0ea1795b25d65261ecb/src/utils.rs#L20).

src/zulip.rs
Kobzol: Ok, changed to `is_team_member`, makes sense.

src/zulip.rs
Kobzol: Added a check for the organization.

Urgau (approved):
Thanks.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2271/1ca5cd03b0e15bb15f17d0ea1795b25d65261ecb..83e37dd005c73951ed5e70a4ba7feda77736ba6d)*

PR #2273, opened 2026-02-06
Allow users to configure per-team rotation mode

Kobzol:
Proposed in https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Extending.20review.20queue.20tracking/with/572421150.

Each reviewer can now set if they want to be assignable *through a team*. For example, if someone is a member of `t-compiler` and someone writes `r? compiler`, you can set if you want to be in the set of people who will be possibly assigned to such a PR, with `work set-team-rotation-mode compiler off/on`.

This is taken into account when computing assignments. The previous user preferences are still valid, and have higher priority in a way, so `set-rotation-mode off` will remove you from all rotations.

Best reviewed commit-by-commit.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -351,6 +351,6 @@ CREATE TABLE IF NOT EXISTS team_review_prefs (
     team TEXT NOT NULL,
     rotation_mode TEXT NOT NULL,
     UNIQUE(user_id, team)
-)
+);
 "#,
 ];


Kobzol:
Rebased.

src/zulip.rs
```diff
@@ -790,6 +790,14 @@ async fn workqueue_commands(
 
             writeln!(response, "Review capacity: `{capacity}`\n")?;
             writeln!(response, "Rotation mode: *{rotation_mode}*\n")?;
+            for (team, team_prefs) in &review_prefs.team_review_prefs {
+                writeln!(
+                    response,
+                    "Team `{team}` rotation mode: *{:?}*\n",
```
Urgau: We should make `RotationMode` implement `Display` instead of using the `Debug` impl, the impl should use the same string as `rotation_mode` above.

src/handlers/assign.rs
```diff
@@ -1254,6 +1263,14 @@ async fn candidate_reviewers_from_names<'a>(
                         username: username.clone(),
                     });
                 }
+                if let ReviewerCandidateOrigin::Expanded { team: Some(team) } = &candidate.origin
+                    && let Some(team_prefs) = review_prefs.team_review_prefs.get(team.as_str())
+                    && team_prefs.rotation_mode == RotationMode::OffRotation
+                {
+                    return Err(FindReviewerError::ReviewerOffRotation {
+                        username: username.clone(),
+                    });
```
Urgau: We may want to have a specific error type/reporting for off-rotation for team.

src/zulip.rs
Kobzol: Added function for formatting rotation mode. We could format it in different ways in different places, so I think that's better than a `Display` impl.

src/handlers/assign.rs
Kobzol: Done.

Kobzol:
Also refactored `team-stats` to reduce duplication and make it work even if you specify a repository, but there is no adhoc group for the given team in that repository.

PR #2276, opened 2026-02-08
Allow finding GitHub ID for any person in the `team` DB

Kobzol:
Previously, it only worked for *actual* team members, i.e. members of the `all` team.

Found about this while wanting to modify review preferences for a member of a project group (https://github.com/rust-lang/rust/pull/152325).

### DIFF
--- a/src/team_data.rs
+++ b/src/team_data.rs
@@ -47,16 +47,15 @@ impl TeamClient {
             .next())
     }
 
-    // Returns the ID of the given user, if the user is in the `all` team.
+    // Returns the ID of the given user, if they are in the `team` database.
     pub async fn get_gh_id_from_username(&self, login: &str) -> anyhow::Result<Option<u64>> {
-        let permission = self.teams().await?;
-        let map = permission.teams;
+        let people = self.people().await?;
         let login = login.to_lowercase();
-        Ok(map["all"]
-            .members
+        Ok(people
+            .people
             .iter()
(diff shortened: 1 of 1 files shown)


src/team_data.rs
```diff
@@ -47,16 +47,15 @@ impl TeamClient {
...
+        let people = self.people().await?;
         let login = login.to_lowercase();
-        Ok(map["all"]
-            .members
+        Ok(people
+            .people
             .iter()
-            .find(|g| g.github.to_lowercase() == login)
-            .map(|u| u.github_id))
+            .find(|(github_login, _)| github_login.to_lowercase() == login)
```
Urgau, line 57: I'm not sure this is right, the first field seems to be the GitHub name, not the login.

https://github.com/rust-lang/team/blob/592ae50f932c9b727e9c637d54d8b2d276cabb70/rust_team_data/src/v1.rs#L288-L291

https://team-api.infra.rust-lang.org/v1/people.json

Urgau: We may want to modify `rust_team_data` to include the login.

Kobzol: The key of the people dictionary is the GitHub login, that's the first item in the tuple I iterate upon.

Kobzol: Ah, the comment in team is misleading, the "name" really means login.

Urgau: I see, my mistake.

PR #2278, opened 2026-02-09
Add "View all comments" link feature

Urgau:
This PR adds a new handler that will automatically add a "View all comments" link to the body of an issue or PR once it has more than 20 (by default) comments on it, and redirect to our `/gh-comments` endpoint.

```toml
[view-all-comments]
threshold = 25
exclude-prs = true
```

cf. [#triagebot > Our own GitHub issue comments viewer @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Our.20own.20GitHub.20issue.20comments.20viewer/near/572837289)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -51,6 +51,7 @@ pub(crate) struct Config {
     pub(crate) backport: Option<BackportConfig>,
     pub(crate) range_diff: Option<RangeDiffConfig>,
     pub(crate) review_changes_since: Option<ReviewChangesSinceConfig>,
+    pub(crate) view_all_comments: Option<ViewAllCommentsConfig>,
 }
 
 #[derive(PartialEq, Eq, Debug, serde::Deserialize)]
@@ -684,6 +685,22 @@ pub(crate) struct RangeDiffConfig {}
 #[serde(deny_unknown_fields)]
 pub(crate) struct ReviewChangesSinceConfig {}
 
+/// Configuration for "View all comments" link feature
+#[derive(Default, PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub(crate) struct ViewAllCommentsConfig {
+    /// Threshold at which to add the link
+    #[serde(default)]
+    pub(crate) threshold: Option<u32>,
+    /// If issues should be prevented from having the link
+    #[serde(default)]
+    pub(crate) exclude_issues: bool,
+    /// If PRs should be prevented from having the link
+    #[serde(default)]
+    pub(crate) exclude_prs: bool,
+}
+
 fn get_cached_config(repo: &str) -> Option<MaybeConfig> {
     let cache = CONFIG_CACHE.read().unwrap();
     cache.get(repo).and_then(|(config, fetch_time)| {
@@ -950,6 +967,7 @@ mod tests {
                 backport: Some(backport_team_config),
                 range_diff: Some(RangeDiffConfig {}),
                 review_changes_since: Some(ReviewChangesSinceConfig {}),
+                view_all_comments: None,
             }
         );
     }
@@ -1039,6 +1057,7 @@ mod tests {
                 backport: None,
                 range_diff: None,
                 review_changes_since: None,
+                view_all_comments: None,
             }
         );
     }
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -684,6 +685,22 @@ pub(crate) struct RangeDiffConfig {}
...
 pub(crate) struct ReviewChangesSinceConfig {}
 
+/// Configuration for "View all comments" link feature
+#[derive(Default, PartialEq, Eq, Debug, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub(crate) struct ViewAllCommentsConfig {
+    /// Threashold at which to add the link
+    #[serde(default)]
+    pub(crate) threashold: Option<u32>,
```
Kobzol:
```suggestion
    pub(crate) threshold: Option<u32>,
```

src/config.rs
```diff
@@ -684,6 +685,22 @@ pub(crate) struct RangeDiffConfig {}
...
+    /// If issues should be prevented from having the link
+    #[serde(default)]
```
Kobzol, line 697: I find positive logic to be easier to follow, maybe `issues = true/false` and `prs = true/false` would be better? In any case, just a nitpick, I'm fine with the current state.

An alternative would be something like `enable_on = ["issues", "prs"]`, but that would be more complicated implementation-wise and is IMO not worth it.

src/handlers/view_all_comments_link.rs
```diff
@@ -0,0 +1,89 @@
...
+    config::ViewAllCommentsConfig,
+    github::{Event, Issue},
+    handlers::Context,
+};
+
+// Based on some crude experiments at around 25 events in timelines GitHub
+// starts being lazy and shows it's "Load more" button.
+//
+// Unfortunatly the webhook don't give us the number of timeline events
+// but we get do the get nuùber of comments (without doing any API calls!).
```
Kobzol:
```suggestion
// Unfortunately the webhook don't give us the number of timeline events
// but we get do the get number of comments (without doing any API calls!).
```

src/handlers/view_all_comments_link.rs
```diff
@@ -0,0 +1,89 @@
...
+//
+// So we approximate to 20 comments (+5 events) the minimum threashold.
+const DEFAULT_THREASHOLD: u32 = 20;
```
Kobzol:
```suggestion
const DEFAULT_THRESHOLD: u32 = 20;
```

src/handlers/view_all_comments_link.rs
```diff
@@ -0,0 +1,89 @@
...
+async fn add_comments_link(ctx: &Context, issue: &Issue, host: &str) -> anyhow::Result<()> {
+    let repo_name = issue.repository().full_repo_name();
+    let type_ = if issue.is_pr() { "pull" } else { "issues" };
+    let issue_number = issue.number;
+
+    let comments_link = format!(
+        "[View all comments](https://{host}/gh-comments/{repo_name}/{type_}/{issue_number})"
+    );
+
+    let new_body = if !issue.body.contains("[View all comments]") {
```
Kobzol: Don't we already have some existing logic in triagebot to append data to issue/PR bodies? 🤔 `EditIssueBody`?

Also, we don't ever have to modify the link, right? It's essentially immutable for the given issue/PR (and I don't think we would ever have to remove it, it's super rare for an issue/PR to go below the threshold if it ever was above it).

src/config.rs
Urgau: I wanted to have positive logic at first as well but trying to do the `if`s was not making my brain happy, it also meant giving up `#[serde(default)]`[^1] and using a custom default function.

Lastly I think it makes more sense to say what you don't otherwise if users says `issues = true` (with `prs`) they might think it only applies to issues, while it applies to both.

[^1]: https://github.com/serde-rs/serde/issues/368

src/handlers/view_all_comments_link.rs
Urgau: `EditIssueBody` is heavier than what is required here, we don't need to store any data in the database, nor do we want I think all the [triagebot sections](https://github.com/rust-lang/triagebot/blob/5c7955fbebbd6f33d21dc7ddd90b2845c523ac0f/src/interactions.rs#L130-L134). `EditIssueBody` also appends it's content at the end, we want it at the top here.

See also this [thread](https://github.com/rust-lang/triagebot/pull/1848#discussion_r1815444186) for more details as to why we also do this "dance" for the "Rendered" link.

src/config.rs
Kobzol: Ok, fair enough.

src/handlers/view_all_comments_link.rs
Kobzol: > EditIssueBody also appends it's content at the end, we want it at the top here.

Oh, I thought it would be at the bottom, same as the "View changes since last review" link. It might be a bit too intrusive at the top, although I can see why for PRs with a long description it's also more useful.

> See also this https://github.com/rust-lang/triagebot/pull/1848#discussion_r1815444186 for more details as to why we also do this "dance" for the "Rendered" link.

Ok, fair enough, for PRs on non-bors repos this is indeed annoying.

Urgau: > Oh, I thought it would be at the bottom, same as the "View changes since last review" link. It might be a bit too intrusive at the top, although I can see why for PRs with a long description it's also more useful.

Review comments tend IMO to be quite small, so it's not a big deal to have the link at the bottom, but for PRs and issues it's much more common to have bodies that take half or more of the size of the screen. So I think having it at the top is much more useful.

PR #2284, opened 2026-02-11
Implement GitHub App authentication

Kobzol:
This means that the [rustbot](http://github.com/rustbot) account will no longer be sending the triagebot comments, which can have... consequences :laughing: Opening as a draft for now.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,6 +1,14 @@
+# PAT token GitHub authentication, useful for local testing.
 # You can get this value from https://github.com/settings/tokens.
 # if `GITHUB_TOKEN` is not set here, the token can also be stored in `~/.gitconfig`
 GITHUB_TOKEN=MUST_BE_CONFIGURED
+# GitHub app GitHub authentication, best for production.
+# GITHUB_APP_ID=MUST_BE_CONFIGURED
+# Private key in PEM format
+# GITHUB_APP_PRIVATE_KEY=MUST_BE_CONFIGURED
+# Installation ID of the GitHub App in a specific repository or organization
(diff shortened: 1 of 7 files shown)


ehuss:
Is there a general overview of what will be required? Some questions:

- Does someone need to create the App, or is one already created? What's its name?
- Is the App installed at the org-level? Does it need to be installed in every repo?
- What permissions does the App need?
- How slow is generating the access token? (Just wondering how it affects our 10s response window.)
- What concerns do you have if it posts under the new App account?
- For posterity, we should probably explain why we are switching to an App. I assume it is for improved security? Is the short-lived token the main thing? Are we changing permission scopes?
- Does the rate limit stay the same? I seem to recall that App limits are complex (but should be the same or better than a PAT?).
- What else might go wrong?

Can we add a little more detail to "How triagebot works" in the README? Something I have neglected to do is fill that out more[^deploy], and I don't want to get further behind. Just briefly explain that the production server runs as a GitHub App, and how the app is installed (org-wide?). And reiterate it doesn't have to run as an App and can run with a PAT.

[^deploy]: For context, the thing that is missing that I've meant to include is the deploy documentation. From building the image to ECS deployment. I always forget how ECS deployment, shutdown, draining, restart, and management works.

Kobzol:
I'll answer all these, just wanted to wait for the discussion in https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Triagebot.20GitHub.20App.20migration/with/573317366 to finish, because right now it's not 100% clear that we can even technically make the switch (though probably we should be able to).

Issue #2214, opened 2025-10-29
Check again case insenstive label actions


apiraino:
Follow-up to #1915, I just noticed that *removing* `E-needs-mcve` is not case insensitive.

Strange, because *adding* the label is case insensitive 😅 

Tested [here](https://togithub.com/rust-lang/rust/pull/140156#issuecomment-3460802826).

@rustbot claim

reddevilmidzy:
(I'm not sure why an error message was shown even though the assignment worked, but...)

I think I found why adding labels works but removing them fails:

https://github.com/rust-lang/triagebot/blob/5f2081f6cf2782ca4936baad5a922c701201bf9c/src/github/issue.rs#L264-L275

In this code, `.filter(|l| self.labels().contains(l))` seems to treat the label as non-existent because `self.labels()` stores the labels with case-sensitive names. As a result, the requested label gets filtered out and the code returns early through `labels.is_empty()`.

apiraino:
@reddevilmidzy feel free to have a look and submit a patch. Can you test the triagebot locally? (In case I can do that)

Urgau:
> (I'm not sure why an error message was shown even though the assignment worked, but...)

It failed because a suspended GitHub account is still technically assigned to the issue, but the API doesn't seems to like that we tried to remove it.

reddevilmidzy:
> It failed because a suspended GitHub account is still technically assigned to the issue, but the API doesn't seems to like that we tried to remove it.

Ah, thank you for checking 

> [@reddevilmidzy](https://github.com/reddevilmidzy) feel free to have a look and submit a patch. Can you test the triagebot locally? (In case I can do that)

Yep, I'll try setting up a local test environment.

PR #2368, opened 2026-04-11
Fix case label removal

JojoFlex1:
So like when removing labels using @rustbot label -E-needs-mcve, the removal fails for labels with capital letters. This is cause the comparison in remove_labels was case-sensitive so -e-needs-mcve wouldn't match E-needs-mcve on the issue.
Adding labels works fine because GitHub's API handles case-insensitivity on their end. But removing requires an exact match with the label name as it exists on GitHub.
Now 
Changed filter and contains to filter_map + find with a case-insensitive comparison. This also ensures we use the actual label name from the issue in the DELETE request, not the user-typed version.
#2214

### DIFF
--- a/src/github/issue.rs
+++ b/src/github/issue.rs
@@ -965,7 +965,6 @@ impl IssueRepository {
     }
 }
 
-
 #[cfg(test)]
 mod tests {
     use super::Label;
@@ -973,13 +972,19 @@ mod tests {
     #[test]
     fn test_case_insensitive_label_matching() {
         let issue_labels = vec![
-            Label { name: "E-needs-mcve".to_string() },
-            Label { name: "T-compiler".to_string() },
+            Label {
+                name: "E-needs-mcve".to_string(),
+            },
+            Label {
+                name: "T-compiler".to_string(),
+            },
         ];
 
         // Simulate what remove_labels does with the fix
         let to_remove = vec![
-            Label { name: "e-needs-mcve".to_string() }, // lowercase version
+            Label {
+                name: "e-needs-mcve".to_string(),
+            }, // lowercase version
         ];
 
         let matched: Vec<Label> = to_remove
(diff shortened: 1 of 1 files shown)


apiraino:
@JojoFlex1 did you (or could you) test this change? 

Self-assigning for review.

thanks

JojoFlex1:
@apiraino at first i ran cargo build and cargo test relabel and  all 7 tests passed. So now  i added a unit test specifically for the case-insensitive matching in src/github/issue.rs.

src/github/issue.rs
```diff
@@ -959,3 +964,44 @@ impl IssueRepository {
...
+
+        let matched: Vec<Label> = to_remove
+            .into_iter()
+            .filter_map(|l| {
+                issue_labels
+                    .iter()
+                    .find(|existing| existing.name.to_lowercase() == l.name.to_lowercase())
+                    .cloned()
+            })
+            .collect();
```
marcoieni: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2368/271dd30871dc331dc268de580e0ce8cc45b0a7da..1aa594d145ac95af2014e1bd73e104b39d552bf1)*

this isn't a unit test because you aren't testing you code. I.e. you aren't calling a function of your code.

JojoFlex1:
@marcoieni  I extracted the filtering logic into a helper function find_label_case_insensitive and updated the test to call it directly.
I didn't do a live end-to-end test since remove_labels makes real GitHub API calls which makes it harder to test directly the helper function approach lets us test the core logic . If there's a better way to test this or if you have any other suggestions, Im open to learn.

apiraino:
> @marcoieni I extracted the filtering logic into a helper function find_label_case_insensitive and updated the test to call it directly. I didn't do a live end-to-end test since remove_labels makes real GitHub API calls which makes it harder to test directly the helper function approach lets us test the core logic . If there's a better way to test this or if you have any other suggestions, Im open to learn.

@JojoFlex1  End-to-end testing is detailed in the [README](https://github.com/rust-lang/triagebot#running-triagebot). If you find yourself not being able to do that, I'll get around to it (but not immediately).

JojoFlex1:
@apiraino okay sure

PR #2378, opened 2026-04-16
github: support multiple webhook secrets

ubiratansoares:
Added this support to beter manage github webhook updates minimizing the impact. 

Kept the envvar name for the sake of simplicity.

### DIFF
--- a/src/github/webhook.rs
+++ b/src/github/webhook.rs
@@ -339,12 +339,20 @@ fn check_payload_signed(signature: &str, payload: &[u8]) -> Result<(), SignedPay
         }
     };
 
-    let mut mac = Hmac::<Sha256>::new_from_slice(
-        std::env::var("GITHUB_WEBHOOK_SECRET")
-            .expect("Missing GITHUB_WEBHOOK_SECRET")
-            .as_bytes(),
-    )
-    .unwrap();
-    mac.update(payload);
-    mac.verify_slice(&signature).map_err(|_| SignedPayloadError)
+    // although the env var name suggests a single webhook secret passed in,
+    // for the sake of easing some operational tasks, internally we support
+    // a comma-separated list of them.
+    let gh_webhook_secrets =
+        std::env::var("GITHUB_WEBHOOK_SECRET").expect("Missing GITHUB_WEBHOOK_SECRET");
+
+    for secret in gh_webhook_secrets.split(',') {
+        let mut mac = Hmac::<Sha256>::new_from_slice(secret.trim().as_bytes()).unwrap();
+
+        mac.update(payload);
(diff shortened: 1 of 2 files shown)


src/github/webhook.rs
```diff
@@ -339,12 +339,20 @@ fn check_payload_signed(signature: &str, payload: &[u8]) -> Result<(), SignedPay
...
-    mac.update(payload);
-    mac.verify_slice(&signature).map_err(|_| SignedPayloadError)
+    // although the env var name suggests a single webhook secret passed in,
+    // for the sake of easing some operational tasks, internally we support
+    // a comma-separated list of them.
+    let gh_webhook_secrets =
+        std::env::var("GITHUB_WEBHOOK_SECRET").expect("Missing GITHUB_WEBHOOK_SECRET");
+
+    for secret in gh_webhook_secrets.split(',') {
+        if let Ok(mut mac) = Hmac::<Sha256>::new_from_slice(secret.as_bytes()) {
```
marcoieni: if this failed, before there was an unwrap, now there's an `if let Ok`.
Why this change? Should we at least log an error? Or is this expected to happen in case of multiple webhooks?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2378/7b6e6970460919ee692a478b298739c5b601c6ae..56c90c6fc7b8cd0c3eecdc73fe261dd02f492741)*

marcoieni:
AI suggestion:

```
Trim each secret and reject empty entries to guard against trailing commas and accidental whitespace:

for secret in gh_webhook_secrets.split(',') {
    let secret = secret.trim();
    if secret.is_empty() {
        // or also consider failing to avoid misconfigurations
        continue;
    }
    let mut mac = Hmac::<Sha256>::new_from_slice(secret.as_bytes())
        .expect("GITHUB_WEBHOOK_SECRET contains an invalid key");
    mac.update(payload);
    if mac.verify_slice(&signature).is_ok() {
        return Ok(());
    }
}


  This restores the fail-loud behavior for truly invalid keys while cleanly skipping empty segments from trailing commas.
```

But maybe it's better to do these validations at startup instead of doing them on a per request basis to avoid impacting performance

marcoieni:
This is also another nice AI suggestion:

```
Consider logging a warning at startup (not per-request) if more than one secret is configured, so operators have visibility that rotation mode is active.
```

src/github/webhook.rs
ubiratansoares: Agree here, thanks for pointing out. 100% my mistake, we should fail fast if any of the secrets don't have a valid HMAC. I'll change that!

marcoieni (approved):

PR #2380, opened 2026-04-17
Hide by-default context-only changes in our range-diff viewer

Urgau:
This PR adds an off-by-default to control the default visibility of each file diff when it only contains context changes, that is when the different is not on the `+`/`-` lines.

This makes the diff more easy to read by hiding by default some mostly useless diff.

| Off | On |
|----|---|
| image | image |

http://localhost:8000/gh-range-diff/rust-lang/rust/12f35ad39ed3e39df4d953c46d4f6cc6c82adc96..e867ac9b8da8eeecbb143192fe0f08eff1c48260/4dbafc340b0f8e9ff9e95974d0ceef81ae2652c8..d3bdea115d929178e0a3189fb4fdf0e827529ed3

### DIFF
--- a/src/gh_range_diff.rs
+++ b/src/gh_range_diff.rs
@@ -281,6 +281,15 @@ fn process_old_new(
       color: white;
       background-color: rgb(63, 128, 94);
     }}
+    .spacer {{
+      margin-bottom: 1rem;
+    }}
+    .hide {{
+      display: none;
+    }}
+    #show-context-only-changes-cb:checked ~ .hide {{
+      display: block;
+    }}
     @media (prefers-color-scheme: dark) {{
       body {{
         background: #0C0C0C;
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Maybe you could store the checkbox value into LocalStorage, to persist it across page reloads? I suspect that people who turn it off would like to do it for all or most diffs. But that's just a suggestion, feel free to merge as-is :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2380/7b6e6970460919ee692a478b298739c5b601c6ae..b5352ee2d7aef346426771e3be531d70151a2df4)*

Urgau:
> Maybe you could store the checkbox value into LocalStorage, to persist it across page reloads?

Since we don't touch the checkbox value from JS, the browser already keep the checkbox value across page reloads. Though it doesn't keep the state across different diffs, we'll see what people think.

PR #2396, opened 2026-05-06
Formatting touches to the T-compiler triage agenda

apiraino:
I've done some fine-tuning to the compile triage agenda generator.

- Removed a mostly redundant query, replaced with adding a team label
- Filtered out some more labels for FCPs we're not interested to see during triage

r? @Kobzol 

thanks!


.git-blame-ignore-revs
apiraino, line 1: Unrelated change: While I was bisecting some changes, I've added also this commit that was confusing git-blame

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2396/9c2b336714f4883998898676e1afbf2236ea8e11..e17d470485555af815f59ee90d7433a8cbb96c3b)*

src/github/issue_query.rs
```diff
@@ -161,24 +161,33 @@ impl IssuesQuery for Query<'_> {
...
+                .iter()
+                .map(|l| l.name.as_ref())
+                .collect::<Vec<&str>>();
+
+            // guess the team this issue belongs to:
+            // get the first T-* label associated with it
+            let t_label = labels
+                .iter()
+                .find(|s| s.starts_with("T-"))
+                .map_or("", |v| v);
```
Urgau, line 175: I'm not sure this will work reliably, since the order of labels is unrelated to the "owner" of the issue/PR, but let's try it.

templates/prioritization_agenda.tt
```diff
@@ -14,7 +14,7 @@ note_id: xxx
 
 - (TIP) add here non-recurrent scheduled meetings, [check the schedule calendar](https://github.com/rust-lang/calendar)
 - (TIP) mention upcoming Rust stable releases, [check the release calendar](https://calendar.google.com/calendar/htmlembed?src=l1b1gkqvfbgunjs18nemq4c580%40group.calendar.google.com)
-- Reminder: if you see a PR/issue that seems like there might be legal implications due to copyright/IP/etc, please let us know (or at least message @_**davidtwco** or @_**Wesley Wiser** so we can pass it along).
+- Reminder: if you see a PR/issue that seems like there might be legal implications due to copyright/IP/etc, please let us know (or at least message @_**davidtwco** or @_**Boxy** so we can pass it along).
```
Urgau, line 17: ❤️

templates/_issue.tt
```diff
@@ -1,11 +1,14 @@
-{% macro render(issue, with_age="", backport_branch="") %}{% if issue.is_blocked %}[blocked]{% endif %} "{{issue.title}}" [{{issue.repo_name}}#{{issue.number}}]({{issue.html_url}}){% if issue.mcp_details.zulip_link %} ([Zulip]({{issue.mcp_details.zulip_link}})){% endif %}{% if with_age %} (last review activity: {{issue.updated_at_hts}}){%- endif -%}
+{% macro render(issue, with_age="", with_team_tag="", backport_branch="") %}
+{%- if with_team_tag %}[{{issue.team}}] {% endif %}
+{%- if issue.is_blocked %}[blocked] {% endif %}"{{issue.title}}" [{{issue.repo_name}}#{{issue.number}}]({{issue.html_url}})
+{%- if issue.mcp_details.zulip_link %} ([Zulip]({{issue.mcp_details.zulip_link}})){% endif %}
```
Urgau, line 4: I'm surprised those `{% end %}` don't require dashes `-` to prevent newlines, but this seems to work as expect. Nice.

Urgau (approved):
I'm not sure about the label changes, but we can always change them later, to a big deal anyway.

Let's try it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2396/f7e38295f06ad55283fd3f4600b5295d221c50fa..1fd50cdba4d31a69085e4d36938e65fdb5f48200)*

templates/_issue.tt
apiraino: Unsure but I think it's because I eat the newline in the `{%- if ` that follows immediately.

(formatting these templates is a bit of a chore...)

PR #2398, opened 2026-05-07
Filter-out context-only hunks in our range-diff

### LINKED ISSUE #2394: range-diff: some context-only chnges are still shown
I noticed range-diff makes an effort to hide context-only changes. That's great. :)  However, it doesn't always work. Consider [this diff](https://triagebot.infra.rust-lang.org/gh-range-diff/rust-lang/miri/98cb1d3d48fa7c8e97994050bfd69e8f3da2a1a7..3f8007e19b8cac3d08fb07579bde3217c58807e4/e0b3afbdd5362aa348abc6661ed6417d1a21f58c..d811aaf958594156e2ce38365b835c361aa88de9):

Image

This is a change only in the separators between the hunks, not even in the hunk itself.

Image

Both the old and new patch here are just adding `test_address_resolution()`, the diff is entirely in the context before that change.

Image

This has both of those cases at the same time.

Urgau:
This PR filter-out context-only hunks in our range-diff.

We currently only hide-by-default file changes that only have context-only changes, but as described in #2394, a range-diff might have plenty of hunks that only have context changes (ie not a diff line, `+`/`-`).

| Currently (with context-only changes) | Currently | This PR |
|------|---|---|
| image | image | image |

Implementation wise, I had to copy and lighting edit the `UnifiedDiff` implementation from `imara-diff` in order to be able to filter-out some hunks.

I sadly had to remove the checkbox (not sure it was very useful to begin with) as knowing with lines to show/hide is highly non-trivial for context line as they can be part of the same context for multiple hunks at the same time (each hunk has +- 3 context line).

I also took the opportunity to switch to `gix-imara-diff` since the original crate has had unreleased fixes for several months.

Testing link: http://localhost:8000/gh-range-diff/rust-lang/miri/98cb1d3d48fa7c8e97994050bfd69e8f3da2a1a7..3f8007e19b8cac3d08fb07579bde3217c58807e4/e0b3afbdd5362aa348abc6661ed6417d1a21f58c..d811aaf958594156e2ce38365b835c361aa88de9

Fixes https://github.com/rust-lang/triagebot/issues/2394
cc @RalfJung

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -915,12 +915,6 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
 name = "foldhash"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
-
-[[package]]
-name = "foldhash"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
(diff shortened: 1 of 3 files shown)


Kobzol (approved):
Can't say I could examine the hunk algorithm in detail at the moment, but the output looks reasonable, and the filtering of non-interesting hunks seems valid.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2398/53ab4386fb1e521b779d4a4fe764d79c28efb988..f4603d35d5a54c2dc9b8ff0b59036b53a2401864)*

Urgau:
> Can't say I could examine the hunk algorithm in detail at the moment

I wouldn't expect you to do it, I didn't examine it in detail my-self.

The only different between our version and the one in the crate is that we give `&[Hunk]` instead of `&Diff`, so we can filter-out some hunks, the rest is strictly the same.

PR #2409, opened 2026-05-25
user-info command: username extract DWIM

apiraino:
I often use the triagebot `user-info` command with a link from GH (f.e. `https://www.github.com/apiraino`) and manually remove everything before the GH handle otherwise I get a "failed to fetch user info" error.

I would like someone else to do this onerous work :) Therefore in this patch I suggest a [DWIM (Do What I Mean)](https://en.wikipedia.org/wiki/DWIM) version of the command.

~~todo: e2e tests~~ tested

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -269,7 +269,7 @@ async fn handle_command<'a>(
             ChatCommand::UserInfo {
                 username,
                 organization,
-            } => user_info_cmd(&ctx, gh_id, username, &organization)
+            } => user_info_cmd(&ctx, gh_id, username, organization)
                 .await
                 .map(Some),
             ChatCommand::TeamStats { name, repo } => {
@@ -664,6 +664,13 @@ async fn user_info_cmd(
 
     let pr_limit = 100;
     let recent_days: u64 = 7;
+
+    // Get username from input string
+    let username = username
+        .trim_end_matches('/')
+        .rsplit_once('/')
+        .map(|(_left, right)| right)
+        .unwrap_or(username);
 
     // Load data concurrently to make the command faster
     let (user, user_prs, org_user_prs, user_repos) = futures::future::try_join4(


src/zulip.rs
```diff
@@ -665,6 +665,14 @@ async fn user_info_cmd(
     let pr_limit = 100;
     let recent_days: u64 = 7;
 
+    // Get username from input string
+    let re = regex::Regex::new("(?P<discarded>.*/)*(?P<username>.*)").unwrap();
+    let username = re
+        .captures(username)
+        .unwrap()
+        .name("username")
+        .map_or(username, |m| m.as_str());
```
Urgau: A regex is a bit over kill IMO for just splitting on the last `/`, we could just do it with `rsplit_once` [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b013312ec8d35a627bc31fd474016623).

However not even checking that it's a link is a bit iffy to me, why not just use `strip_prefix` [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d784b070a0d989f2eb4d3ceb52fa7549)?

```suggestion
    // Get username from input string
    let username = username
        .strip_prefix("https://www.github.com/")
        .unwrap_or(username);
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2409/754c9cda5154996e33f083287d818214d8e90ffc..7c0484dd077c57d3030acabaca261b88bdc55847)*

apiraino: because (as expected) it would only catch the case where I paste the full GH link.

I tested the regexp with:
```
    let ss = [
        "/",
        "apiraino",
        "https://www.github.com/apiraino",
        "/apiraino",
        "https://github.com/apiraino",
        "www.github.com/apiraino",
        "github.com/apiraino",
    ];
```

maybe overkill? I wanted to check as many cases as I could think of :shrug:

Urgau: In that case the `rsplit_once` [should do it](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=dea5416512004c8591e290b402783427).

```suggestion
    // Get username from input string
    let username = username 
        .trim_end_matches('/')
        .rsplit_once('/')
        .map(|(_left, right)| right)
        .unwrap_or(username);
```

apiraino: :shrug:

Urgau (approved):
LGTM, thanks.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2409/a88792b170141288b79eac6e3bafaabce7828767..727e8674930cb5ff861f939b3b357f50b58a889c)*

PR #2410, opened 2026-05-25
Use GitHub `relative-time-element` library to display elapsed time

### LINKED ISSUE #2408: Ambiguous date format in gh-comments
The gh-comments view renders dates in a way that is ambiguous to interpret in an international context:

Image

([example](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rfcs/pull/3946))

Is this July 4th or April 7th? Most countries (AFAIK) put the day first, but the US puts the month first, causing endless confusion.

Github itself renders this unambiguously:

Image

Would be nice if triagebot could also use a date format that is less prone to misinterpretation. :)

Urgau:
Seems like `toLocaleString` JS function is not perfect and can be confusing to user when displaying the datetime, in particular due to `day/month` vs `month/user` confusion.

Instead of displaying the exact date time, let's use GitHub own [`relative-time-element` library](https://www.npmjs.com/package/@github/relative-time-element) to display elapsed time instead.

| Currently | This PR |
|-----|-----|
| image | image |
| image | image |

Fixes https://github.com/rust-lang/triagebot/issues/2408
cc @Ralfjung

### DIFF
--- a/src/gh_comments.rs
+++ b/src/gh_comments.rs
@@ -32,6 +32,7 @@ use crate::{
 pub const STYLE_URL: &str = "/gh-comments/style@0.0.8.css";
 pub const MARKDOWN_URL: &str = "/gh-comments/github-markdown@20260117.css";
 pub const SELF_CONTAINED_URL: &str = "/gh-comments/self_contained@0.0.3.js";
+pub const RELATIVE_TIME_ELEMENT_URL: &str = "/gh-comments/relative-time-element@5.0.0.js";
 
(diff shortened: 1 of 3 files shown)


Kobzol (approved):
Oh, that's cool that it is available as a standalone library.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2410/754c9cda5154996e33f083287d818214d8e90ffc..7840e8b160cfb6044bf5c39dd7cc715d6a5e9c76)*

PR #2412, opened 2026-05-28
Add basic `merge` command for merge queues

Urgau:
This pull-request adds a basic `merge` to triagebot, it's currently only intended to work with native GitHub merge queues (so no `bors` support, or normal merge/squash/rebase support).

It's short termm goal is to be able to delegate the merging to the author (or some arbitrary person? tbd), like it's currently possible in `bors`, where a reviewer often ask for a change and than `r=me`. Probably with some `@bot delegate+` like bors.

Technically we don't much, we check the user GitHub permissions on the repo (at least write) and if it's okay we enqueue the PR in the merge queue. Nothing else.

I was unable to fully test it, since I need a public org with a public repo to be able to setup a merge queue, but the error handling works well. I'm planning on testing it in this repo.

image

This was requested by Clippy (cc @samueltardieu).

### DIFF
--- a/parser/src/command/merge.rs
+++ b/parser/src/command/merge.rs
@@ -1,0 +1,15 @@
+use crate::error::Error;
+use crate::token::{Token, Tokenizer};
+
+#[derive(PartialEq, Eq, Debug)]
+pub struct MergeCommand;
+
+impl MergeCommand {
+    pub fn parse<'a>(input: &mut Tokenizer<'a>) -> Result<Option<Self>, Error<'a>> {
+        if let Some(Token::Word("merge")) = input.peek_token()? {
+            Ok(Some(Self))
+        } else {
+            Ok(None)
+        }
+    }
+}
--- a/src/github/issue.rs
+++ b/src/github/issue.rs
@@ -987,7 +987,17 @@ impl IssueRepository {
 
 #[derive(Debug, serde::Deserialize)]
 pub struct CollaboratorPermission {
-    pub permission: String,
+    pub permission: GitHubPermission,
+}
+
+// https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2026-03-10#get-repository-permissions-for-a-user
+#[derive(Debug, serde::Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum GitHubPermission {
+    Admin,
+    Write,
+    Read,
+    None,
 }
 
 impl IssueRepository {
@@ -999,5 +1009,14 @@ impl IssueRepository {
         let url = format!("{}/collaborators/{username}/permission", self.url(client));
         let permission = client.json(client.get(&url)).await?;
         Ok(permission)
+    }
+}
+
+impl GitHubPermission {
+    pub fn has_write_permissions(&self) -> bool {
+        match self {
+            GitHubPermission::Admin | GitHubPermission::Write => true,
+            GitHubPermission::Read | GitHubPermission::None => false,
+        }
     }
 }
--- a/src/github/issue.rs
+++ b/src/github/issue.rs
@@ -565,15 +565,20 @@ impl Issue {
         Ok(commits)
     }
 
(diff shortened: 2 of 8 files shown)


Kobzol:
I'm a bit confused, what's the point of this command without having the ability to delegate to e.g. the PR author? The way it is implemented now, you can only use the command if you can also click on the merge button, no?

Urgau:
I wanted to do it in two steps, first the add the `merge` command (this PR) and second add the `delegate` command (implemented in this [tree](https://github.com/Urgau/triagebot/tree/merge-delegate-command)), but if you prefer I can add both in this PR.

parser/src/command/merge.rs
```diff
@@ -0,0 +1,15 @@
+use crate::error::Error;
+use crate::token::{Token, Tokenizer};
+
+#[derive(PartialEq, Eq, Debug)]
+pub struct MergeCommand;
```
Kobzol, line 5:
```suggestion
pub struct AddToMergeQueueCommand;
```
The text command can stay `merge`, but I think that this is more accurate with respect to what the command does.

src/github/issue.rs
```diff
@@ -622,6 +627,26 @@ impl Issue {
...
 
+    /// Enqueues this pull request into the repository's merge queue.
+    pub async fn enqueue_to_merge_queue(&self, client: &GithubClient) -> anyhow::Result<()> {
+        let pr_id = self.graphql_issue_id(client).await?;
+
+        client
+            .graphql_query(
+                "mutation ($pullRequestId: ID!) {
+                  enqueuePullRequest(input: { pullRequestId: $pullRequestId }) {
+                    clientMutationId
```
Kobzol, line 638:
```suggestion

```
This is not used for anything currently, right?

src/handlers/merge.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    let has_perm = {
+        let perm = issue_comment
+            .issue
+            .repository()
+            .collaborator_permission(&ctx.github, &issue_comment.comment.user.login)
+            .await
+            .context("failed to get the user repository permission")?;
+
+        perm.permission == "write" || perm.permission == "admin"
```
Kobzol: Since this is the second place where we use this permission, it would be nice to map it to an enum, to avoid working with arbitrary strings.

Kobzol (commented):
Ok, let's land it per-partes. Tested it [here](https://github.com/lerncz/citest/pull/47#issuecomment-4705368402), seems to work 🎉

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2412/361982f82066163615fbd47d09e0a0a791b93188..bd030d71e52029fc77b87092295c7b1b660b0ce3)*

parser/src/command/merge.rs
Urgau: So the command would be `add-to-merge-queue` as well? or do you just want to rename the internal struct?

Kobzol: Actually, I take it back. After I read the rest of the PR, it seems like you plan to maybe eventually have other ways of merging. So `merge` as a general concept is fine, nevermind.

src/github/issue.rs
Urgau: It's necessary, even if unused afterwars:
> Field must have selections (field 'enqueuePullRequest' returns EnqueuePullRequestPayload but has no selections. Did you mean 'enqueuePullRequest { ... }'?)
    Variable $pullRequestId is declared by anonymous mutation but not used

Kobzol: What fun :)

Kobzol (approved):
Thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2412/361982f82066163615fbd47d09e0a0a791b93188..8335caa8e561c40cb751b5089fa7897434cc1837)*

PR #2413, opened 2026-05-28
Show which users reacted in our GitHub comments viewer

Urgau:
People have been asking since the beginning (and even at the All-Hands) about seeing who reacted on a issue body, comments, ... and I always though it wasn't possible.

Turns out we can actually fetch the users who reacted to each individual reaction. (I'm just really bad at reading GraphQl)

Capture d’écran du 2026-05-28 20-01-59

### DIFF
--- a/src/gh_comments.rs
+++ b/src/gh_comments.rs
@@ -888,33 +888,54 @@ fn write_reaction_groups_as_html(
     buffer: &mut String,
     reaction_groups: &[GitHubGraphQlReactionGroup],
 ) -> anyhow::Result<()> {
-    let any_reactions = reaction_groups.iter().any(|rg| rg.users.total_count > 0);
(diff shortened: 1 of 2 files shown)


Kobzol (approved):
`GitHubGraphQlReactionGroupReactors` is sure a mouthful 😆 Maybe we could sort the displayed users, but not sure if GraphQL doesn't already return them in non-deterministic order.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2413/c4d01629e0de377455210600423b5c6972d95184..7741bdf0a35bcdda5ae7a5dfe3e10e9279e990d1)*

Urgau:
> Maybe we could sort the displayed users, but not sure if GraphQL doesn't already return them in non-deterministic order.

GraphQL returns them in the same order as GitHub UI, which I suspect is the order people clicked on it. I don't think we should change that.

PR #2414, opened 2026-05-28
Add `lock`/`unlock` commands for GitHub issues and pull-requests

Urgau:
While talking with the moderation team @oli-obk expressed the desired to have triagebot commands to lock and unlock issues and pull-requests on repos they don't have write access.

This PR therefore adds the following commands:
 - On GitHub:
   - `lock`
   - `unlock`
 - On Zulip (stream and DM):
   - `unlock [--org <org>] <repo> <id>` *(to be able to unlock after locking, otherwise can't post a comment)*

I didn't add a `lock` command on Zulip as it seems preferable to do it in the open, let me @oli-obk if you also want a `lock` command on Zulip.

All the commands are restricted to team members and the GitHub commands are also behind the `[lock]` config (which we should enable org-wide after some basic testing).

I've tested the GitHub commands on my repo without an issue. I couldn't test the Zulip command unfortunately.

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -10,7 +10,7 @@ use crate::db::users::DbUser;
 use crate::github::queries::user_comments_in_org::UserComment;
 use crate::github::queries::user_prs::PullRequestState;
 use crate::github::queries::user_prs::UserPullRequest;
-use crate::github::{self, Repository};
+use crate::github::{self, PullRequestNumber, Repository};
 use crate::handlers::Context;
 use crate::handlers::docs_update::docs_update;
 use crate::handlers::pr_tracking::{ReviewerWorkqueue, get_assigned_prs};
@@ -261,6 +261,11 @@ async fn handle_command<'a>(
         let output = match &cmd {
             ChatCommand::Whoami => whoami_cmd(&ctx, gh_id).await,
             ChatCommand::Lookup(cmd) => lookup_cmd(&ctx, cmd).await,
+            ChatCommand::Unlock {
+                organization,
+                repo,
+                id,
+            } => unlock_cmd(&ctx, gh_id, organization, repo, *id).await,
             ChatCommand::Work(cmd) => workqueue_commands(&ctx, gh_id, cmd).await,
             ChatCommand::PingGoals(args) => {
                 ping_goals_cmd(ctx.clone(), gh_id, message_data, args).await
@@ -390,6 +395,11 @@ async fn handle_command<'a>(
                     };
                     Ok(None)
                 }
+                StreamCommand::Unlock {
+                    organization,
+                    repo,
+                    id,
+                } => unlock_cmd(&ctx, gh_id, &organization, &repo, id).await,
             };
         }
 
@@ -634,6 +644,51 @@ async fn ping_goals_cmd(
             "That command is only permitted for those running the project-goal program.",
         ))
     }
+}
+
+/// Unlock a specific issue in our managed repos.
+/// This command can only be used by team members.
+async fn unlock_cmd(
+    ctx: &Context,
+    gh_id: u64,
+    org: &str,
+    repo: &str,
+    issue_id: PullRequestNumber,
+) -> anyhow::Result<Option<String>> {
+    let gh_login = ctx
+        .team
+        .username_from_gh_id(gh_id)
+        .await?
+        .ok_or_else(|| anyhow::anyhow!("Username for GitHub user {gh_id} not found"))?;
+
+    if !ctx.team.is_team_member(&gh_login).await? {
+        return Err(anyhow::anyhow!(
+            "This command is only available to team members."
+        ));
+    }
+
+    if ctx.team.repos().await?.repos.get(org).is_none() {
+        return Err(anyhow::anyhow!(
+            "Organization `{org}` is not managed by the team database."
+        ));
+    }
+
+    let issue_repo = github::IssueRepository {
+        organization: org.to_string(),
+        repository: repo.to_string(),
+    };
+
+    let issue = ctx
+        .github
+        .issue(&issue_repo, issue_id)
+        .await
+        .context(format!("Could not get {issue_repo}#{issue_id} details"))?;
+
+    issue.unlock(&ctx.github).await?;
+
+    Ok(Some(format!(
+        "Successfully unlocked {issue_repo}#{issue_id}!"
+    )))
 }
 
 /// Output recent GitHub activity made by a given user (both globally and in a given organization).
@@ -1108,6 +1163,7 @@ fn get_cmd_impersonation_mode(cmd: &ChatCommand) -> ImpersonationMode {
         | ChatCommand::PingGoals(_)
         | ChatCommand::UserInfo { .. }
         | ChatCommand::TeamStats { .. }
+        | ChatCommand::Unlock { .. }
         | ChatCommand::Lookup(_) => ImpersonationMode::Disabled,
         ChatCommand::Whoami => ImpersonationMode::Silent,
         ChatCommand::Work(cmd) => match cmd {
--- a/src/zulip/commands.rs
+++ b/src/zulip/commands.rs
@@ -17,6 +17,16 @@ pub enum ChatCommand {
     /// Inspect or modify your reviewer workqueue.
     #[clap(subcommand)]
     Work(WorkqueueCmd),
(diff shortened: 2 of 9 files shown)


oli-obk:
> I didn't add a `lock` command on Zulip as it seems preferable to do it in the open, let me @oli-obk if you also want a `lock` command on Zulip.

for now not. There is a larger discussion going on on making it opaque as to which moderator performed an action or sent a message, but we'll address that holistically in case we want to go down that route

apiraino:
>  I couldn't test the Zulip command unfortunately.

I'll test this (give me some time, I'll update this comment)

src/zulip/commands.rs
```diff
@@ -17,6 +17,15 @@ pub enum ChatCommand {
     /// Inspect or modify your reviewer workqueue.
     #[clap(subcommand)]
     Work(WorkqueueCmd),
+    /// Unlock a specific GitHub issue or pull-request.
+    Unlock {
+        /// Repository that is specifically being queried.
+        /// Example: `rust-lang/rust`.
```
apiraino: Can we assume it will always be under `rust-lang/` or do we want to leave the door open to act on other organization?

The difference would be between writing `unlock rust-lang/rust 123456` and `unlock rust 123456` (shorter and less prone to typos)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2414/7bd318187b4a428151b1c11b9b1e6ca07e7b932d..97291818c70a141487175677afb82763fb96d5dd)*

src/zulip.rs
```diff
@@ -1108,6 +1159,7 @@ fn get_cmd_impersonation_mode(cmd: &ChatCommand) -> ImpersonationMode {
         | ChatCommand::PingGoals(_)
         | ChatCommand::UserInfo { .. }
         | ChatCommand::TeamStats { .. }
+        | ChatCommand::Unlock { .. }
```
apiraino, line 1166: unless there is a specific reason, the `unlock` command can be a `StreamCommand` to allow using it everywhere on Zulip

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2414/7bd318187b4a428151b1c11b9b1e6ca07e7b932d..97291818c70a141487175677afb82763fb96d5dd)*

src/zulip.rs
```diff
@@ -636,6 +637,56 @@ async fn ping_goals_cmd(
     }
 }
 
+/// Unlock a specific issue in our managed repos.
+/// This command can only be used by team members.
+async fn unlock_cmd(
+    ctx: &Context,
+    gh_id: u64,
+    repo: &str,
+    issue_id: u64,
```
apiraino: for consistency elsewhere maybe use `crate::github::PullRequestNumber`?

(unsure about it because we have a bit of both everywhere)

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2414/7bd318187b4a428151b1c11b9b1e6ca07e7b932d..97291818c70a141487175677afb82763fb96d5dd)*

src/zulip.rs
```diff
@@ -636,6 +637,56 @@ async fn ping_goals_cmd(
     }
 }
 
+/// Unlock a specific issue in our managed repos.
+/// This command can only be used by team members.
+async fn unlock_cmd(
```
apiraino, line 651: the zulip cmd doesn't seem to work. looking into it

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2414/7bd318187b4a428151b1c11b9b1e6ca07e7b932d..97291818c70a141487175677afb82763fb96d5dd)*

apiraino (changes requested):

src/zulip/commands.rs
```diff
@@ -17,6 +17,15 @@ pub enum ChatCommand {
...
+        #[arg(long)]
```
apiraino: can we remove this so that the user doesn't need to write `unlock --repo <repo-name> <id>`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2414/7bd318187b4a428151b1c11b9b1e6ca07e7b932d..97291818c70a141487175677afb82763fb96d5dd)*

apiraino (changes requested):

src/zulip/commands.rs
Urgau: So only, `unlock <repo-name> <id>` ?

src/zulip/commands.rs
Urgau: Yes, changed it to assume `rust-lang` org.

src/zulip.rs
Urgau: Changed the command to be possible on stream as well.

src/zulip/commands.rs
Urgau: Changed the command to be `unlock [--org <org>] <repo> <id>` (with org being optional and defaulting to `rust-lang`).

src/zulip.rs
apiraino: finally got it (the setup was a bit finicky)

src/zulip/commands.rs
```diff
@@ -17,6 +17,16 @@ pub enum ChatCommand {
     /// Inspect or modify your reviewer workqueue.
     #[clap(subcommand)]
     Work(WorkqueueCmd),
+    /// Unlock a specific GitHub issue or pull-request.
+    Unlock {
+        /// Orgnization that is specifically being queried.
+        #[arg(long = "org", default_value = "rust-lang")]
+        organization: String,
```
apiraino, line 24: Now that you have removed the need to specify the `organization` and we are assuming `rust-lang`, all this can go away (otherwise it will interfere).

src/zulip.rs
```diff
@@ -636,6 +646,51 @@ async fn ping_goals_cmd(
...
+    ctx: &Context,
+    gh_id: u64,
+    org: &str,
```
apiraino, line 654: the `org` can now go away because we assume `rust-lang` and If you prefer, you can do something like [here](https://github.com/rust-lang/triagebot/blob/master/src/zulip.rs#L423):
```
    let organization = std::env::var("MAIN_GH_REPO_OWNER").unwrap_or("rust-lang".to_string());
```

(it will help when testing with a different GH orga)

apiraino (changes requested):
2 small things

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2414/7bd318187b4a428151b1c11b9b1e6ca07e7b932d..2ef73e5eb5db4496d2c96cfd32d693b9b54b0815)*

src/zulip/commands.rs
Urgau: We are only assuming the *default* is `rust-lang`, I think we still want the ability to be able to specify one of the other orgs we handle.

apiraino: so how is the command supposed to be now? can you give me some examples?

src/zulip.rs
Urgau: That's a good idea, I've update all the commands to make use of those two environment variables.

src/zulip/commands.rs
Urgau: - `unlock rust 123` (defaults to `rust-lang` org)
- `unlock --org rust-lang-nursery rust-toolstate 14` (use `rust-lang-nursery` as the org)

you can also just re-specify `rust-lang` as the org:
- `unlock --org rust-lang rust 123`

src/zulip.rs
apiraino: After that, probably what you wrote [here](https://github.com/rust-lang/triagebot/pull/2414#discussion_r3325992718) will work (right now the command invocations don't work)

src/zulip/commands.rs
apiraino: grafik

The exclamation mark is a bit cringe but we can fix that later XD

apiraino (approved):
LGTM! thanks for your patience @Urgau and thanks for the patch 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2414/7bd318187b4a428151b1c11b9b1e6ca07e7b932d..509c94c419e35344df9f4ae376e82cb3b92f296d)*

PR #2415, opened 2026-05-29
Make the cursor style reflect that it can be clicked on hidden comment

Urgau:
image

image

cc @apiraino (since we talked about at RustWeek)

### DIFF
--- a/src/gh_comments.rs
+++ b/src/gh_comments.rs
@@ -29,7 +29,7 @@ use crate::{
     utils::{immutable_headers, is_known_and_public_repo},
 };
 
-pub const STYLE_URL: &str = "/gh-comments/style@0.0.8.css";
+pub const STYLE_URL: &str = "/gh-comments/style@0.0.9.css";
 pub const MARKDOWN_URL: &str = "/gh-comments/github-markdown@20260117.css";
 pub const SELF_CONTAINED_URL: &str = "/gh-comments/self_contained@0.0.3.js";
 pub const RELATIVE_TIME_ELEMENT_URL: &str = "/gh-comments/relative-time-element@5.0.0.js";
--- a/src/gh_comments.rs
+++ b/src/gh_comments.rs
@@ -787,7 +787,7 @@ fn write_review_thread_as_html(
         r###"
       <details class="review-thread" data-expandable="{default_open}" {open}>
         <summary class="review-thread-header">
-            <span>{path_html}{status}</span>
+            <span><span class="indicator"></span>{path_html}{status}</span>
         </summary>
 
         <div class="review-thread-comments">
--- a/src/gh_comments.rs
+++ b/src/gh_comments.rs
@@ -496,7 +496,7 @@ pub async fn gh_comments(
     headers.insert(
         CONTENT_SECURITY_POLICY,
         HeaderValue::from_static(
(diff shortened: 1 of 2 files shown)


apiraino:
maybe I am missing something: what is the difference after this patch?

Also now the cursors becomes a 'hand'

<details><summary>Details</summary>

https://github.com/user-attachments/assets/7e9c5645-f237-4e08-be38-c67739b95d96

</details> 


The thing that I think is missing, is a hint when a comment is collapsed

<details><summary>Details</summary>

grafik

</details>

Urgau:
> what is the difference after this patch?

regular comments (not reviews or review comments, as they already show the hand) now show a hand (to indicate that it can be clicked) instead of the regular cursor pointer

Urgau:
> The thing that I think is missing, is a hint when a comment is collapsed

Doesn't the "hidden" text, next to the time, already serve that purpose?

EDIT: I see what you mean, we don't show this text for review threads.

Urgau:
What about like this?

|Collapsed|Expanded|
|---|----|
|image|image|

PS: Doing the fold/unfold at the right is bit tricky, there doesn't seems to be unicode set of characters we could use, and none of [the arrays](https://codepoints.net/arrows) are convincing.

apiraino:
> What about like this?

definitively better! Do you think it's possible to add this svg inline, to make it look like just like GitHub's?

<details><summary>Collapse SVG</summary>

```

<svg aria-hidden="true" data-component="Octicon" height="16" viewBox="0 0 16 16" version="1.1" width="16" data-view-component="true" class="octicon octicon-unfold">
    <path d="m8.177.677 2.896 2.896a.25.25 0 0 1-.177.427H8.75v1.25a.75.75 0 0 1-1.5 0V4H5.104a.25.25 0 0 1-.177-.427L7.823.677a.25.25 0 0 1 .354 0ZM7.25 10.75a.75.75 0 0 1 1.5 0V12h2.146a.25.25 0 0 1 .177.427l-2.896 2.896a.25.25 0 0 1-.354 0l-2.896-2.896A.25.25 0 0 1 5.104 12H7.25v-1.25Zm-5-2a.75.75 0 0 0 0-1.5h-.5a.75.75 0 0 0 0 1.5h.5ZM6 8a.75.75 0 0 1-.75.75h-.5a.75.75 0 0 1 0-1.5h.5A.75.75 0 0 1 6 8Zm2.25.75a.75.75 0 0 0 0-1.5h-.5a.75.75 0 0 0 0 1.5h.5ZM12 8a.75.75 0 0 1-.75.75h-.5a.75.75 0 0 1 0-1.5h.5A.75.75 0 0 1 12 8Zm2.25.75a.75.75 0 0 0 0-1.5h-.5a.75.75 0 0 0 0 1.5h.5Z"></path>
</svg>

```

</details>

Urgau:
> Do you think it's possible to add this svg inline, to make it look like just like GitHub's?

Maybe not as inline svg, since it needs to change depending on the state (there is two icons).

But the original SVG are open-source MIT Licensed: [fold](https://github.com/primer/octicons/blob/main/icons/fold-24.svg) and [unfold](https://github.com/primer/octicons/blob/main/icons/unfold-24.svg)

Let's me see what I can do.

Urgau:
What about now?
image

apiraino:
looks great! Pretty much GitHub's own design

Kobzol (approved):

PR #2417, opened 2026-05-29
Delay by 1 min adding the review changes since link on review comments

Urgau:
This PRs delay by 1 minute the addition of the review changes since link on review comments, to hopefully be less annoying for contributors.

The 1 minute delay is subject to change.

Impl-wise it's implemented as a one-off job, as a simple `tokio::spawn` wouldn't be less work, as we need to re-fetch the review comment body anyway, so doing it properly is not much more work.

Context: [#triagebot > Rustbot "view changes since the review" makes editing harder @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Rustbot.20.22view.20changes.20since.20the.20review.22.20makes.20editing.20harder/near/598563160)

cc @RalfJung

### DIFF
--- a/src/handlers/review_changes_since.rs
+++ b/src/handlers/review_changes_since.rs
@@ -1,11 +1,13 @@
 use std::sync::{Arc, LazyLock};
 
 use anyhow::Context as _;
+use async_trait::async_trait;
+use chrono::{Duration, Utc};
 
 use crate::{
     cache,
     config::ReviewChangesSinceConfig,
-    github::{Comment, Event, Issue, IssueCommentAction, IssueCommentEvent},
+    github::{Comment, Event, Issue, IssueCommentAction, IssueCommentEvent, IssueRepository},
     handlers::Context,
 };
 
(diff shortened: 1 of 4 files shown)


src/handlers/review_changes_since.rs
```diff
@@ -137,19 +139,84 @@ pub(crate) async fn handle(
...
-                event
-                    .issue
-                    .edit_review_comment(&ctx.github, event.comment.id, &new_body)
-                    .await
-                    .context("failed to update the review comment body")?;
+                // unfortunaly, review comments are not updated by GitHub in the UI, which
+                // creates friction for contributors as they might want to edit the review
+                // comment, but get blocked since we modified it in the mean time
+                //
+                // we therefor defer adding the link by one minute
```
Kobzol:
```suggestion
                // unfortunately, review comments are not updated by GitHub in the UI, which
                // creates friction for contributors as they might want to edit the review
                // comment, but get blocked since we modified it in the mean time
                //
                // we therefore defer adding the link by one minute
```

src/handlers/review_changes_since.rs
```diff
@@ -137,19 +139,84 @@ pub(crate) async fn handle(
...
+                let pr_repo = event.issue.repository();
+                let args = AddReviewChangesSinceLinkJobArgs {
+                    org: pr_repo.organization.clone(),
+                    repo: pr_repo.repository.clone(),
+                    pr_id: event.issue.number,
+                    review_comment_id: event.comment.id,
+                    link,
+                };
+
+                crate::db::schedule_job(
```
Kobzol, line 157: If two such jobs are scheduled at the exact same time, only one would be executed. But that's an edge-case that we don't have to deal with, I think.

Kobzol (approved):
I kinda hate this =D I wonder if we shouldn't instead change the logic in some larger way, e.g. send a triagebot comment if a review without body is posted, to have a singular place for the link, or only add the links after a push happens or something.

But technically it seems like it should work.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2417/361982f82066163615fbd47d09e0a0a791b93188..73c3e0e2b83517ac171b3a711637503cb445b9dc)*

Urgau:
> I kinda hate this =D

Oh me too, this is just a giant hack.

> I wonder if we shouldn't instead change the logic in some larger way, e.g. send a triagebot comment if a review without body is posted, to have a singular place for the link, or only add the links after a push happens or something.

Here me out, what about sending the link as a reply to the review comment? That should avoid the concurrent edit for good, but it's a bit gross.

Kobzol:
I think that is worse, it would be too much noise.

src/handlers/review_changes_since.rs
```diff
@@ -137,19 +139,84 @@ pub(crate) async fn handle(
...
+        };
+
+        let pr = ctx
+            .github
+            .pull_request(&pr_repo, args.pr_id)
+            .await
+            .context("failed to get the pr")?;
+
+        let review_comment = pr
+            .get_review(&ctx.github, args.review_comment_id)
```
Urgau: I think that's wrong, we should call https://docs.github.com/en/rest/pulls/comments?apiVersion=2026-03-10#get-a-review-comment-for-a-pull-request

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2417/361982f82066163615fbd47d09e0a0a791b93188..9bc78fbac8b8e5bb1735bee36ba7ab3657d44141)*

Urgau: Fixed

Urgau:
I've tested the logic locally, works as expect. Let's try it.

src/handlers/review_changes_since.rs
```diff
@@ -137,19 +139,84 @@ pub(crate) async fn handle(
...
-                event
-                    .issue
-                    .edit_review_comment(&ctx.github, event.comment.id, &new_body)
-                    .await
-                    .context("failed to update the review comment body")?;
+                // unfortunately, review comments are not updated by GitHub in the UI, which
+                // creates friction for contributors as they might want to edit the review
+                // comment, but get blocked since we modified it in the mean time
+                //
+                // we therefore defer adding the link by one minute
```
Urgau, line 146: Does is it work?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2417/a88792b170141288b79eac6e3bafaabce7828767..f87404f131b9fc805643af808517afec75cf8079)*

PR #2418, opened 2026-05-29
Restore welcome messages by also checking for the `None` association

Urgau:
Context: [#triagebot > Welcome message isn't being posted](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Welcome.20message.20isn.27t.20being.20posted/with/598629929)

### DIFF
--- a/src/github/issue.rs
+++ b/src/github/issue.rs
@@ -1,6 +1,5 @@
 use anyhow::Context;
 use chrono::Utc;
-use octocrab::models::AuthorAssociation;
 use reqwest::StatusCode;
 use std::fmt;
 use std::sync::OnceLock;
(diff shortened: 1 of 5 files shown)


Kobzol (commented):
Since we know that `FirstTimer` and `FirstTimeContributor` is essentially borked, I would prefer making the `author_association` field private, and exposing a method that will return `true` if the issue's author is a first-time contributor (well, it's a guess, but at least it's something). Otherwise we can unknowingly start using the first timer variants elsewhere and not recognize that they don't work (I think we already use them at one other place, IIRC).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2418/361982f82066163615fbd47d09e0a0a791b93188..3f9faacc3c4a726bbbeb39873227c645eb552823)*

Urgau:
I added a newtype around the `octocrab` enum so we only expose methods. Wdyt?

Kobzol (approved):
Looks great, thank you!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2418/361982f82066163615fbd47d09e0a0a791b93188..8ecf940ece39f212ff968cd312e099e68c72a9ee)*

Issue #2419, opened 2026-05-29
Testing issue for lock handler


Urgau:
See https://github.com/rust-lang/triagebot/pull/2416

Urgau:
trying locking a second time

@rustbot lock

Urgau:
gonna try unlocking via Zulip

Urgau:
everything seems to be working as expect

@rustbot close

PR #2420, opened 2026-05-29
Enable `[lock]` handler in the `rust-lang` organization

Urgau:
I've done some extra testing in https://github.com/rust-lang/triagebot/issues/2419, everything worked as expected (both from GitHub and Zulip).

I think we are ready to enable both commands org-wide, so that they can be used on all triagebot-enabled repository by mods and members.

cc @oli-obk

### DIFF
--- a/rust-lang.triagebot.toml
+++ b/rust-lang.triagebot.toml
@@ -45,6 +45,13 @@ excluded-repos = [
   "rust-lang/team", # no write permissions
 ]
 
+# Allow mods/members to lock/unlock an issue or PR (`@rustbot lock`)
+# Documentation at: https://forge.rust-lang.org/triagebot/lock.html
+[lock]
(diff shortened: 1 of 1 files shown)


apiraino (approved):
we will slightly change the command line invocation (as discussed with Urgau: removing the `--org` portion) but considering that there are very few users, it should be fine :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2420/a88792b170141288b79eac6e3bafaabce7828767..a4fd51bd530cafad773e86492381c9ee8977d9f3)*

Urgau:
Yeah, I don't expect anyone to use it for now.

Also the `--org` part is Zulip only, which already doesn't need the `[lock]` handler to be enabled.

PR #2421, opened 2026-05-31
Disable welcome message if no automatic assignment is configured

Urgau:
This follows what we did in https://github.com/rust-lang/triagebot/pull/2040

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -185,15 +185,27 @@ pub(super) async fn handle_input(
             }
         } else if event.issue.author_association.is_probably_first_timer() {
             let assignee_text = match &assignee {
-                Some(assignee) => messages::welcome_with_reviewer(&assignee.name),
-                None => messages::WELCOME_WITHOUT_REVIEWER.to_string(),
+                Some(assignee) => Some(messages::welcome_with_reviewer(&assignee.name)),
+                None => {
+                    // If the assign fallback group is empty, then we don't expect any automatic
+                    // assignment, and this message would just be spam.
+                    if config.fallback_review_group().is_some() {
+                        Some(messages::WELCOME_WITHOUT_REVIEWER.to_string())
+                    } else {
+                        None
+                    }
+                }
             };
-            let mut welcome = messages::new_user_welcome_message(&assignee_text);
-            if let Some(contrib) = &config.contributing_url {
-                welcome.push_str("\n\n");
-                welcome.push_str(&messages::contribution_message(contrib, &ctx.username));
+            if let Some(assignee_text) = assignee_text {
+                let mut welcome = messages::new_user_welcome_message(&assignee_text);
+                if let Some(contrib) = &config.contributing_url {
+                    welcome.push_str("\n\n");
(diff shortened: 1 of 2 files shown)


src/handlers/assign/messages.rs
```diff
@@ -68,7 +68,7 @@ Please choose another assignee by using `r? @reviewer`."
     )
 }
 
-pub const WELCOME_WITHOUT_REVIEWER: &str = "@Mark-Simulacrum (NB. this repo may be misconfigured)";
+pub const WELCOME_WITHOUT_REVIEWER: &str = "a reviewer";
```
Kobzol, line 71: I like this change.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2421/cc9ef00d2011f13c1d8485b36677cd9a5cb96959..ea760008ab69ba18bc65d07c78082e0dd413a862)*

src/handlers/assign.rs
```diff
@@ -185,15 +185,27 @@ pub(super) async fn handle_input(
             }
         } else if event.issue.author_association.is_probably_first_timer() {
             let assignee_text = match &assignee {
-                Some(assignee) => messages::welcome_with_reviewer(&assignee.name),
-                None => messages::WELCOME_WITHOUT_REVIEWER.to_string(),
+                Some(assignee) => Some(messages::welcome_with_reviewer(&assignee.name)),
```
Kobzol, line 188: I'm not sure if we also need the other changes? If there is no fallback group, or if there is a fallback group, but all fallback reviewers have their queue full, then it still seems good to post at least something on the PR. It doesn't mean that no one will ever look at the PR.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2421/cc9ef00d2011f13c1d8485b36677cd9a5cb96959..ea760008ab69ba18bc65d07c78082e0dd413a862)*

Urgau: Well in https://github.com/rust-lang/triagebot/pull/2040, people asked us to not show a message if they don't have automatic assignment is configured, so I followed that.

Otherwise it's a bit weird I think to have a welcome message the first time, but nothing after, or you think it's fine?

Kobzol: By automatic assignment you mean just having `[assign]` in `triagebot.toml`? :thinking: Because if that is not enabled, then the `assign` handler and the welcome machinery doesn't run at all, right?

Kobzol (approved):
Let's do it! 😆

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2421/cc9ef00d2011f13c1d8485b36677cd9a5cb96959..ea760008ab69ba18bc65d07c78082e0dd413a862)*

src/handlers/assign.rs
Urgau: For the record, we talked about this at [#triagebot > Welcome message isn't being posted @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Welcome.20message.20isn.27t.20being.20posted/near/599177582)

PR #2422, opened 2026-05-31
Add commit message diff in our range-diff implementation

Urgau:
It has been mentioned to me by multiple people that they would like to see the different commit messages in our range-diff, like is done in the official `git range-diff`.

This PR therefore adds such feature. It's not exactly the same, as we show the changes across the full range, and not per commit as is done by `git range-diff`.

As such I added a new section called "Commits" which list all the commits and the diff if there is a change in the message.

| New commits | Changed message |
|--------|-----------|
| image| image |

Testing links:
 - http://localhost:8000/gh-range-diff/rust-lang/rust/6769f690f947f12e36db6b6503bab265b7b2cced..87d009021e6e5aa59fd17fe91751de91e084db27/1a70f8d36e818b45829a5c065fd6004c9047e8d1..57e3c83223818777372d156134c56ee31af0b1a6
 - http://localhost:8000/gh-range-diff/rust-lang/rust/3bf5c6d99bc8a0c0d5b2f69826ed4f6d256a0a21..b3687efe9eaed1c1de123100ece0e54d9e59787f/3bf5c6d99bc8a0c0d5b2f69826ed4f6d256a0a21..c24ec53db05b64c5a9817160aae04029da58dd24

cc @camelid

### DIFF
--- a/src/github/repos.rs
+++ b/src/github/repos.rs
@@ -1108,6 +1108,8 @@ pub struct GitTreeObject {
 #[derive(Debug, serde::Deserialize)]
 pub struct GitUser {
     pub date: DateTime<FixedOffset>,
+    pub name: Option<String>,
+    pub email: Option<String>,
 }
 
 #[derive(Debug, serde::Deserialize)]
--- a/src/gh_range_diff.rs
+++ b/src/gh_range_diff.rs
@@ -385,7 +385,11 @@ fn process_old_new(
 
         // Show the changes if there are any hunks to be shown
         if !hunks.is_empty() {
-            let printer = HtmlDiffPrinter(&input.interner, filename);
+            let printer = HtmlDiffPrinter {
+                interner: &input.interner,
+                filename: Some(filename),
+                mode: HtmlDiffPrinterMode::DoubleDiff,
+            };
             let unified_diff = CustomUnifiedDiff {
                 printer: &printer,
                 hunks: &hunks,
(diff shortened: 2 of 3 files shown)


src/github/repos.rs
```diff
@@ -1108,6 +1108,8 @@ pub struct GitTreeObject {
 #[derive(Debug, serde::Deserialize)]
 pub struct GitUser {
     pub date: DateTime<FixedOffset>,
+    pub name: String,
```
Kobzol: I think that at least the e-mail can be missing (https://docs.rs/octocrab/latest/octocrab/models/struct.SimpleUser.html#structfield.name). I definitely ran into this in bors. I'd perhaps make both an `Option`, but definitely the e-mail should be an `Option`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2422/cc9ef00d2011f13c1d8485b36677cd9a5cb96959..06185559f0bdc02054606beedb97c7ff533f6ee8)*

Urgau: Tbh, I'm not even sure having the author and email is that useful.

Urgau: Alright, I made both optional.

Kobzol: I think the message is the most important thing in the diff, yeah.

Kobzol (approved):
Thanks! Really cool feature.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2422/8b17696fc71cb5f6f4a99ea791f034962868d2c7..51886155b6e3c8959e13bc481cb7259e027e3586)*

PR #2426, opened 2026-06-03
Add the concept of community reviews before PR assignment

Urgau:
This PR adds the concept of community reviews before PR assignment.

It's enabled in the `[assign]` handler with a new config:
```toml
[assign.community_reviews]
minimum_approvals = 2
label = "S-waiting-on-community-reviews"
```

The normal workflow for community reviews is like this:
1. PR is being opened *(non-draft, otherwise ignore until ready)*
    * Add the label
    * Post the welcome message if new user, with small explanation of community reviews
        * No message for recurring users 
3. Min approval reviews is reached (or label is manually removed)
    * Normal PR assignment
    * Assignment message

If someone is being assigned with our commands (`r?`, reroll), we remove the label after the assignment, thus preventing further auto-assign.

I tested what I could on my private repo, the core logic seems to work, although I couldn't test everything since I'm alone.

cc @samueltardieu

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,6 +3,7 @@ use crate::github::{GithubClient, Repository};
 use parser::command::relabel::{Label, LabelDelta, RelabelCommand};
 use std::collections::{HashMap, HashSet};
 use std::fmt;
+use std::num::NonZeroU8;
 use std::sync::{Arc, LazyLock, RwLock};
 use std::time::{Duration, Instant};
 use tracing as log;
@@ -160,6 +161,9 @@ pub(crate) struct AssignConfig {
     pub(crate) warn_non_default_branch: WarnNonDefaultBranchConfig,
     /// A URL to include in the welcome message.
     pub(crate) contributing_url: Option<String>,
+    /// Community reviews before auto-assignement
+    #[serde(default)]
+    pub(crate) community_reviews: Option<AssignCommunityReviewsConfig>,
     /// Ad-hoc groups that can be referred to in `owners`.
     #[serde(default)]
     pub(crate) adhoc_groups: HashMap<String, Vec<String>>,
@@ -192,6 +196,15 @@ impl AssignConfig {
     pub(crate) fn fallback_review_group(&self) -> Option<&[String]> {
         self.adhoc_groups.get("fallback").map(Vec::as_slice)
     }
+}
+
+#[derive(PartialEq, Eq, Debug, Clone, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct AssignCommunityReviewsConfig {
+    /// The minimum number of community reviews before automatic assignment is performed
+    pub(crate) minimum_approvals: NonZeroU8,
+    /// Label indicating that community reviews are required
+    pub(crate) label: String,
 }
 
 #[derive(PartialEq, Eq, Debug, Clone, serde::Deserialize)]
@@ -993,6 +1006,7 @@ mod tests {
                     users_on_vacation: HashSet::from(["jyn514".into()]),
                     review_prefs: None,
                     custom_messages: None,
+                    community_reviews: None,
                 }),
                 note: Some(NoteConfig { _empty: () }),
                 ping: Some(PingConfig { teams: ping_teams }),
@@ -1120,6 +1134,7 @@ mod tests {
                     owners: HashMap::new(),
                     users_on_vacation: HashSet::new(),
                     review_prefs: None,
+                    community_reviews: None,
                 }),
                 note: None,
                 ping: None,
@@ -1166,6 +1181,23 @@ mod tests {
             config.assign.and_then(|c| c.review_prefs),
             Some(AssignReviewPrefsConfig {})
         ));
+    }
+
+    #[test]
+    fn assign_review_min_reviews() {
+        let config = r#"
+            [assign.community_reviews]
+            minimum_approvals = 2
+            label = "S-waiting-on-community-reviews"
+        "#;
+        let config = toml::from_str::<Config>(&config).unwrap();
+        assert_eq!(
+            config.assign.and_then(|c| c.community_reviews),
+            Some(AssignCommunityReviewsConfig {
+                minimum_approvals: NonZeroU8::new(2).unwrap(),
+                label: "S-waiting-on-community-reviews".to_string(),
+            })
+        );
     }
 
     #[test]
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -22,14 +22,18 @@
 //! `assign.owners` config, it will auto-select an assignee based on the files
 //! the PR modifies.
 
+use crate::config::AssignCommunityReviewsConfig;
 use crate::db::issue_data::IssueData;
 use crate::db::review_prefs::{RotationMode, get_review_prefs_batch};
 use crate::errors::{self, AssignmentError, user_error};
 use crate::handlers::pr_tracking::ReviewerWorkqueue;
 use crate::{
     config::AssignConfig,
-    github::{self, Event, FileDiff, Issue, IssuesAction, utils::Selection},
-    handlers::{Context, GithubClient, IssuesEvent},
+    github::{
+        self, Comment, Event, FileDiff, Issue, IssueCommentAction, IssueCommentEvent, IssuesAction,
+        PullRequestReviewState, utils::Selection,
+    },
+    handlers::{Context, IssuesEvent},
     interactions::EditIssueBody,
 };
 use anyhow::{Context as _, bail};
@@ -84,6 +88,7 @@ fn get_repo_workqueue(ctx: &Context, repo: &str) -> Arc<RwLock<ReviewerWorkqueue
 pub(super) enum AssignInput {
     Opened { draft: bool },
     ReadyForReview,
+    UnlabeledMinReviews { draft: bool },
 }
 
 /// Prepares the input when a new PR is opened.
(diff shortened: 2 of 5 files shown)


src/handlers/assign/messages.rs
```diff
@@ -6,11 +6,18 @@
...
-The Rust team is excited to review your changes, and you should hear from {reviewer} \
+The Rust project is excited to review your changes, and you should hear from {reviewer} \
 some time within the next two weeks."
     )
 }
 
+pub fn new_user_welcome_message_community_reviews(min_reviews: u8) -> String {
+    format!(
+        "Thanks for the pull request, and welcome!\
+You should hear from the project after a minimum of {min_reviews} community reviews have been reached."
```
blyxyas: Could you make this message configurable? We'd love to link [this blog post](https://github.com/rust-lang/blog.rust-lang.org/pull/1869), at least temporarily, and possibly make a reviewing guide.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2426/83104c9f1d195d26c05c5d8202268779d2b5676b..583bfebb5ee2453c1de9f038acf1014d615a2b0a)*

blyxyas (changes requested):

src/handlers/assign/messages.rs
Urgau: We already have a configurable link in the `[assign]` handler: [`contributing_url`](https://github.com/rust-lang/rust-clippy/blob/master/triagebot.toml#L90). You could change it to your blog post or link it to inside your contributing file. (The associated text is [defined here](https://github.com/Urgau/triagebot/blob/583bfebb5ee2453c1de9f038acf1014d615a2b0a/src/handlers/assign/messages.rs#L21-L32))

But in general I would like to avoid making more messages configurable, it's a pain to deal with, but if the option above doesn't work let me know.

Urgau: Btw, feel free to suggest any text changes. I didn't think too much about them.

src/handlers/assign/messages.rs
```diff
@@ -6,11 +6,18 @@
 pub fn new_user_welcome_message(reviewer: &str) -> String {
     format!(
         "Thanks for the pull request, and welcome! \
-The Rust team is excited to review your changes, and you should hear from {reviewer} \
+The Rust project is excited to review your changes, and you should hear from {reviewer} \
```
Kobzol:
```suggestion
The Rust Project is excited to review your changes, and you should hear from {reviewer} \
```

src/handlers/assign/messages.rs
Kobzol: Maybe something like this?
```suggestion
        "Thanks for the pull request, and welcome!\
You should hear from one of our reviewers after this PR is reviewed by at least {min_reviews} reviewers from the community."
```
Plus pluralizing the `s` if you feel up for it :P

src/handlers/assign.rs
```diff
@@ -115,20 +134,82 @@ pub(super) async fn handle_input(
...
+                    event
+                        .issue
+                        .post_comment(&ctx.github, &welcome)
+                        .await
+                        .context("couldn't post welcome message")?;
+                }
+                false
+            } else {
+                // PR was opened normally, perform assignment if not assignees manually set
+                event.issue.assignees.is_empty()
```
Kobzol, line 170: This used to be true, any specific reason for changing it? Though it makes sense to ignore auto assign if there already are some selected assignees.

src/handlers/assign.rs
```diff
@@ -321,11 +414,63 @@ fn is_self_assign(assignee: &str, pr_author: &str) -> bool {
...
+        .filter(|c| c.pr_review_state == Some(PullRequestReviewState::Approved))
+        .map(|c| c.user.id)
+        .collect::<HashSet<_>>();
+
+    Ok(approval_reviews.len() >= config.minimum_approvals.get().into())
+}
+
+/// Determine if the issue has enough approval reviews and update the PR
+/// label to refect it.
+async fn update_and_has_enough_approvals(
```
Kobzol: Not a very CQRS-like function 😆

```suggestion
/// label to reflect it.
async fn update_and_has_enough_approvals(
```

```suggestion
/// Update the min community reviews label on the PR.
/// Returns true if the PR already has enough community approvals.
async fn update_community_review_label(
```

src/handlers/assign.rs
```diff
@@ -321,11 +414,63 @@ fn is_self_assign(assignee: &str, pr_author: &str) -> bool {
...
+    ctx: &Context,
+    config: &AssignCommunityReviewsConfig,
+    issue: &Issue,
+) -> anyhow::Result<bool> {
+    let reviews = issue
+        .get_reviews(&ctx.github)
+        .await
+        .context("unable to fetch reviews for determining assignement")?;
+
+    let approval_reviews: HashSet<_> = reviews
```
Kobzol: This assumes that PR authors cannot approve their own PRs. That is currently true on GitHub, but it would be good to document that :) Please add a comment here to explain it.

src/handlers/assign.rs
```diff
@@ -321,11 +414,63 @@ fn is_self_assign(assignee: &str, pr_author: &str) -> bool {
     assignee.to_lowercase() == pr_author.to_lowercase()
 }
 
+/// Determine if the issue has enough approval reviews
+async fn has_enough_approvals(
```
Kobzol:
```suggestion
async fn has_enough_community_approvals(
```

src/handlers/assign.rs
```diff
@@ -797,6 +955,48 @@ pub(super) async fn handle_command(
     Ok(())
 }
 
+/// Process a comment/review/review comment
+pub(super) async fn handle_comment(
```
Kobzol, line 963: We could also handle this in the "main" assignment handler, to keep the logic centralized, to avoid yet another handler dispatcher in `handlers.rs`.

src/config.rs
```diff
@@ -194,6 +198,15 @@ impl AssignConfig {
     }
 }
 
+#[derive(PartialEq, Eq, Debug, Clone, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct AssignCommunityReviewsConfig {
+    /// The minimum number of community reviews before automatic assignment
+    pub(crate) minimum_approvals: NonZeroU8,
+    /// Label indicating that comminuty reviews are required
```
Kobzol:
```suggestion
    /// Label indicating that community reviews are required
```

src/config.rs
```diff
@@ -194,6 +198,15 @@ impl AssignConfig {
     }
 }
 
+#[derive(PartialEq, Eq, Debug, Clone, serde::Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct AssignCommunityReviewsConfig {
+    /// The minimum number of community reviews before automatic assignment
```
Kobzol:
```suggestion
    /// The minimum number of community reviews before automatic assignment is performed
```

Kobzol (commented):
Thank you, this looks great, and relatively simple.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2426/83104c9f1d195d26c05c5d8202268779d2b5676b..583bfebb5ee2453c1de9f038acf1014d615a2b0a)*

src/handlers/assign.rs
Urgau: Unfortunately we can't do it in the main handler, it only receives issues actions, not issue/PR comments; at least with the current setup.

However we could migrate away the `assign` handler from the `issue_handlers!` to a custom handling that handles both, but that's much more involved I think.

Kobzol: Ah, I didn't realize that sending a review is not a PR action. Nevermind then!

src/handlers/assign.rs
Urgau: No particular reason, it was just easier for me to follow the logic. I think it already is the case, the condition is just below.

src/handlers/assign.rs
Urgau: I added a comment and removed the PR author (just in case).

src/handlers/assign.rs
```diff
@@ -321,11 +414,67 @@ fn is_self_assign(assignee: &str, pr_author: &str) -> bool {
...
+    ctx: &Context,
+    config: &AssignCommunityReviewsConfig,
+    issue: &Issue,
+) -> anyhow::Result<bool> {
+    let reviews = issue
+        .get_reviews(&ctx.github)
+        .await
+        .context("unable to fetch reviews for determining assignement")?;
+
+    let mut approval_reviews: HashSet<_> = reviews
```
Kobzol, line 428: This is an edge case, but have you tried what happens if a user first does an approval, and then later posts Changes requested? 😆 Does the previous review stay there, and would it be thus counted as approved?

Kobzol (approved):
Left one additional comment, but that's not super important I think. Feel free to merge, great work!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2426/0f9de7f2dd19faed7c23b0d231b07d61323b0c23..130460d4de7d89ba5da45dc6fc73ee5899c1ae23)*

src/handlers/assign.rs
Urgau: No, I haven't tested it, but I'm pretty sure it will only count it as approved since the logic only count approval reviews, not the final state per user.

Urgau: I don't think it matters much though, as the only case where it may be unexpected is if it's the first review (assuming 2 approvals min), ie, User 1: Approve, User 1: Changes requested, User 2: Approve; which seems like an edge case.

I don't it's worth the extra complexity for now, we can improve the logic if people find it unexpected.

PR #2427, opened 2026-06-04
Allow text after claim/unclaim/reroll commands

### LINKED ISSUE #284: Allowing text after command?
I just tried to add an explanatory text after `@rustbot claim` but it seems the command parser doesn't accept this:

https://github.com/rust-lang/rust/issues/56982#issuecomment-578482365

Allowing this would bring triagebot in line with bors as well

reddevilmidzy:
I don't think it's necessary to require a newline at the end of a command. Do you agree?

close: https://github.com/rust-lang/triagebot/issues/284

### DIFF
--- a/parser/src/command.rs
+++ b/parser/src/command.rs
@@ -268,6 +268,27 @@ fn resumes_after_code() {
 }
 
 #[test]
+fn text_after_claim() {
+    let input = "@bot claim I'd like to work on this";
+    let mut input = Input::new(input, vec!["bot"]);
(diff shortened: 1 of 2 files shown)


Urgau (approved):
We are generally pretty relaxed about it in the other commands too, so this makes sense to do it here as well.

Looks good to me.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2427/83104c9f1d195d26c05c5d8202268779d2b5676b..ba7786c9e0e4436343f0f4d9033bda44c78f2e47)*

PR #2430, opened 2026-06-07
Add automatic release of inactive claimed assignments

Urgau:
This PR adds a system to automatically release inactive claimed (`claim` and `assign <username>` commands) assignments.

The logic is loosely inspired by [Zulip's bot](https://github.com/zulip/zulipbot/blob/7a83a0ab804169a328860cbb1d0fa6acd68e85c9/src/events/activity.ts#L111). We check after `inactivity_reminder` days the last update on the issue and linked PRs (via the cross-references), we post the reminder message and if nothing moves `inactivity_limit - inactivity_reminder` days later we remove the assignment and post a comment.

It's configured like this:
```toml
[assign.check_activity]
inactivity_reminder = 6 #days
inactivity_limit = 7 #days
```

Output for the reminder:
> Hey @Urgau, just checking in! It's been 6 days since the last update on this issue or linked PRs.
> 
> No worries if you're busy! If you still want to tackle this, just drop a quick comment below. If you've run out of time or interest, you can release it by commenting `@rustbot release-assignment` so another contributor can jump in.
> 
> If we don't hear from you in 1 days, you will be automatically unassigned.

Output for the release assignment:
> Hey @Urgau, just letting you know we've freed up this issue since we haven't seen any updates for 7 days. If you find the time later on, feel free to reclaim it and pick up where you left off.

Context: [#triagebot > Releasing lingering assignments](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Releasing.20lingering.20assignments/with/597163574)
cc @samueltardieu

### DIFF
--- a/src/handlers.rs
+++ b/src/handlers.rs
@@ -11,7 +11,7 @@ use std::fmt;
 use std::sync::Arc;
 use tracing as log;
 
-mod assign;
+pub(crate) mod assign;
 mod autolabel;
 mod backport;
 mod bot_pull_requests;
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -49,6 +49,7 @@ use tokio_postgres::Client as DbClient;
 use tracing as log;
 
 mod messages;
+pub(crate) mod release_inactive_assignments;
 
 #[cfg(test)]
 mod tests {
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -60,6 +60,9 @@ mod tests {
 const GHOST_ACCOUNT: &str = "ghost";
 
 /// Key for the state in the database
+const ASSIGN_KEY: &str = "ASSIGN";
+
+/// Key for the state in the database
 const PREVIOUS_REVIEWERS_KEY: &str = "previous-reviewers";
 
 /// State stored in the database
@@ -893,7 +896,7 @@ pub(super) async fn handle_command(
     } else {
         let mut client = ctx.db.get().await;
         let mut e: EditIssueBody<'_, AssignData> =
-            EditIssueBody::load(&mut client, issue, "ASSIGN").await?;
+            EditIssueBody::load(&mut client, issue, ASSIGN_KEY).await?;
         let d = e.data_mut();
 
         let to_assign = match cmd {
--- a/src/handlers/assign/release_inactive_assignments.rs
+++ b/src/handlers/assign/release_inactive_assignments.rs
@@ -1,0 +1,247 @@
+//! This module defines the logic for releasing inactive issue assignments
+
+use std::collections::HashMap;
+
+use anyhow::Context as _;
+use async_trait::async_trait;
+use chrono::{DateTime, Duration, Utc};
+use serde::{Deserialize, Serialize};
+
+const INACTIVITY_REMINDER_DAYS: i64 = 49; // 7 weeks
+const INACTIVITY_LIMIT_DAYS: i64 = 56; // 8 weeks
+const INACTIVITY_GRACE_PERIOD_DAYS: i64 = INACTIVITY_LIMIT_DAYS - INACTIVITY_REMINDER_DAYS;
+
+use crate::{
+    db::issue_data::IssueData,
+    github::{IssueNumber, IssueRepository, ReportedContentClassifiers, utils::Selection},
+    handlers::Context,
+    jobs::Job,
+};
+
+pub(crate) const RELEASE_INACTIVE_ASSIGNMENTS_JOB_NAME: &str = "release_inactive_assignments";
+
+pub(crate) struct ReleaseInactiveAssignmentsJob;
+
+#[async_trait]
+impl Job for ReleaseInactiveAssignmentsJob {
+    fn name(&self) -> &'static str {
+        RELEASE_INACTIVE_ASSIGNMENTS_JOB_NAME
+    }
+
+    async fn run(&self, ctx: &Context, _metadata: &serde_json::Value) -> anyhow::Result<()> {
+        tracing::info!("Starting job to release inactive assignments");
+
+        for (owner, repo) in [("rust-lang", "rust-clippy")] {
+            tracing::info!("Starting checking {owner}/{repo}");
+            match update_assignments(ctx, owner, repo).await {
+                Ok(()) => {
+                    tracing::info!("Checking of {owner}/{repo} succesful");
+                }
+                Err(err) => {
+                    tracing::error!("Failed to check {owner}/{repo}: {err}");
+                }
+            }
+        }
+
+        tracing::info!("Job finished to release inactive assignments");
+        Ok(())
+    }
+}
+
+async fn update_assignments(ctx: &Context, owner: &str, repo: &str) -> anyhow::Result<()> {
+    let issues_assigned = ctx
+        .github
+        .issues_assigned(owner, repo)
+        .await
+        .context("unable to load the assigned issues")?;
+
+    let prs_with_closing_issues_references = ctx
+        .github
+        .closing_issues_references(owner, repo)
+        .await
+        .context("unable to load the closing issues references")?;
+
+    tracing::info!("Loaded {} assigned issues", issues_assigned.len());
+    tracing::info!(
+        "Loaded {} PRs (with their closing issues references)",
+        prs_with_closing_issues_references.len()
+    );
+
+    let references: HashMap<IssueNumber, Vec<&_>> = prs_with_closing_issues_references
+        .iter()
+        .flat_map(|pr| {
+            pr.closing_issues_references
+                .iter()
+                .map(move |i| (i.number, pr))
+        })
+        .fold(HashMap::new(), |mut map, (issue_num, pr)| {
+            map.entry(issue_num).or_default().push(pr);
+            map
+        });
+
+    let now = Utc::now();
+    let owner_repo = format!("{owner}/{repo}");
+
+    for issue_assigned in issues_assigned {
+        let [assignee] = &issue_assigned.assignees[..] else {
+            continue;
+        };
+
+        let last_update_time = references
+            .get(&issue_assigned.number)
+            .iter()
+            .flat_map(|references| references.iter().map(|r| r.updated_at))
+            .max()
+            .unwrap_or(DateTime::<Utc>::MIN_UTC)
+            .max(issue_assigned.updated_at);
+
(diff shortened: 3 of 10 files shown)


src/handlers/assign/activity.rs
```diff
@@ -0,0 +1,314 @@
+use std::fmt::Display;
```
Kobzol: Can you please add a module doc comment to explain what this handler does? Thanks. Maybe also rename the module to `check_activity`, or `release_inactive_assignments`?

src/handlers/assign/activity.rs
```diff
@@ -0,0 +1,314 @@
...
+    fn name(&self) -> &'static str {
+        ASSIGN_CHECK_ACTIVITY_JOB_NAME
+    }
+
+    async fn run(&self, ctx: &super::Context, metadata: &serde_json::Value) -> anyhow::Result<()> {
+        let input: AssignCheckActivityInput = serde_json::from_value(metadata.clone())
+            .context("unable to deserialize the metadata in major change acceptance job")?;
+
+        let now = Utc::now();
+        tracing::info!("processing: {input:?}");
```
Kobzol: This log doesn't make it very obvious what is it about, because we don't use spans around jobs, I think, maybe just add `self.name()` to it.

src/handlers/assign/activity.rs
```diff
@@ -0,0 +1,314 @@
...
+            }
+        }
+    }
+}
+
+async fn check_activity(
+    ctx: &super::Context,
+    input: &AssignCheckActivityInput,
+    now: DateTime<Utc>,
+) -> anyhow::Result<()> {
```
Kobzol: Using downcasts for error is not something I particularly enjoy seeing in the codebase 😆 What do you think about inverting the error representation, and adding a `GenericError(anyhow::Error)` variant to `ActivityCheckingError`, and adding a `From` impl to convert from anyhow errors to it via `?`.

src/handlers.rs
```diff
@@ -11,7 +11,7 @@ use std::fmt;
 use std::sync::Arc;
 use tracing as log;
 
-mod assign;
+pub(crate) mod assign;
```
Kobzol, line 14: Jobs are not handlers, and even though I know that there are some jobs inside the handlers module currently, I don't find that very readable. What do you think about having a separate module (directory) for jobs?

src/handlers/assign.rs
```diff
@@ -773,25 +775,47 @@ pub(super) async fn handle_command(
...
+                    "This issue has been assigned to @{} via [this comment]({}).",
+                    &to_assign.login,
                     event.html_url().unwrap()
                 );
                 e.apply(&ctx.github, cmt_body).await?;
             }
             Err(e) => return Err(e.into()),
         }
+
+        if let Some(activity_config) = &config.check_activity {
```
Kobzol: If it's implemented like this, it won't help clippy much, because they will still have 100+ legacy assigned issues for which the check won't trigger.

With this implementation, the number of jobs will scale with the number of claimed issues, which means that over time we might have hundreds or thousands of jobs in the repo, and I don't think that's really necessary.

I would prefer a different system, where we have a single job per repo (which has the config activated), which runs e.g. once per day, and checks all claimed issues. I guess that the problem is that triagebot is probably not really architectured to have such jobs at the moment 🤔

src/handlers/assign/activity.rs
```diff
@@ -0,0 +1,314 @@
...
+        .check_activity
+        .as_ref()
+        .ok_or(ActivityCheckingLogicError::MissingAssignCheckActivityConfig)?;
+
+    let issue = repo
+        .get_issue(&ctx.github, input.issue)
+        .await
+        .context("unable to get the associated issue")?;
+
+    if !issue.is_open() || issue.draft {
```
Kobzol: This is only for issues, not PRs, right? Those can't be draft, AFAIK?

src/github/queries/cross_references.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CrossReferenceSource {
+    pub updated_at: DateTime<Utc>,
+}
+
+impl Issue {
+    /// Fetches basic public information about a GitHub user.
```
Kobzol: Copy-pasted comment.

src/github/queries/cross_references.rs
```diff
@@ -0,0 +1,67 @@
...
+        &self,
+        client: &GithubClient,
+    ) -> anyhow::Result<Vec<CrossReference>> {
+        let mut data = client
+            .graphql_query(
+                r#"
+query ($owner: String!, $repo: String!, $issue: Int!) {
+  repository(owner: $owner, name: $repo) {
+    issue(number: $issue) {
+      timelineItems(first: 100, itemTypes: [CROSS_REFERENCED_EVENT]) {
```
Kobzol: Can we filter for *last* 100?

src/github/queries/cross_references.rs
```diff
@@ -0,0 +1,67 @@
...
+        nodes {
+          ... on CrossReferencedEvent {
+            createdAt
+            source {
+              ... on Issue {
+                updatedAt
```
Kobzol: Do we know what does `updatedAt` mean? I wonder if it corresponds to what we think of as "there was some activity on the PR".

src/handlers/assign/activity.rs
```diff
@@ -0,0 +1,314 @@
...
+                    ),
+                )
+                .await
+                .context("failed to post inactivity comment")?;
+
+            schedule_activity_job(
+                ctx,
+                AssignCheckActivityInput {
+                    warning_details: Some(WarningDetails {
+                        posted_at: comment.created_at.unwrap(),
```
Kobzol: I'd just use the current time, to avoid depending on GitHub to set the `created_at` field 😆 🙈

Kobzol (commented):
This is a pretty complex piece of functionality, I kinda wish we had bors' infra here, with that it would be relatively simple to implement robustly 😆

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2430/3d741225c7b85a3ad11774471786b19a60c04759..f519407b47034f203989755687fe2488bed677c8)*

src/handlers/assign/activity.rs
Urgau: `tracing` add the module names to the logs which makes it kind-of okay, but I see your point, I will add the job name.

src/handlers/assign.rs
Urgau: I though of doing a once per day job, but that has several issues:
 1. clippy currently has 2500 issues, just getting all the issues and assignees for them is probably going to cost us our 5000 req/h ratelimit, that not even counting getting the PRs, the pulls links, ...
    1. it also wouldn't scale for rust-lang/rust which I think we also want to handle, and it has >10 000 issues
 2. we only want to handle assignee's via `claim`/`assign` commands, not manual assignment, which would be difficult (impossible?) to support

So for those reason I think having small targeted jobs is better than cron-like job.

> they will still have 100+ legacy assigned issues for which the check won't trigger

Yeah, that's unfortunate. I think someone is gonna have to go through all of them and unassigned them manually (with a copy/paste text should be "quick").

src/github/queries/cross_references.rs
Urgau: As far as I know it correspond to a manual activity on the issue: a new comment, an update of a comment, new labels, ... but not any change to the timeline (cross-references).

Urgau: It's also what zulipbot is using.

src/handlers/assign/activity.rs
Urgau: Using our current time would be problematic here, as we check the `updatedAt` field, which will match the `created_at` field of the comment if nothing changes; using something else risk having a discrepancy with GitHub.

src/handlers/assign.rs
Kobzol: Most of those issues are unassigned (https://github.com/rust-lang/rust-clippy/issues?q=is%3Aissue%20state%3Aopen%20no%3Aassignee), so maybe we could fetch only the assigned ones? https://docs.github.com/en/graphql/reference/issues#input-object-issuefilters

Or maybe better, simply record claimed issues in the DB, which makes the whole thing much simpler, and solves the issue 2. (which is annoying to deal with using GH API only, we'd have to check `IssueData` on each issue to see if it was claimed).

src/handlers/assign/activity.rs
Kobzol: Well if we handle timezones correctly, it shouldn't matter much. But this is a small thing, no need to change it.

src/handlers.rs
Urgau: > What do you think about having a separate module (directory) for jobs?

Conceptually I like the idea, but when the handler and the job is so tightly integrated into each other I think it's better to have them close to each other.

Urgau:
I re-rewrote the whole logic as daily cron-job that checks all the opened issues. I haven't tested it, and it's a little rough on the edges.

Unsure if it's simpler, I will let you be the judge of it. ;-)

src/handlers/assign/release_inactive_assignments.rs
```diff
@@ -0,0 +1,247 @@
...
+#[async_trait]
+impl Job for ReleaseInactiveAssignmentsJob {
+    fn name(&self) -> &'static str {
+        RELEASE_INACTIVE_ASSIGNMENTS_JOB_NAME
+    }
+
+    async fn run(&self, ctx: &Context, _metadata: &serde_json::Value) -> anyhow::Result<()> {
+        tracing::info!("Starting job to release inactive assignments");
+
+        for (owner, repo) in [("rust-lang", "rust-clippy")] {
```
Kobzol, line 34: Ah, yes, this is a big Achilles' heel of triagebot. I haven't looked at the rest of the implementation yet, but I think that we should finally move triagebot to a GitHub App, so that we have an explicit list of repositories that are actually using it, so that we can avoid hardcoding repositories like this.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2430/98f85d92ba7547ce6a1634e5883a899cce7a5411..23b699490e47aa5e9647dbb83ceb769e52e2a63e)*

Urgau:
I'm gonna close this PR, it's clear that I don't have the motivation to pursue it.

The current version is roughly what Jakub wanted, so if someone wants to pursue it, feel free.

Issue #2403, opened 2026-05-13
Split meeting agenda generators


marcoieni:
People would like to use the agenda generator independently from triagebot.
This might be a good idea since there's only one active use of this feature: the compiler.

I never used this feature myself and I'm not familiar with the triagebot codebase, so I leave the details of this to the users of the agenda generator and the triage bot maintainers 👍

apiraino:
Sorry, is here a bit of AI-generated confusion? Do *people* have expressed this wish or was it just *me*?

This issue in isolation lacks context: there is an issue somewhere with some thoughts from you T-infra folks on why this is difficult and possibly not worth the effort.

~~@Kobzol I remember you wrote something about it (and I also replied!) but I don't remember anymore where.~~

Found it: https://github.com/rust-lang/infra-team/issues/266#issuecomment-4374857938

Kobzol:
I think that *people* here was meant more generally, as in "at least one Rust Project member" 😆 I think it's worth keeping as a separate issue, if you'd still like for this to happen.

marcoieni:
I used no AI to write this. Kobzol is right, I used "people" but instead I should have written "one person in the infra survey". I also DMd Urgau about this and he also liked the idea 👍

apiraino:
Thank you for granting me more weight than I deserve 😆 

I am [exploring](https://github.com/apiraino/triagebot/tree/embed-templates) the possibility to at least remove the constraint of running the agenda generator from the triagebot source repository.

Then I think I'll try duplicating all the code the agenda needs, then trim it down and try plugging APIs from the triagebot codebase. We'll see. Being the only user, I don't expect anyone to put resources into this, so don't worry too much about it 🙂

marcoieni:
> Thank you for granting me more weight than I deserve 😆

oh, it was you 😂 I didn't realize it when I read your reply previously early morning 😁

> so don't worry too much about it

Ok, thanks!

PR #2431, opened 2026-06-10
Add T-compiler P-high issues triage module

apiraino:
In an effort to make the triagebot more modular, I wanted to experiment with workspaces.

I need a new binary to run another meeting (the P-high issues review we do from time to time) and instead of adding code to the main triagebot codebase, I've tried creating a workspace ("utils", not a great name). Ideally I want to detach these binaries that now live under `./src/bin` and make them as independent as possible. Their main dependency i think is the GitHub client.

Build with: `cargo [ check | build ] --workspace`

Unsure if this is the right direction, so happy to get feedback!

r? @Urgau (for a first opinion, thanks!)

### DIFF
--- a/agendas/compiler-p-high/templates/p_high_issues.tt
+++ b/agendas/compiler-p-high/templates/p_high_issues.tt
@@ -1,0 +1,24 @@
+{% import "_issues.tt" as issues %}
+
+---
+title: T-compiler P-high issues
+tags: p-high-issues
+---
+
+Issues snapshot collected on: {{ CURRENT_DATE }}
+
+## Table of contents:
+- [P-high T-compiler issues unassigned](#P-high-T-compiler-issues-unassigned) ({{ p_high_tcompiler_unassigned | length }})
+- [P-high T-compiler issues assigned](#P-high-T-compiler-issues-assigned) ({{ p_high_tcompiler_assigned | length }})
+- [P-high without a team label](#P-high-issues-without-a-team-label) ({{ p_high_without_tlabel | length }})
+
+# 2026 Q2 T-compiler P-high Issue Review
+
+## P-high T-compiler issues unassigned
+{{-issues::render(issues=p_high_tcompiler_unassigned)}}
+
+## P-high T-compiler issues assigned
+{{-issues::render(issues=p_high_tcompiler_assigned)}}
+
+## P-high without a team label
+{{-issues::render(issues=p_high_without_tlabel)}}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -524,6 +524,32 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5b63caa9aa9397e2d9480a9b13673856c78d8ac123288526c37d7839f2a86990"
 
 [[package]]
+name = "compiler-p-high"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "async-trait",
+ "bytes",
+ "chrono",
+ "futures",
+ "http",
+ "http-body-util",
+ "itertools",
+ "octocrab",
+ "regex",
+ "reqwest",
+ "rust_team_data",
+ "secrecy",
+ "serde",
+ "serde_json",
+ "tera",
+ "tokio",
+ "tracing",
+ "tracing-subscriber",
+ "url",
+]
+
+[[package]]
 name = "comrak"
 version = "0.38.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,8 @@ authors = ["Mark Rousskov <mark.simulacrum@gmail.com>"]
 edition = "2024"
 
 [workspace]
+resolver = "3"
+members = ["agendas/compiler-p-high"]
 
 [dependencies]
 serde_json = "1"
--- a/agendas/compiler-p-high/Cargo.toml
+++ b/agendas/compiler-p-high/Cargo.toml
@@ -1,0 +1,30 @@
+[package]
+name = "compiler-p-high"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+anyhow = "1.0.98"
+async-trait = "0.1.31"
+chrono = { version = "0.4.38", default-features = false, features = ["serde"] }
+octocrab = { version = "0.44" }
+tera = { version = "1.3.1", default-features = false, features = ["builtins"] }
+tokio = { version = "1", default-features = false, features = ["macros", "time", "rt"] }
+tracing-subscriber = { version = "0.3", default-features = false, features = ["env-filter", "fmt"] }
+
+# Rust team data and its deps
+rust_team_data = { git = "https://github.com/rust-lang/team" }
+reqwest = { version = "0.12", features = ["json", "blocking"] }
+serde = { version = "1", default-features = false, features = ["derive"] }
+tracing = "0.1"
+
+# github.rs and its deps
+serde_json = "1"
+regex = "1"
+bytes = "1.10.1"
+itertools = "0.14.0"
+secrecy = { version = "0.10", default-features = false, features = ["serde"] }
+http = "1.4.0"
+http-body-util = "0.1.3"
+futures = { version = "0.3", default-features = false, features = ["std"] }
+url = "2.1.0"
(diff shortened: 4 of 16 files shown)


utils/compiler-p-high/src/main.rs
```diff
@@ -0,0 +1,148 @@
+use std::collections::HashMap;
+use std::sync::{Arc, LazyLock};
+
+use async_trait::async_trait;
+use tera::{Context, Tera};
+use triagebot::actions::{Action, Query, QueryKind, QueryMap};
+use triagebot::github::issue_query::Query as IssueQuery;
+use triagebot::github::{GithubClient, Repository};
+use triagebot::team_data::TeamClient;
```
apiraino: So far, these are the dependencies on the triagebot. I'm thinking to move some code out of the triagebot that is used only for these binaries (`triagebot::action::*`) and pull the rest (`triagebot::{github,team_data}`) as a triagebot dependency.

`team_data` is a crate so easily detachable from the triagebot (right?). `github` ideally could become an external library.

Anyway, just thoughts for now.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2431/a0f3af9a426e6015e3b0aa1aa685247fb6d7e7e6..eb133377e45c2612aa090d23e06f857711204c35)*

utils/compiler-p-high/src/main.rs
```diff
@@ -0,0 +1,148 @@
+use std::collections::HashMap;
+use std::sync::{Arc, LazyLock};
+
+use async_trait::async_trait;
```
apiraino, line 6: Do we need this crate in 2026? I mean, don't we have something in stdlib?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2431/a0f3af9a426e6015e3b0aa1aa685247fb6d7e7e6..eb133377e45c2612aa090d23e06f857711204c35)*

agendas/compiler-p-high/templates/p_high_issues.tt
apiraino, line 1: This is the only template I need here, I will cleanup the rest

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2431/a0f3af9a426e6015e3b0aa1aa685247fb6d7e7e6..eb133377e45c2612aa090d23e06f857711204c35)*

Urgau: I'm surprised we still have the T-lang templates. We don't have the generators anymore, we should remove them.

Urgau:
> Unsure if this is the right direction, so happy to get feedback!

I think that as long as the main code, that is `issue_query::Query` and the `actions` module are in triagebot, having a workspace doesn't bring much compared to having individual binaries in `bin/`.

I think we should got a step further and completely rip-out all the bits and pieces required used the agendas (for the github client, I would switch to the `octocrab` crate).

That way the agenda generators are completely independent, and we wouldn't even need workspaces.

apiraino:
> I think we should got a step further and completely rip-out all the bits and pieces required used the agendas (for the github client, I would switch to the `octocrab` crate).

Thanks! I also think this is the way to go (see [triagebot#2403](https://github.com/rust-lang/triagebot/issues/2403)). I'll keep on working on this path (as a workspace) to get a feel of the umbilical cord with the triagebot, then I'll see if I can cut it completely.

apiraino:
@Urgau when you have time, I think we can start reviewing this. You mentioned that you wanted some redundant code removed: do you still see anything I can pull out (I left some of the basic functions in the client intact).

Thanks

Urgau:
Th extracted parts looks fine, though I'm wondering what the next steps would be? Only using octocrab (ie. removing our own GitHub client)? Porting the other agendas?

I'm also wondering if we even need a workspace member since the new logic doesn't link directly to triagebot, we could also better name the folder, maybe `agendas/`?

> do you still see anything I can pull out

I think `Repository::submodule`, `PullRequestDetails::compare` (and all the related `GitHubCommit`, `FileDiff` structs and such).

apiraino:
I *think* we cannot just yet use `octocrab`, there are 330 SLOC in `./utils/compiler-p-high/src/github/client.rs` that I think cannot be replaced by the crate. Ideally an outcome of this experimentation could be trying to upstream some of our stuff into `octocrab` (if it's not too custom and ofc if the maintainer agrees :wink: ).

The GH client for the compiler triage meeting (`prioritization_agenda.rs`) is likely be much bigger, there is all the GraphQL code.

The workspace was meant to dramatically shave off compile time, which is one of my annoyances when touching these agenda generators: but I still don't have numbers to support this claim.

(agree on the file naming, thanks for the suggestion)

Thanks for your comments, it was really helpful. I'll circle back to this hopefully in the near future. Now I have to park this for a bit.

Urgau:
Given that `octocrab` [allow custom HTTP requests](https://docs.rs/octocrab/latest/octocrab/#http-api), I think we could entirely remove our own custom GitHub client (for agendas and the current triagebot).

(The only thing we would be losing if we only used `octocrab` client is the automatic retry of requests, but we don't use that currently so.)

Same for the GraphQl queries, we could use the [`graphql`](https://docs.rs/octocrab/latest/octocrab/struct.Octocrab.html#method.graphql) method with our custom queries.

Anyway, those are just some things I noticed while quickly looking at the documentation.

apiraino:
I'm back and looked again at the previous [review comment](https://github.com/rust-lang/triagebot/pull/2431#issuecomment-4886222607). I've renamed the workspace and removed some more redundant code.

I would like to work on the other things left (like removing the internal GH client) in a future patch.

Urgau:
Happy with this experiment and have further PRs refine it.

Could you create a issue with a list of tasks to do after this PR?

cc @Kobzol @Mark-Simulacrum in case you have an objection

Kobzol:
Soo, in general using a workspace or splitting triagebot to be more of a library and then having a bunch of related binaries using it sounds fine to me.

But in the current state, this PR just adds new code that doesn't use triagebot at all, right? And it just copy pastes some of triagebot's code into the new binary.

So at that point, this could just live entirely separately, there is no need for this to be under the triagebot repo, if it doesn't use any of its code (?).

apiraino:
> So at that point, this could just live entirely separately, there is no need for this to be under the triagebot repo, if it doesn't use any of its code (?).

that's correct, It's a reduced copy-paste of the triagebot code and this code has exactly 1 user (me :smile: ).

If you're hinting at moving this code under another `rust-lang/` It's totally fine for me[^1]. Though if tomorrow I want to move also the other agenda generator (which has more code in common with the triagebot), that would be a bit more heftier move.

[^1]: `rust-lang/compiler-team`? I should probably ask

Urgau:
> So at that point, this could just live entirely separately, there is no need for this to be under the triagebot repo, if it doesn't use any of its code (?).

I had the same though of mind, and think we should eventually do that. Whenever it should happen sooner or later, I'm fine with both.

Kobzol:
I think that either of the following makes sense:
- Create this as a separate code in a separate repository. It doesn't even need to live under `rust-lang/rusŧ` (though it could), if it's essentially only for you at the moment, you can just have it under your account.
- Add it to triagebot, but in that case it should use `triagebot` as a library, and not copy-paste its files into a separate workspace (it could also be just a separate binary, if it shares most dependencies anyway, then that shouldn't make triagebot's compilation much slower). But this would likely require some refactoring, both of the agenda generator, and of triagebot, to export the required functionality.

apiraino:
> I think that either of the following makes sense:
> 
>     * Create this as a separate code in a separate repository.
> 
>     * Add it to triagebot, but in that case it should use `triagebot` as a library

OK, then I'll choose the option to host in my GH account because it's easier.

I'm a bit sad about that, in the end what I wanted to do was refactoring and "upstreaming" this tool based on an idea from Jie You (https://github.com/jieyouxu/pcr-util/) for the benefit of the project, whether I am the one using it today or someone else tomorrow when I get hit by a bus. Maybe I'll just mention this tool it in the T-ops documentation :shrug: 

anyway, thanks for the review.

(I will still remove the dead code from the triagebot, it was mentioned with Urgau via Zulip DM)

Kobzol:
Upstreaming that tool sounds good, but in that case we should start with the refactoring and integration into triagebot, rather than dumping 1.5k new lines of code into the repo, with the promise that it will be integrated later - because often these things stay as they are for a long time. So I'm happy to accept the new code, if it uses existing triagebot machinery, and doesn't copy-paste its files.

PR #2432, opened 2026-06-10
Make the reporting instructions less distracting when reporting an error

Urgau:
Currently:
image

This PR:
image

With this change I feel like it's easier to see the error, and not be distracted by the reporting instructions.


Kobzol (approved):
Makes sense :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2432/a0f3af9a426e6015e3b0aa1aa685247fb6d7e7e6..c895c1a2265097d67ff75118371359dd8159cd24)*

PR #2436, opened 2026-06-15
Add `delegate` command to allow others to merge PRs

Urgau:
This PR adds the `delegate` command to allow other users to merge PRs using the `merge` command for merge-queues.

Follow-up to https://github.com/rust-lang/triagebot/pull/2412
cc @samueltardieu

### DIFF
--- a/parser/src/command/merge.rs
+++ b/parser/src/command/merge.rs
@@ -2,12 +2,24 @@ use crate::error::Error;
 use crate::token::{Token, Tokenizer};
 
 #[derive(PartialEq, Eq, Debug)]
-pub struct MergeCommand;
+pub enum MergeCommand {
+    Merge,
+    DelegateToAuthor,
+    Delegate { login: String },
+}
 
 impl MergeCommand {
     pub fn parse<'a>(input: &mut Tokenizer<'a>) -> Result<Option<Self>, Error<'a>> {
         if let Some(Token::Word("merge")) = input.peek_token()? {
-            Ok(Some(Self))
+            Ok(Some(MergeCommand::Merge))
+        } else if let Some(Token::Word("delegate" | "delegate+")) = input.peek_token()? {
+            Ok(Some(MergeCommand::DelegateToAuthor))
+        } else if let Some(Token::Word(word)) = input.peek_token()?
+            && let Some(login) = word.strip_prefix("delegate=")
+        {
+            Ok(Some(MergeCommand::Delegate {
+                login: login.to_string(),
+            }))
         } else {
             Ok(None)
         }
--- a/src/handlers/merge.rs
+++ b/src/handlers/merge.rs
@@ -3,27 +3,40 @@ use parser::command::merge::MergeCommand;
 
 use crate::{
     config::MergeConfig,
+    db::issue_data::IssueData,
     errors::user_error,
-    github::{Event, client::GraphQlErrors},
+    github::{Event, IssueCommentEvent, UserId as GitHubUserId, client::GraphQlErrors},
     handlers::Context,
 };
+
+/// Key for the state in the database
+const DELEGATIONS_KEY: &str = "merge-delegations";
+
+/// State stored in the database
+#[derive(Debug, Default, serde::Deserialize, serde::Serialize, Clone, PartialEq)]
+struct DelegationsState {
+    // List of all the user IDs who have been delegated the power to merge the PR.
+    delegations: Vec<GitHubUserId>,
+}
 
 pub(super) async fn handle_command(
     ctx: &Context,
     config: &MergeConfig,
     event: &Event,
-    _cmd: MergeCommand,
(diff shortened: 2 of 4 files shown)


samueltardieu:
Thanks for doing this!

The way I would use it is, do a review with "request changes" and, in the message, use "@rustbot delegate+".

However, will "@rustbot merge" be successful if the review is still in the "request changes" state? Right now, when a user with write permission has requested changes, it blocks the merge.

Should triagebot remember *who* (possibly plural) has delegated the right to merge, and invalidate "request changes" reviews from those people before attempting the merge?

Urgau:
I don't think triagebot should mess-up the approvals on the PR.

I would say that the workflow should be to use an "Approval" review when you `delegate+`: that (conceptually) says "approved modulo the changes asked", which is consistent with the message displayed to the user:

> :v: @Urgau, you can now merge this pull request!
>
> If @samueltardieu told you to merge after making some further change, then please make that change and post `@rustbot merge`.
> 
> [View changes since this delegation](https://triagebot.infra.rust-lang.org/gh-changes-since/{org_repo}/{pr_num}/{base_sha}..{head_sha}).

--------

Also, looking at the configured [team repo](https://github.com/rust-lang/team/blob/d0f85d9db7eec5a7fc0996775e70f54fb0f80d85/repos/rust-lang/rust-clippy.toml#L19), it seems like it's currently setup for 0 approvals, so it might not matter.

samueltardieu:
Makes sense.

parser/src/command/merge.rs
```diff
@@ -2,12 +2,24 @@ use crate::error::Error;
...
+    DelegateToAuthor,
+    Delegate { login: String },
+}
 
 impl MergeCommand {
     pub fn parse<'a>(input: &mut Tokenizer<'a>) -> Result<Option<Self>, Error<'a>> {
         if let Some(Token::Word("merge")) = input.peek_token()? {
-            Ok(Some(Self))
+            Ok(Some(MergeCommand::Merge))
+        } else if let Some(Token::Word("delegate+")) = input.peek_token()? {
```
Kobzol: I always found the `+` confusing. Isn't `delegate` enough? Or maybe we can support both.

src/handlers/merge.rs
```diff
@@ -62,3 +121,63 @@ pub(super) async fn handle_command(
...
+    issue_comment: &IssueCommentEvent,
+    delegate_to: &str,
+) -> anyhow::Result<()> {
+    let delegatee = ctx.github.user_info(delegate_to).await?;
+
+    let mut db = ctx.db.get().await;
+    let mut state: IssueData<'_, DelegationsState> =
+        IssueData::load(&mut db, &issue_comment.issue, DELEGATIONS_KEY).await?;
+
+    state.data.delegations.push(delegatee.id);
```
Kobzol, line 143: We should make it a set or check if the ID wasn't already there.

src/handlers/merge.rs
```diff
@@ -62,3 +121,63 @@ pub(super) async fn handle_command(
...
+    let state: IssueData<'_, DelegationsState> =
+        IssueData::load(&mut db, &issue_comment.issue, DELEGATIONS_KEY).await?;
+
+    for delegated_user_id in &state.data.delegations {
+        if delegated_user_id == &issue_comment.comment.user.id {
+            return Ok(true);
+        }
+    }
+
+    // Manually drop the state to release the DB lock
```
Kobzol: The variable will be dropped at the end of the function just below, right?

Kobzol (commented):
Thanks, looks good, left a couple of comments.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2436/21e60800066f4cfcbffe1cbc4cee883e9fdea85b..da1d42e9534a71c66a16cc027e210beb6b6e23e2)*

parser/src/command/merge.rs
Urgau: Sure, added support for `delegate` without the trailing `+`

src/handlers/merge.rs
Urgau: Initially the function was doing more stuff. Removed the manual drop.

src/handlers/merge.rs
Urgau: I don't think it matters much if we are adding a user twice, but added a check anyway.

Kobzol: Well, mostly as a sort of a DoS protection, to avoid someone spamming our DB.

Kobzol (approved):
Looks good to me!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2436/07724d7149e8bdae18e08c1c7bc00869c2107578..19f8df028ee56bdac437d611707b6358ddda8473)*

PR #2440, opened 2026-06-19
Add warning against Unicode Bidi characters in our range-diff

Urgau:
Like GitHub, let's show a warning if we detect bidirectional characters in the patch.

| Dark | Light |
|-----|----|
| image | image |

Links:
 - https://github.co/hiddenchars
 - https://github.com/rust-lang/rust/commit/c0b134582a949062f9f3a8ba3def88b98a98df6a
 - http://localhost:8000/gh-range-diff/rust-lang/rust/38b01d9..38b01d9/38b01d9..c0b134582a949062f9f3a8ba3def88b98a98df6a

Suggested by @camelid

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3448,6 +3448,7 @@ dependencies = [
  "hyper",
  "ignore",
  "itertools",
+ "memchr",
  "native-tls",
  "octocrab",
  "parser",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ globset = { version = "0.4.18", default-features = false }
 tower_governor = { version = "0.8.0", default-features = false, features = ["axum", "tracing"] }
 http-body-util = "0.1.3"
 http = "1.4.0"
+memchr = "2.7.5"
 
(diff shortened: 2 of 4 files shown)


Kobzol (approved):
Not sure if I'm a fan of adding a new dependency (and more code) for something so niche, tbh. But I won't block it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2440/9099bc419364d6a180022d6b7d6fa21d0d2f8ac4..ed57738472d4f6b34bf3620ea837b8b2ea84fe2a)*

Urgau:
It could be done without the dependency on `memchr`, but since we already have that dependency 18 times in our `Cargo.lock`, I think it's fine.

As for adding something this "niche", I would agree at first, but since GitHub decided to add such warning, that we had a `rustc` CVE for it (CVE-2021-42574), and that we added a deny-by-default lint for it (`text_direction_codepoint_in_literal`). I don't think it's that niche.

camelid:
Thanks @Urgau!

PR #2443, opened 2026-06-22
Block aggressive User-Agents

Urgau:
Context: [#triagebot > graphql rate-limit exhausted](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/graphql.20rate-limit.20exhausted/with/605784431)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,7 +4,8 @@ use anyhow::Context as _;
 use axum::body::Body;
 use axum::error_handling::HandleErrorLayer;
 use axum::http::HeaderName;
-use axum::response::{Html, Response};
+use axum::middleware::{self, Next};
+use axum::response::{Html, IntoResponse, Response};
 use axum::routing::{get, post};
 use axum::{BoxError, Router};
 use hyper::{Request, StatusCode};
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -292,6 +296,20 @@ async fn run_server(addr: SocketAddr) -> anyhow::Result<()> {
...
 }
 
+/// Check if the request is coming from one of our banned UA, reject the request if that's the case
+async fn block_user_agents(req: Request<axum::body::Body>, next: Next) -> Response {
+    if let Some(user_agent_value) = req.headers().get(hyper::header::USER_AGENT)
+        && let Ok(ua) = user_agent_value.to_str()
+    {
+        if BANNED_USER_AGENTS.iter().any(|banned| ua.contains(banned)) {
+            // Reject immediately with a 403 Forbidden status
+            return (StatusCode::FORBIDDEN, "Access Denied: Banned User-Agent").into_response();
```
apiraino: this approach, if extended to other UAs (esp. if we add regexps) will be subject to false positives. Would it make sense tuning the msg more or less like this?
```suggestion
            return (StatusCode::FORBIDDEN, "Access Denied: Banned User-Agent. If you think this is a mistake, please open an issue at github.com/rust-lang/triagebot/issues").into_response();
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2443/2cc28c654a4c3913a409ccc36542fa995f07b64b..12e7d841e1925319b2aa71569650d720daa855eb)*

Mark-Simulacrum (approved):

PR #2445, opened 2026-06-23
Block our expensive endpoints if our GitHub rate limits are too low

Urgau:
Following up [#triagebot > graphql rate-limit exhausted](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/graphql.20rate-limit.20exhausted/with/605988237) and https://github.com/rust-lang/triagebot/pull/2443.

Let's block our exposed (and expensive) endpoints if our GitHub rate limits are too low (< 750 for either core or graphql), in order to keep our core functionalities (webhooks, github commands, ...) working.

Fetching GitHub [`/rate_limit`](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2026-03-10#checking-the-status-of-your-rate-limit):
> does not count against your primary rate limit, but it can count against your secondary rate limit.

(If we are worried about the secondary rate limits, we could try to cache the headers, but that's much more complicated.)

cc @Mark-Simulacrum

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,6 +3,7 @@
 use anyhow::Context as _;
 use axum::body::Body;
 use axum::error_handling::HandleErrorLayer;
+use axum::extract::State;
 use axum::http::HeaderName;
 use axum::middleware::{self, Next};
 use axum::response::{Html, IntoResponse, Response};
@@ -237,7 +238,11 @@ async fn run_server(addr: SocketAddr) -> anyhow::Result<()> {
             "/gh-comments/{owner}/{repo}/pull/{pr}",
             get(triagebot::gh_comments::gh_comments),
         )
-        .layer(GovernorLayer::new(ratelimit_config));
+        .layer(GovernorLayer::new(ratelimit_config))
+        .layer(middleware::from_fn_with_state(
+            ctx.clone(),
+            block_if_github_ratelimit_too_low,
+        ));
 
     let app = Router::new()
         .route("/", get(|| async { "Triagebot is awaiting triage." }))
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -310,6 +315,31 @@ async fn block_user_agents(req: Request<axum::body::Body>, next: Next) -> Respon
...
+            StatusCode::SERVICE_UNAVAILABLE,
+            "Could not read GitHub rate limit",
+        )
+            .into_response();
+    };
+
+    if rl.graphql.remaining < 750 || rl.core.remaining < 750 {
+        return (
+            StatusCode::TOO_MANY_REQUESTS,
+            "Low GitHub rate limit. Core functionnality preserved; try again later.",
```
Mark-Simulacrum:
```suggestion
            "Low GitHub rate limit. Core functionality preserved; try again later.",
```

Mark-Simulacrum (approved):
I think the more general solution here would be to use two different tokens for these modes of operation... but that would take some infra work (e.g., spinning up separate GitHub apps). It does seem like a good idea long term though.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2445/6f2b514931a99f305cf2eeb22ad49651861b0133..c75092e65bbea4f4140b87fcf98029bb12841955)*

Urgau:
Yeah, using a second (or multiple) GitHub app(s) would fundamentally separate those issues, but has you said it would take some infra work.

So for now, let's rate limit before GitHub rate limit us.

PR #2446, opened 2026-06-24
Tell the user that nothing changed rather than a cryptic GitHub message

Urgau:
Instead of redirecting the user to GitHub's compare view when nothing changed in our "View changes this review" link, show a clear message, instead of the cryptic GitHub message.

| GitHub compare | This PR |
|---------|--------|
| image| image |

cc @camelid

### DIFF
--- a/src/gh_changes_since.rs
+++ b/src/gh_changes_since.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 use anyhow::Context as _;
 use axum::{
     extract::{Path, State},
-    response::{IntoResponse, Redirect, Response},
+    response::{Html, IntoResponse, Redirect, Response},
 };
 use hyper::StatusCode;
 
@@ -46,6 +46,20 @@ pub async fn gh_changes_since(
 
     let newbase = &pr.base.as_ref().context("no base")?.sha;
     let newhead = &pr.head.as_ref().context("no head")?.sha;
+
+    // Did something changed, no?
+    if oldhead == newhead {
+        // Inform the user that nothing changed.
+        return Ok((StatusCode::OK, Html(format!(
+r##"<!DOCTYPE html>
+<title>Changes since for {owner}/{repo}#{pr_num}</title>
+<style>
+  :root {{ color-scheme: light dark; font-family: system-ui; margin: 2rem auto; max-width: 60ch; }}
+</style>
+<h1>No changes since this review.</h1>
+<p>Head back to the <a href="https://github.com/{owner}/{repo}/pull/{pr_num}/changes">latest changes here</a>.</p>
+"##))).into_response());
+    }
 
     // Has the base changed?
     if oldbase == newbase {


src/gh_changes_since.rs
```diff
@@ -3,12 +3,22 @@ use std::sync::Arc;
...
     extract::{Path, State},
-    response::{IntoResponse, Redirect, Response},
+    response::{Html, IntoResponse, Redirect, Response},
 };
 use hyper::StatusCode;
 
 use crate::{errors::AppError, github, handlers::Context, utils::is_known_and_public_repo};
 
+const NO_CHANGES_HTML: &str = "
+<!DOCTYPE html>
```
camelid: Should probably emit complete HTML, and in particular a meta tag specifying the utf8 charset so the "—" works properly.

Urgau: I though so too, but surprisingly the [official W3 validator](https://validator.w3.org/#validate_by_input) doesn't have a issue with it, nor does Firefox.

Urgau: As for UTF-8, it's in the response header: `content-type: text/html; charset=utf-8`

Urgau: > surprisingly the [official W3 validator](https://validator.w3.org/#validate_by_input) doesn't have a issue with it

That's because it's valid [TM](https://html.spec.whatwg.org/multipage/syntax.html#optional-tags) to omit those.

camelid: Huh fair enough!

src/gh_changes_since.rs
```diff
@@ -3,12 +3,22 @@ use std::sync::Arc;
...
+<title>Changes since</title>
+<style>
+  :root { color-scheme: light dark; font-family: system-ui; margin: 2rem auto; max-width: 60ch; }
+</style>
+<h1>No changes since this review.</h1>
+<p>—your humble servant rustbot</p>
```
apiraino: I won't stop opposing to these cringe comments in our codebase and politely ask to remove them when I see them.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2446/98234128f581bb076a373c18706246ff8fbdb0b7..fec02093773f8e52dee531b31e59e3fe562dee93)*

Urgau: It's supposed to be joke, to lighten a bit the page, to reduce a bit the *unnecessary* strictness of the page.

Anyway, removed it.

Kobzol (approved):
As a person who routinely switches between 5 different PRs and continuously opens tabs in the background, I'd appreciate if the page contained a link to the original PR, so that I can find what PR is the page referring to 😆

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2446/98234128f581bb076a373c18706246ff8fbdb0b7..9268ba7d8f629ef929155adf7e918ad975288304)*

Urgau:
> I'd appreciate if the page contained a link to the original PR

Added, in the same as GitHub: "Head back to the [latest changes here](https://github.com/rust-lang/rust/pull/158318/changes)."

PR #2447, opened 2026-06-30
Pin GitHub API version in our GitHub client

apiraino:
The current GitHub API version ("2022-11-28") we use will be deprecated in about 2 years from now, so at some point we should start testing the new version.
Our client is just a part of the story because we heavily use the `octocrab` crate (I don't know what is the situation there).
I thought about adding a check and a "reminder".

GitHub blog post: https://docs.github.com/en/rest/about-the-rest-api/breaking-changes?apiVersion=2026-03-10#version-2026-03-10

Zulip discussion: https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/New.20GitHub.20API.20version/near/579038018

What do you think? Can this be helpful?

r? @Kobzol / @Mark-Simulacrum 

thanks

### DIFF
--- a/src/github/client.rs
+++ b/src/github/client.rs
@@ -14,6 +14,12 @@ use bytes::Bytes;
 
 use crate::jobs::Job;
 
+// TODO: Update to "2026-03-10" and see what breaks
+// current version 2022-11-28 (supported until March 2028)
+// Note: If you specify an API version that is no longer supported, you will receive a 410 Gone response.
+// see: https://docs.github.com/rest/about-the-rest-api/api-versions?apiVersion=2026-03-10
+const GITHUB_API_VERSION: &str = "2022-11-28";
+
 /// Finds the token in the user's environment, panicking if no suitable token
 /// can be found.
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (approved):
Seems not unreasonable, though I guess bumping to latest and seeing what happens is probably the best we can do (at least for triagebot).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2447/58c2658e1a911cc59aef44301216ef6146035316..efb2e0905601096f1c1c9892e8ac2dbba505a09c)*

Kobzol (commented):
Makes sense, thank you. If we forget, this will at least clearly notify us that something is wrong :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2447/58c2658e1a911cc59aef44301216ef6146035316..efb2e0905601096f1c1c9892e8ac2dbba505a09c)*

PR #2449, opened 2026-07-01
Display our own message when CI checks are not green for enqueuing PR

Urgau:
This PR adds a small logic to detect and display our own message when CI checks are not green when trying to enqueue PR (`rustbot merge` command).

> All mandatory CI checks must be green before this PR can enter the merge queue.
>
> Please wait for all checks to pass, and if necessary resolve any failures, then retry the command.

Unfortunately GitHub doesn't give us any code or information about the failure it-self other than the error message it-self, so we have to do some matching on them, let's hope those error messages don't change too often.

cc @samueltardieu

### DIFF
--- a/src/handlers/merge.rs
+++ b/src/handlers/merge.rs
@@ -111,6 +111,19 @@ async fn merge_pr(
                     && error.path == &["enqueuePullRequest"]
                     && error.type_ == "UNPROCESSABLE"
                 {
+                    // Provide our own message in some specific cases, as the one given by GitHub
+                    // can be too technical and lead to confusion.
+                    if error
+                        .message
+                        .ends_with("required status checks are expected.")
+                    {
+                        // GitHub message: Pull request 3 of 4 required status checks are expected.
+                        // https://github.com/rust-lang/rust-clippy/pull/17330#issuecomment-4859026904
+                        return user_error!("All mandatory CI checks must be green before this PR can enter the merge queue.
+
+Please wait for all checks to pass, and if necessary resolve any failures, then retry the command.".to_string());
+                    }
+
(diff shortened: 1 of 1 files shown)


Kobzol:
Shouldn't it be possible to use "Merge when ready" using the GraphQL call? That sounds much more practical than displaying an error message and forcing the user to wait.

Urgau:
Seems like it would be possible with `enablePullRequestAutoMerge` ("Enable the default auto-merge on a pull request"), but do we want to do that? Is auto-merge status reset when a new push has been made?

Kobzol:
I don't think so, I remember that I sometimes enable it on bors and then push a few times until CI passes. That might be preventable by requiring a review of the most recently pushed commit though. It just seems nicer to do what the UI could already do for you, rather than forcing the user to retry, which is an annoyance (same in bors, where we would someday like to automate the "r=me once CI is green" workflow).

Urgau:
I'm fine changing it, if @samueltardieu agrees.

samueltardieu:
Do we get a webhook when CI fails? Would it be possible to disable auto-merge then? That could be good in general, both in the context of rustbot merge and in the context of a manual auto-merge.

Urgau:
We probably get one, but I don't think that's necessary, as GitHub disable auto-merge when new changes are pushed by a non-member.

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request#about-auto-merge

Urgau:
Continuing in https://github.com/rust-lang/triagebot/pull/2450

PR #2450, opened 2026-07-02
Use auto merge unless ready to be enqueued in the merge queue

Urgau:
As discussed in https://github.com/rust-lang/triagebot/pull/2449#issuecomment-4860165623, let's use the auto merge functionality when the PR is not yet ready to be enqueued in the merge queue (PR CI still running, ...).

That way PRs that are not yet ready can still be `@rustbot merge`-ed.

I tested the logic in my fork and it seems to work as expected.

<details>

image

</details>

cc @samueltardieu

### DIFF
--- a/src/handlers/merge.rs
+++ b/src/handlers/merge.rs
@@ -1,4 +1,5 @@
 use anyhow::Context as _;
+use octocrab::models::pulls::MergeableState;
 use parser::command::merge::MergeCommand;
 
 use crate::{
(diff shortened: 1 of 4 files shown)


src/handlers/merge.rs
```diff
@@ -101,21 +101,59 @@ async fn merge_pr(
...
+    tracing::info!(
+        "pr mergeable status={:?} ({:?})",
+        &pr.mergeable,
+        &pr.mergeable_state
+    );
+
+    // Unless the mergeable state is "clean" we enable auto merge, if it's "clean"
+    // we can't use auto merge (GitHub blocks it), so let's enqueue the PR in the
+    // merge queue, since we know by the "clean" state that we can.
+    if pr.mergeable_state.as_deref() != Some("clean") {
```
Kobzol: non-blocking nit: we could use https://docs.rs/octocrab/latest/octocrab/models/pulls/enum.MergeableState.html (or our local copy with a string enum variant fallback) to avoid the string comparison.

Kobzol (approved):
The mergeable statuses always confuse me :) Not sure if HasHooks is also something that would work for directly enqueuing? We'll see in practice, I guess.

I wonder what happens if you enable auto-merge when the PR is already mergeable? Does it also work? I'd hope it would.. :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2450/c77b369fe5d07eec2674c67782eb0676a62f4d04..20982ec137e7a09416db7a027845ee785d132ba2)*

Urgau:
> I wonder what happens if you enable auto-merge when the PR is already mergeable?

Depends on what you mean by mergeable, if you mean no conflicts, it works, but if you mean CI is green, then no, it gets rejected by GitHub as being already ready (the "clean" state"), and as such we need to enqueue it in the merge queue instead.

Kobzol:
If we could reliably detect that, then we could always try to auto merge, and if that fails, then enqueue. But the current solution of this PR should also work, so fine by me.

Urgau:
That was my first instinct as well, but GitHub doesn't give us nice error code, only an error message: "Pull request Pull request is in clean status" *(yes, there really is a double "Pull request" at the start of the message, not a typo from me)*.

If you're fine doing some error message matching, we can switch to auto-merge first and if rejected for "clean" then enqueue.

src/handlers/merge.rs
Urgau: Done, switch-ed to the `octocrab` enum.

Urgau:
Let's go with the `mergeable_state` check (feels cleaner to me).

PR #2452, opened 2026-07-06
Only try once to add the view changes since link

Urgau:
Only try to add the view changes since link once to avoid retrying multiple times for comments that may have been deleted, or review comments we can't modify, ...

### DIFF
--- a/src/handlers/review_changes_since.rs
+++ b/src/handlers/review_changes_since.rs
@@ -187,35 +187,44 @@ impl crate::jobs::Job for AddReviewChangesSinceLinkJob {
     }
 
     async fn run(&self, ctx: &Context, metadata: &serde_json::Value) -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Were the jobs repeated many times? 🥶 Good catch!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2452/8fe35ac13efb881598a4f1d7768d5dee5a3fff09..f5104916035c269608c20d98f4631732e580fa34)*

Urgau:
> Were the jobs repeated many times?

I don't know, I think they repeated indefinitely, but it's hard to tell.

https://github.com/rust-lang/triagebot/blob/8fe35ac13efb881598a4f1d7768d5dee5a3fff09/src/db/jobs.rs#L95-L97

PR #2456, opened 2026-07-08
Add support for `@bot claim @user` to assign to another user

### LINKED ISSUE #2451: `@rustbot claim @user` should assign to @user
I accidentally assigned myself rather than the person working on the ticket
https://github.com/rust-lang/rust/issues/158800#issuecomment-4891752108

I know now that the correct syntax is `@rustbot assign @user`.
But I think attempting to "claim" a ticket for someone else should fail.

Zulip thread: https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/.60rustbot.20claim.20.40user.60.20unexpected.20behaviour/near/608414614

DanielEScherzer:
Fixes #2451

### DIFF
--- a/parser/src/command/assign.rs
+++ b/parser/src/command/assign.rs
@@ -154,6 +154,11 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_5() {
+        assert_eq!(parse("claim @user"), Ok(Some(AssignCommand::Claim)),);
+    }
+
     fn parse_review<'a>(input: &'a str) -> Result<Option<AssignCommand>, Error<'a>> {
         let mut toks = Tokenizer::new(input);
         Ok(AssignCommand::parse_review(&mut toks)?)
--- a/parser/src/command/assign.rs
+++ b/parser/src/command/assign.rs
@@ -159,6 +159,14 @@ mod tests {
         assert_eq!(parse("claim @user"), Ok(Some(AssignCommand::Claim)),);
     }
 
+    #[test]
+    fn test_6() {
+        assert_eq!(
+            parse("claim does not start with @"),
+            Ok(Some(AssignCommand::Claim)),
+        );
+    }
+
     fn parse_review<'a>(input: &'a str) -> Result<Option<AssignCommand>, Error<'a>> {
         let mut toks = Tokenizer::new(input);
         Ok(AssignCommand::parse_review(&mut toks)?)
--- a/parser/src/command/assign.rs
+++ b/parser/src/command/assign.rs
@@ -5,7 +5,8 @@
 //! The grammar is as follows:
 //!
 //! ```text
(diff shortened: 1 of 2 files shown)


parser/src/command/assign.rs
```diff
@@ -51,10 +52,22 @@ impl fmt::Display for ParseError {
 }
 
 impl AssignCommand {
-    pub fn parse<'a>(input: &mut Tokenizer<'a>) -> Result<Option<Self>, Error<'a>> {
+    pub fn parse<'a>(
+        bots: &Vec<&'a str>,
```
Kobzol: Why do you filter out bots? I don't think we do that elsewhere, or do we?

Kobzol (commented):
Hi, thanks for the PR. Left one question.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2456/6368d607b5d1520f2c6d6b5ee667d133de766680..f8fded97cb5953f71af7f0d54c7ac752b249588e)*

parser/src/command/assign.rs
Urgau: I'm not aware either of any other place where we stop parsing at a bot handle. Generally, when we want to stop parsing we look for a punctuation like `.`.

DanielEScherzer: I wanted to ensure commands like `@rustbot claim @rustbot label +bug` continued to be parsed as two different commands rather than assigning the bot itself

Kobzol: That's not consistent with what we do for other commands here, I think. bors does stop at `@` to let other bots handle the rest of the line, but triagebot AFAIK doesn't, and to add insult to injury, it actually requires `@` to be present before usernames for the assign (and now claim) commands, which is IMO unnecessary.

Please remove the bot handling, we don't do it elsewhere in triagebot.

DanielEScherzer: Okay, removed - it turns out that the bot still handles the labeling after the assignment, see the changes in `test_follow_commands()`

Kobzol (approved):
Looks fine to me, I'll leave @Urgau to take a look if you want.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2456/3c11f50e61f1b315567400690e6eb4cb86767a27..277cbca215a786bddd3b3a59a0f03e28be712787)*

Urgau (approved):
Looks good to me as well. Thanks.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2456/3c11f50e61f1b315567400690e6eb4cb86767a27..277cbca215a786bddd3b3a59a0f03e28be712787)*

PR #2457, opened 2026-07-09
Allow glob matches for autolabels and rendered links

tgross35:
Use the same logic that currently exists for mentions to match `autolabel.trigger-files` and `render-links.{trigger-files, exclude-files}`.

The first commit just moves things around, the rest is less noisy.


src/handlers/mentions.rs
```diff
@@ -86,21 +86,28 @@ pub(super) async fn parse_input(
...
-                    modified_files
+                    let it = modified_files
                         .iter()
                         .filter(|f| patch_adds(&f.patch, entry))
-                        .map(|f| PathBuf::from(&f.filename))
-                        .collect()
+                        .map(|f| Path::new(&f.filename));
+                    Box::from(it)
                 }
             };
```
tgross35, line 108: Using iterators to avoid the path copy + collect might be overkill, happy to drop if preferred

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2457/6368d607b5d1520f2c6d6b5ee667d133de766680..42b7fe2a2555e1570020a2fad229f2064ec2028e)*

Urgau: I'm all for those kind of optimizations, but I'm not sure this one even gains anything, I think it's even detrimental since it does one more allocation than before, as the `collect` should only allocate when there are any elements. The `Box<dyn Iterator<Item = &Path>>` being the unnecessary allocation.

In other words, collecting an empty `Vec` shouldn't cost any allocation, and since want all or nothing we should only have one allocation.

Anyway, if don't mind dropping it, that would be appreciated.

tgross35: Oh whoops, you're right, we weren't paying anything if not returning, except in the case that `has_relevant_ccs` fails. Dropped this now

src/utils.rs
```diff
@@ -81,3 +82,399 @@ pub(crate) fn immutable_headers(content_type: &'static str) -> HeaderMap {
...
+
+            // Create the glob pattern and log an error (should have already been reported to
+            // the user).
+            let glob = match globset::GlobBuilder::new(&pattern)
+                .empty_alternates(true)
+                .build()
+            {
+                Ok(pattern) => pattern,
+                Err(err) => {
+                    tracing::error!("invalid glob pattern for [mentions.\"{entry}\"]: {err}");
```
Urgau:
```suggestion
                    tracing::error!("invalid glob pattern "{entry}": {err}");
```

src/utils.rs
```diff
@@ -81,3 +82,399 @@ pub(crate) fn immutable_headers(content_type: &'static str) -> HeaderMap {
...
+
+            // Prepare the glob pattern from the entry.
+            //
+            // We first trim any excess `/` at the end of the pattern and then add an alternate
+            // to match on the path as is or in any sub-directories.
+            let pattern = entry.trim_end_matches('/');
+            let pattern = format!("{pattern}{{,/*}}");
+
+            // Create the glob pattern and log an error (should have already been reported to
+            // the user).
```
Urgau, line 117: Since this is adding glob patterns to other places, would you mind adding some validation for users in `validate_config.rs`, similar to this one: https://github.com/rust-lang/triagebot/blob/6368d607b5d1520f2c6d6b5ee667d133de766680/src/handlers/check_commits/validate_config.rs#L71-L90

If you don't want to do it that's fine, just tell me.

Urgau (commented):
Looks good to me. Left two nits.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2457/6368d607b5d1520f2c6d6b5ee667d133de766680..80ddcb9e8d5c663b0e34bb9dbcfebbdfc68a5ad5)*

src/utils.rs
tgross35: Changed in the latest push

src/utils.rs
tgross35: Ah I didn't know this existed, added a commit changing it. Added a commit changing this though I'm not quite sure how to test. (Feel free to just push if there's something I'm missing)

Urgau: Testing requires a setting up triagebot on a repo and sending a PR changing `traigebot.toml`. I will test it for you.

Urgau: Works as expected!

Mentions:
image

Autolabel:
image

Rendered link:
image

<details>

```toml
[autolabel."test"]
trigger_files = [
    "inva{id"
]

[rendered-link]
trigger-files = ["inva{id"]
exclude-files = ["inva{iid"]

[mentions."inval{id"]
cc = ["Urgau"]
```

</details>

*The only sub-optimal thing is that we only show one error at a time, but that's pre-existing.*

src/utils.rs
```diff
@@ -81,3 +82,418 @@ pub(crate) fn immutable_headers(content_type: &'static str) -> HeaderMap {
...
+                return true;
+            }
+        }
+
+        false
+    }
+
+    /// Check for invalid globs or absolute paths.
+    pub fn validate_entry(entry: &str) -> Result<(), PathMatcherError> {
+        if let Err(e) = globset::Glob::new(entry) {
```
Urgau: I think we need to use the same builder otherwise we will run into mismatch.


```suggestion
        if let Err(e) = globset::GlobBuilder::new(entry)
            .empty_alternates(true)
            .build()
        {
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2457/ae8607011b2af8d419248517e68e0d4337b7d0a7..a97045eb9681d39cd3c86a9634630a5121ea1bef)*

Urgau (approved):
Looks good to me. Thanks for the change (and the improvements!).

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2457/ae8607011b2af8d419248517e68e0d4337b7d0a7..95a33250185b0cef41331def906455e3ccd7e441)*

tgross35:
Thanks!

PR #2458, opened 2026-07-09
Fix detection of PR merged state

### LINKED ISSUE #2455: Don't put "View changes since the review" for post-merge reviews
Subset of #2373 
If someone comments *after* a PR is merged, this should never be any changes, so the "View changes since the review" comment is not helpful
Real example that motivated this: rust-lang/reference#2262

Urgau:
For some reason GitHub stopped sending the boolean `merged` field in the `pull_request_review_comment.created` and `pull_request_review.submitted` webhook events.

We still have the `merged_at` field, so use that instead:
- https://api.github.com/repos/rust-lang/rust/pulls/142286

Fixes https://github.com/rust-lang/triagebot/issues/2455


Kobzol (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2458/6368d607b5d1520f2c6d6b5ee667d133de766680..4819796b62bf3622707b40226e381f58a0739cf8)*

PR #2462, opened 2026-07-11
Support filtering content mentions by paths

joshtriplett:
Some repositories would like to filter by both paths and content. For
instance, any mention of "miri" within "library/". Add support for this.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -299,6 +299,9 @@ pub(crate) struct MentionsEntryConfig {
     #[serde(alias = "type")]
     #[serde(default)]
     pub(crate) type_: MentionsEntryType,
+    /// Additional path globs to filter; primarily useful for `type = "content"`
+    #[serde(default)]
+    pub(crate) trigger_files: Vec<String>,
     pub(crate) message: Option<String>,
     #[serde(default)]
     pub(crate) cc: Vec<String>,
@@ -951,6 +954,12 @@ mod tests {
             message = "This is a message."
             cc = ["@someone"]
 
+            [mentions."miri"]
+            type = "content"
+            trigger_files = ["library/1", "library/2"]
+            message = "This is a message."
+            cc = ["@someone"]
+
             [shortcut]
 
             [issue-links]
@@ -1048,6 +1057,7 @@ mod tests {
                             "src/".to_string(),
                             MentionsEntryConfig {
                                 type_: MentionsEntryType::Filename,
+                                trigger_files: vec![],
                                 message: None,
                                 cc: vec!["@someone".to_string()]
                             }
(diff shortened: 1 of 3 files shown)


Urgau:
We have two kind of mentions, the content based one and the filename/filepath one.

This PR adds the `paths` config to both, and while it makes sense for the content based one, it doesn't really make sense for the filename one, it's already restricted by path.

We should IMO ignore (or better error) on `paths` with `type = "filename"` (the default).

src/config.rs
```diff
@@ -299,6 +299,9 @@ pub(crate) struct MentionsEntryConfig {
     #[serde(alias = "type")]
     #[serde(default)]
     pub(crate) type_: MentionsEntryType,
+    /// Additional path globs to filter; primarily useful for `type = "content"`
+    #[serde(default)]
+    pub(crate) paths: Vec<String>,
```
Urgau: You you also rename it to `trigger_paths`, we use that term in other configs.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2462/d1ddc97209ae4291d1e3a9092dd9f2e4249de270..ad9b55bf407892c3c505909b57a5e4f656432c2e)*

joshtriplett:
> We have two kind of mentions, the content based one and the filename/filepath one.
> 
> This PR adds the `paths` config to both, and while it makes sense for the content based one, it doesn't really make sense for the filename one, it's already restricted by path.
> 
> We should IMO ignore (or better error) on `paths` with `type = "filename"` (the default).

While I did primarily add it for content mentions, I do think it makes sense for filename mentions as well. The semantic there is "the primary filename  matches and at least one of the paths matches". So, you could use that intersection for "files named xyz, anywhere in one of these five directories", for instance. Or, conversely, "under this prefix, files with one of these names".

Urgau:
Except that's not how it works, filename is really just a path, to be precise a starts-with glob.

So if someone wants the "files named xyz, anywhere in one of these five directories" they can already do that:
```toml
[mentions."library/{dir1,dir2,dir3,dir4,dir5}/xyz"]
```

joshtriplett:
@Urgau I had not realized that `globset`, *unlike* `glob`, supports brace expansion. That obviates some of the use cases that I otherwise thought would be easier to write using multiple patterns.

Okay, I'll reject it during validation.

src/config.rs
joshtriplett: Will do. (The other configs use `trigger_files`, so I'll use that for consistency.)

Urgau:
We used to use the `glob` crate but switched to `globset` specifically to have support for alternates. It's unfortunate that our own crate `glob` doesn't support them.

Urgau (approved):
Thanks. Looks good to me.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2462/d1ddc97209ae4291d1e3a9092dd9f2e4249de270..5635e12981cd8ac46e53bd6b31941bea83ce9a9d)*

Urgau:
Would you mind sending a PR updating the documentation at https://github.com/rust-lang/rust-forge/blob/master/src/triagebot/mentions.md?

joshtriplett:
@Urgau Done: https://github.com/rust-lang/rust-forge/pull/1083

PR #2466, opened 2026-07-24
Permalink-ify branch file links to SHA file links

### LINKED ISSUE #2465: Detect non-permalinks in commit messages
I was about to write a CI check for this then remembered triagebot already does some link scanning, so this might be reasonable enough to add. In libc we require source links in the commit messages and/or PR descriptions, but sometimes we get links like:

https://github.com/torvalds/linux/blob/master/include/uapi/linux/audit.h#L389-L451

Which are prone to getting out of date whenever the file changes. A permalink would be better:

https://github.com/torvalds/linux/blob/248951ddc14de84de3910f9b13f51491a8cd91df/include/uapi/linux/audit.h#L389-L451
https://github.com/torvalds/linux/blob/248951/include/uapi/linux/audit.h#L389-L451

I think this could work by scanning for `github\.com/([^/\s]+)/([^/\s]+)/blob/(?P<ref>[^/\s]+)` then checking that `ref` matches `[[:xdigit:]]{8,}`.

Urgau:
This PR implements a canonicalization of branch name based GitHub file link.

So instead of having `https://github.com/torvalds/linux/blob/master/include/uapi/linux/audit.h#L389-L451` this PR would automatically update the PR description to `https://github.com/torvalds/linux/blob/248951ddc14de84de3910f9b13f51491a8cd91df/include/uapi/linux/audit.h#L389-L451`, or output a warning when found in commit messages.

Implements the plan from https://github.com/rust-lang/triagebot/issues/2465#issuecomment-5049505615

I re-used the `[issue-links]` config, but it feels similar enough and to avoid conflict if both wants to update the PR description.

Fixes https://github.com/rust-lang/triagebot/issues/2465
cc @tgross35

### DIFF
--- a/src/handlers/check_commits/branch_links.rs
+++ b/src/handlers/check_commits/branch_links.rs
@@ -1,0 +1,51 @@
+use crate::{
+    config::{IssueLinksCheckCommitsConfig, IssueLinksConfig},
+    github::GithubCommit,
+    handlers::{Context, check_commits::MERGE_IGNORE_LIST},
+};
+
+pub(super) async fn branch_links_in_commits(
+    ctx: &Context,
+    conf: &IssueLinksConfig,
+    commits: &[GithubCommit],
+) -> Option<String> {
+    match conf.check_commits {
+        IssueLinksCheckCommitsConfig::Off => return None,
+        IssueLinksCheckCommitsConfig::All | IssueLinksCheckCommitsConfig::Uncanonicalized => {}
+    }
+
+    let branch_links_commits = futures::future::join_all(
+        commits
+            .iter()
+            .filter(|c| {
+                !MERGE_IGNORE_LIST
+                    .iter()
+                    .any(|i| c.commit.message.starts_with(i))
+            })
+            .map(|c| async {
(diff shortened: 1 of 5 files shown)


src/handlers/check_commits/branch_links.rs
```diff
@@ -0,0 +1,51 @@
...
+    .await
+    .into_iter()
+    .flatten()
+    .collect::<String>();
+
+    if branch_links_commits.is_empty() {
+        None
+    } else {
+        Some(format!(
+            r"There are non-permalink in the commit message of the following commits. Please use a permalink instead.
```
Kobzol:
```suggestion
            r"There are links that are not permanent in the commit message of the following commits. Please use a permalink instead.
```

Kobzol (approved):
Seems a bit niche, but ok :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2466/f679945fe4a99853f67ebf27cab348e112a30829..c487fcfdbe63c052f1a9dfc8a644f14a0d7ade85)*

PR #2468, opened 2026-07-29
Fix race when removing the community review label and assigning last approvee

Urgau:
Happened at https://github.com/rust-lang/rust-clippy/pull/17461#issuecomment-5096921945
cc @samueltardieu

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -497,17 +497,8 @@ async fn update_community_review_assignment(
     let enough_reviews = approval_reviews.len() >= config.minimum_approvals.get().into();
 
     if enough_reviews {
-        issue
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
I'm not sure whether the race can't still happen, setting the assignment doesn't IMO guarantee that the next webhook about the removed label will already contain the latest state (I saw some cases in bors where this wasn't always the case).

But in any case, this should help make that much less likely, so 👍Thanks.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2468/54ea9e8b7d21c3304368c07d0712870d86cb571e..815bcbf5c978ced897b79d5db14853daa652867d)*

PR #2471, opened 2026-08-03
Remove our `cynic` dependency

Urgau:
Our `cycnic` dependency has bothered me for quite a while now.

This PR removes it by moving the two queries to our queries system. Saving a build script and 40 deps to build!

I've tested both migrated queries. Seems to work as expected.


src/github/queries/least_recently_reviewed.rs
```diff
@@ -0,0 +1,188 @@
...
+        name: &str,
+    ) -> anyhow::Result<Vec<PullRequest>> {
+        let mut prs = Vec::new();
+        let mut after = Option::<String>::None;
+
+        loop {
+            let mut data = self
+                .graphql_query(
+                    r#"
+        query LeastRecentlyReviewedPullRequests(
```
Kobzol, line 103: Seems like a good case for the local GH database mirror, to avoid these kinds of remote GH queries :)

Kobzol (commented):
Fully agreed that just writing these queries by hand is easier to read, understand and modify, and in general just much less magic happening. I'd be happy to see this dependency go away :)

That being said, since this is a rather large change, I think that we should do a vibe check on the triagebot Zulip channel about it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2471/9178a9b91a91fbdac52e2f8721da8a53edcf71a8..378c26f3122b2733565ce2e666eca8dd73b3ecf5)*

src/github/queries/least_recently_reviewed.rs
apiraino: what do you mean by local gh db? do you mean that to cache results of this query? This specific query is run when I prepare the weekly triage agenda, I need fresh data everytime

Kobzol: I meant https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Integrating.20a.20GitHub.20mirror.20.2B.20dashboard.20in.20triagebot/with/612107852, if we had the GitHub mirror in the triagebot DB, this could just be an SQL query (potentially with a force-refresh before running the data if up-to-date state is required).

apiraino: the "force-refresh" part is def. needed for my queries :)

Kobzol (approved):

PR #2474, opened 2026-08-12
Remove T-libs-api labels from code

clarfonthey:
Part of [RFC 3984](https://github.com/rust-lang/rfcs/pull/3984). Most of these don't actually matter, but I figured it's good to just remove mentions of T-libs-api from code for completeness so we can just grep for `libs-api` to make sure it's not being added to things.

In most cases, `T-libs` is added anyway as a precaution, but I figure we want to try and centralise on one label.

### DIFF
--- a/src/agenda.rs
+++ b/src/agenda.rs
@@ -29,7 +29,6 @@ pub fn prioritization() -> Box<dyn Action> {
                                 "final-comment-period",
                                 "major-change-accepted",
                                 "t-libs",
-                                "t-libs-api",
                                 "t-rustdoc",
                             ],
                         }),
@@ -46,7 +45,6 @@ pub fn prioritization() -> Box<dyn Action> {
                                 "finished-final-comment-period",
                                 "final-comment-period",
                                 "t-libs",
-                                "t-libs-api",
                                 "t-rustdoc",
                                 "S-waiting-on-author",
                             ],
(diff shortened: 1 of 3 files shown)


Urgau:
> make sure it's not being added to things

All of the references here are in the compiler agenda generator, which doesn't add any labels.

---------------

Regarding the change it-self, [I can still see 30 PRs with the label](https://github.com/rust-lang/rust/issues?q=is%3Aopen%20is%3Apr%20label%3A%22T-libs-api%22), some of which without `T-libs`, so this may regress some queries by not excluding some issues.

cc @apiraino

clarfonthey:
Note: I'm also currently working on fixing that as we speak, FWIW. Just takes a while to run a script with API limits and such. Also, I was under the impression the backporting here is actually the only place in the entire org where the labels are actually added to things, and that is outside the agenda generator.

Urgau:
> Also, I was under the impression the backporting here is actually the only place in the entire org where the labels are actually added to things

The one you modified are in our unit tests, they don't do anything, unless you are talking about another place.

clarfonthey:
Wow, you are actually right there. This is what I get for not reading closely.

Urgau (approved):
I've checked that the label is no longer in-used in any of the repo the agenda referenced.

The rest looks good to me as well.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2474/26271146bcbf326b445e7b8a0193f8ea614919c4..4ccaf0737129c7959f036b30c6eb0dbf46da2647)*

apiraino:
nice, thanks for doing the work for me :)

PR #2480, opened 2026-08-17
Shorter welcome message for new PRs

apiraino:
As I mentioned on [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/LLM.20policy.20link.20in.20welcome.20messages/near/614935488), I wanted to take a stab at cutting some words from the welcome message when a PR from a new contributor is opened.

<details><summary>Current version</summary>
<p>

grafik

</p>
</details> 

<details><summary>New version</summary>
<p>

grafik

</p>
</details> 

How do you feel about this proposal? :)

thanks

### DIFF
--- a/src/handlers/assign/messages.rs
+++ b/src/handlers/assign/messages.rs
@@ -8,7 +8,7 @@ use crate::github::IssueRepository;
 pub fn new_user_welcome_message(reviewer: &str) -> String {
     format!(
         "Thanks for the pull request, and welcome! \
-The Rust Project has assigned {reviewer} for review, you should hear from them (or someone else) \
+The Rust Project has assigned {reviewer} to review your changes, you should hear from them (or someone else) \
 within the next two weeks."
     )
 }
--- a/src/handlers/assign/messages.rs
+++ b/src/handlers/assign/messages.rs
@@ -8,8 +8,8 @@ use crate::github::IssueRepository;
 pub fn new_user_welcome_message(reviewer: &str) -> String {
     format!(
         "Thanks for the pull request, and welcome! \
-The Rust Project is excited to review your changes, and you should hear from {reviewer} \
-some time within the next two weeks."
+The Rust Project has assigned {reviewer} for review, you should hear from them (or someone else) \
+within the next two weeks."
     )
 }
 
@@ -36,13 +36,7 @@ In the meantime, we would highly appreciate if you could try to review any of [P
 pub fn contribution_message(contributing_url: &str, bot: &str) -> String {
     format!(
         "Please see [the contribution \
-instructions]({contributing_url}) for more information. Namely, in order to ensure the \
-minimum review times lag, PR authors and assigned reviewers should ensure that the review \
-label (`S-waiting-on-review` and `S-waiting-on-author`) stays updated, invoking these commands \
-when appropriate:
-
-- `@{bot} author`: the review is finished, PR author should check the comments and take action accordingly
-- `@{bot} review`: the author is ready for a review, this PR will be queued again in the reviewer's queue"
+         instructions]({contributing_url}) for more information."
     )
 }
 


src/handlers/assign/messages.rs
```diff
@@ -8,8 +8,8 @@ use crate::github::IssueRepository;
 pub fn new_user_welcome_message(reviewer: &str) -> String {
     format!(
         "Thanks for the pull request, and welcome! \
-The Rust Project is excited to review your changes, and you should hear from {reviewer} \
-some time within the next two weeks."
+The Rust Project has assigned {reviewer} to review your changes, you should hear from them \
```
Kobzol:
```suggestion
The Rust Project has assigned {reviewer} to review your changes, you should hear from them (or someone else) \
```
What do you think about keeping this? It is more accurate, after all.

Kobzol (commented):
I like it! It clearly separates the contribution link (which might also contain the LLM policy link in the future) from the minutiae of describing how the labels work, which is quite low-level, and before it was combined with the rest of the text too much.

Regarding the tip: does it really make sense to include `@rustbot author` there? After all, this command most likely won't be used by (new) PR authors, and it doesn't make sense to me to post instructions that are essentially for the reviewer (who likely knows about this already). I think that we should focus on tips for the PR author themselves, so maybe just keep the `@rustbot ready` command.

In fact, maybe we should just remove the tip altogether, because rustbot now actually responds to `@rustbot author` with a message saying that `@rustbot ready` should be used. So I'm not sure if it has to also be in the welcome message. It can be a bit confusing, because it describes a situation that might not happen at all, and that is not applicable immediately once the message is sent, because at that time the PR is ready for a review.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2480/847d96408906d93f897005622c697bdee30f203e..bbbc938de605fc48485d05c8da795a0cefd10d19)*

apiraino:
> In fact, maybe we should just remove the tip altogether

Ok, why not. So, instead of the tip block, something more like this?

---

To try to keep review times lag at a minimum, PR authors and assigned reviewers should ensure that these labels stays updated:
- `S-waiting-on-author`: PR is reviewed, ask the author to check the comments and take action accordingly
- `S-waiting-on-review`: PR is ready for another review, it will be queued again in the reviewer's queue

Kobzol:
To clarify, by removing the tip, I meant removing this entire block of text :sweat_smile: If we keep the text, then I think it should be wrapped in the tip block.

apiraino:
sorry, now I am confused: are you suggesting to remove the entire block of instructions about how to operate the review switch (either by using the bot or changing the labels)?

i.e. the comment in full to be like this:

---

Thanks for the pull request, and welcome! \
The Rust Project has assigned @ghost for review, you should hear from them (or someone else) \
within the next two weeks.

Please see [the contribution instructions](#) for more information.

<details><summary>Why was this reviewer chosen?</summary>
<p>
blahblah
</p>
</details>

Kobzol:
Yes, that was my proposal. Because many contributions will not even interact with this system. PRs are implicitly ready for a review when not opened as a draft. If a reviewer uses `@rustbot author`, triagebot will post a message that explains what to do, so this doesn't have to be in the initial comment. And reviewers also don't need to read about this in the initial comment, as they know how the system works.

apiraino:
> Yes, that was my proposal. Because many contributions will not even interact with this system. 

Hm not sure about that. The original intent of this text (IIRC) was to instruct (and remind) both authors (esp. first-timers) and reviewers about what to do and *to do it*. We don't have hard data about how effective has this been, but do you think it's not a useful reminder?

Kobzol:
What I meant was that many PRs are reviewed and merged without ever `@rustbot ready/author` being used. I don't think that we necessarily have to remind the reviewers about this, I kind doubt that reviewers ever read this message (I only ever look at it to figure out why was someone assigned, but I don't read the text).

And we explain this to PR authors when it actually becomes relevant (when the reviewer uses `@rustbot author`).

So I think that we have the following options:
1. Keep showing both `@rustbot ready` and `@rustbot author` (and put both into the Tip block, I think that is good).
2. Only show `@rustbot ready`, because reviewers probably won't be reading this text anyway (IMO)
3. Don't show either of them

Pick what you prefer :)

apiraino:
ok, I see. Well, I don't know either, from me it would just be an uneducated guess :shrug: 

I will remove it and call it a day with this experiment (where we tried to educate people handling the PR review status)

src/handlers/assign/messages.rs
```diff
@@ -8,8 +8,8 @@ use crate::github::IssueRepository;
 pub fn new_user_welcome_message(reviewer: &str) -> String {
     format!(
         "Thanks for the pull request, and welcome! \
-The Rust Project is excited to review your changes, and you should hear from {reviewer} \
-some time within the next two weeks."
+The Rust Project has assigned {reviewer} for review, you should hear from them (or someone else) \
```
Kobzol:
```suggestion
The Rust Project has assigned {reviewer} to review your changes, you should hear from them (or someone else) \
```

Kobzol (approved):
I think that the previous review messaging was clearer. Otherwise LGTM, let's land this and if someone doesn't like it, we can always easily revert or modify it further.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2480/8df98a13a61dbff2fb325ebe6d4d32009a3f7b39..a1520568b20262874abe85a5e1d82370abe805a7)*

Kobzol (commented):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2480/8df98a13a61dbff2fb325ebe6d4d32009a3f7b39..9d38ad9540a15bd15561ac33a1ab2fff0381e8e5)*

PR #2483, opened 2026-08-17
Bubble-up a proper user error instead of swallowing failed assignment

### LINKED ISSUE #2481: Welcome message inconsistent with invalid assignee
In https://github.com/rust-lang/rust/pull/161207, the author of the PR tried to “r? @kpreid”, but I am not a valid assignee, and rustbot posted two comments, 1:

> Failed to set assignee to `kpreid`: invalid assignee
> 
> > **Note**: Only org members with at least the repository "read" role, users with write permissions, or people who have commented on the PR may be assigned.

2:

> Thanks for the pull request, and welcome! The Rust Project is excited to review your changes, and you should hear from @kpreid (or someone else) some time within the next two weeks.

The welcome message should account for the failure to assign, and not claim that I (or any other non-member GitHub user) will do anything.

Urgau:
Currently the assign handler is swallowing failed assignment errors from the surrounding control flow, which means that we post the error but still post the assignments messages and update the DB like the assignment successfully happened, while it didn't.

Instead of that, let's return a `UserError` from the handler so we can properly exit the handler and post only the error.

Fixes #2481

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -587,23 +587,14 @@ async fn set_assignee(
             issue.global_id(),
             reviewer.name,
         );
-        if let Err(e) = issue
-            .post_comment(
-                &ctx.github,
-                &format!(
-                    "Failed to set assignee to `{}`: {err}\n\
-                     \n\
-                     > **Note**: Only org members with at least the repository \"read\" role, \
-                       users with write permissions, or people who have commented on the PR may \
-                       be assigned.",
-                    reviewer.name
-                ),
-            )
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Makes sense, thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2483/177d3a7b02d3053d2ca9d13eac47d57c87787006..bfbbcadccd21000bb90a4d69cbc45d85c962603d)*

PR #2485, opened 2026-08-20
Deny warnings on CI

Kobzol:
We recently let a warning slip through, we should not allow that on CI.


Urgau (approved):
I'm not always a huge fan of denying all warnings due to lints changes, but let's try it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2485/1341b138588bba194dc811ad1a39f5148a246b6b..283f9d6827c7b94bf8d94cdbe58d989c5806ecff)*

Kobzol:
Note that this is only for the basic rustc lints, not Clippy. So I don't expect for it to be too annoying.

PR #2487, opened 2026-08-20
Fix sending of Zulip DMs

Kobzol:
The e-mail field shouldn't really have been used, I removed it and switched to IDs instead. Also migrated the DM message type to `direct` from `private`, as per https://zulip.com/api/send-message#parameter-type.

Note: according to the documentation, the `to` field for DMs should actually be a list. But it seems that Zulip accepts also just a single value there.

### DIFF
--- a/src/handlers/project_goals.rs
+++ b/src/handlers/project_goals.rs
@@ -755,7 +755,7 @@ async fn send_dm(
     }
 
     MessageApiRequest {
-        recipient: Recipient::Private { id: owner.0 },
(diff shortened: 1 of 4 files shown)


Urgau (approved):
The usage of email is indeed a bit weird. The rest of the changes looks good.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2487/57bd599b3984477d006398d287a6609b98b57998..54687a0d217c677595d69de9bd0628cec52ad837)*

PR #2490, opened 2026-08-25
Fetch all the reviews instead of just the first 100s

Urgau:
Follow-up to https://github.com/rust-lang/triagebot/pull/2489 where we now actually fetch and check all the reviews.

Used `futures::stream::unfold` to lazily load the pages.

<details>

Tested with:

```rust
    {
        let gh = github::GithubClient::new_from_env();
        let issue = gh
            .issue(
                &github::IssueRepository {
                    organization: "rust-lang".to_string(),
                    repository: "rust".to_string(),
                },
                137944,
            )
            .await
            .unwrap();

        use futures::StreamExt;

        let review_stream = issue.get_all_reviews(&gh);
        tokio::pin!(review_stream);

        while let Some(reviews) = review_stream.next().await {
            dbg!(&reviews);
        }
    }
```

</details>

### DIFF
--- a/src/github/issue.rs
+++ b/src/github/issue.rs
@@ -674,7 +674,11 @@ impl Issue {
 
         Ok(())
     }
+}
 
+const REVIEWS_PAGE_SIZE: usize = 100;
+
+impl Issue {
     pub async fn get_review(
         &self,
         client: &GithubClient,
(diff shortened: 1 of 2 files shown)


Kobzol (approved):
I guess that it would be a bit more natural if `get_all_reviews` returned a stream of `Comment`, not a stream of `Vec<Comment>`.

I tried something with `futures`, but it's not super pretty:
```rust
pub fn get_all_reviews(
        &self,
        client: &GithubClient,
    ) -> impl futures::Stream<Item = anyhow::Result<Comment>> {
        futures::stream::try_unfold(1, move |current_page| async move {
            match self.get_reviews(client, current_page).await {
                Ok(reviews) => {
                    let next_state = if reviews.len() >= REVIEWS_PAGE_SIZE {
                        current_page + 1
                    } else {
                        return Ok(None);
                    };

                    Ok(Some((
                        futures::stream::iter(reviews.into_iter().map(Ok)),
                        next_state,
                    )))
                }
                Err(e) => Err(e), // Stop on error
            }
        })
        .try_flatten()
    }
```
Async generators can't come soon enough 😿

So let's keep it like this. Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2490/a3fd33d6fb7dd31e1a10d63e8d819e1d986b7aff..55d9a34ca914bd5e7f5ff29b9c31bc47e13fbf70)*

Urgau:
It's actually that not bad. I had a completely different approach of unwrapping at the call-site, but decided that it wasn't worth the extract complexity.

PR #2491, opened 2026-08-27
resolve zulip topic on issue priority assigned

### LINKED ISSUE #2139: Add an option to `NotifyZulipLabelConfig` to automatically resolve the zulip topic when the label is removed
For some zulip channels (such as `t-compiler/prioritization/alerts`), this seems like an useful option to distinguish the unprioritized from the priorized issues without having to visit each topic manually.
On the other hand, a channel like `t-compiler/backports` may not want the topic to be resolved just by removing the `*-nominated` label, so this should be off by default.
If this is desired and possible, I will try to implement this in the next days.

@rustbot claim

apiraino:
Closes #2139

As per [Zulip discussion](https://rust-lang.zulipchat.com/#narrow/channel/227806-t-compiler.2Fops/topic/.E2.9C.94.20alarts.20mark.20as.20resolved.20once.20.20gets.20removed/near/619483354).

This should resolve the Zulip topic when a regression has been assigned a priority label.

Note after a first test: if an issue is solved and then given again a `I-prioritize` label, a new zulip topic will be created (because the title of the old one has now a "✔ ".

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -375,10 +375,19 @@ async fn handle_command<'a>(
                 StreamCommand::AssignPriority { issue_num, prio } => {
                     let _ = match assign_issue_prio(&ctx, message_data, issue_num, prio).await {
                         // give user feedback
-                        Ok(_) => ctx.zulip.add_reaction(message_data.id, "check").await,
+                        Ok(_) => {
+                            // emoji reaction for the triagebot command
+                            ctx.zulip.add_reaction(message_data.id, "check").await?;
+                            // resolve the Zulip topic
+                            if let Some(ref topic) = message_data.subject {
+                                ctx.zulip.resolve_topic(message_data.id, &topic).await?;
+                            } else {
+                                log::warn!("Zulip topic has empty subject");
+                            }
+                        }
                         Err(err) => {
                             log::error!("Could not assign priority to #{}: {:?}", issue_num, err);
-                            ctx.zulip.add_reaction(message_data.id, "scream").await
+                            ctx.zulip.add_reaction(message_data.id, "scream").await?;
                         }
                     };
                     Ok(None)
--- a/src/zulip/client.rs
+++ b/src/zulip/client.rs
@@ -203,7 +203,11 @@ impl ZulipClient {
     }
 
     // https://zulip.com/api/update-message#parameter-topic
-    pub(crate) async fn resolve_topic(&self, message_id: u64, topic: &str) -> anyhow::Result<()> {
+    pub(crate) async fn resolve_topic(
+        &self,
+        message_id: u64,
+        message_topic: &str,
+    ) -> anyhow::Result<()> {
         #[derive(serde::Serialize)]
         struct ResolveTopic<'a> {
             topic: &'a str,
@@ -212,10 +216,19 @@ impl ZulipClient {
             send_notification_to_new_thread: bool,
         }
 
+        let mut topic = message_topic.replacen("", "✔ ", 1);
+        // Maximum 60, minus the elipsis
+        let mut chars = topic.char_indices().skip(60 - 1);
+        if let Some((len, _)) = chars.next()
+            && chars.next().is_some()
+        {
+            topic = format!("{}…", &topic[..len]);
+        }
+
         let resp = self
             .make_request(Method::PATCH, &format!("messages/{message_id}"))
             .form(&ResolveTopic {
-                topic: &topic.replacen("", "✔ ", 1),
+                topic: &topic,
                 propagate_mode: "change_all",
                 send_notification_to_old_thread: false,
                 send_notification_to_new_thread: false,
--- a/src/zulip/client.rs
+++ b/src/zulip/client.rs
@@ -201,6 +201,40 @@ impl ZulipClient {
                 .into()
         })
     }
+
+    // https://zulip.com/api/update-message#parameter-topic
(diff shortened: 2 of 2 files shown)


apiraino:
r? @Urgau 

(maybe? feel free to reroll if you don't have time)

Urgau:
@apiraino to assign someone from T-triagebot you can use `r? triagebot`

Urgau:
> if an issue is solved and then given again a `I-prioritize` label, a new zulip topic will be created (because the title of the old one has now a "✔ ".

Hopefully this is not too much of an issue.

src/zulip/client.rs
```diff
@@ -208,6 +201,44 @@ impl ZulipClient {
                 .into()
         })
     }
+
+    // https://zulip.com/api/update-message#parameter-topic
+    pub(crate) async fn resolve_topic(&self, message_data: super::Message) -> anyhow::Result<()> {
```
Urgau: This is `update_message` but with `send_notification_to_new_thread` set to false right?

Should we combine them together?

I'm fine keeping them separate but without taking `message_data: super::Message` and instead taking the message ID and previous topic name, like this:

```suggestion
    pub(crate) async fn resolve_topic(&self, message_id: u64, topic: &str) -> anyhow::Result<()> {
```

src/zulip/client.rs
```diff
@@ -208,6 +201,44 @@ impl ZulipClient {
...
+            send_notification_to_old_thread: bool,
+            send_notification_to_new_thread: bool,
+        }
+
+        let resp = self
+            .make_request(Method::PATCH, &format!("messages/{message_id}"))
+            .form(&ResolveTopic {
+                topic: &topic,
+                propagate_mode: "change_all",
+                ..Default::default()
```
Urgau: I know this is temping but can you explicit the fields since their values are load-bearing.


```suggestion
                send_notification_to_old_thread: false,
                send_notification_to_new_thread: false,
```

src/zulip.rs
```diff
@@ -72,7 +71,7 @@ pub struct Request {
 }
 
 #[derive(Clone, Debug, serde::Deserialize)]
-struct Message {
+pub(crate) struct Message {
```
Urgau:
```suggestion
struct Message {
```

Urgau (commented):
Looks mostly good to me. Only two small nits.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2491/32ce419db5a158a00b4f9881723c03c4c2fdaef0..75f1600d8557ff1766f01c704bcfe4b10b835527)*

apiraino:
thanks for the review

src/zulip/client.rs
```diff
@@ -208,6 +201,40 @@ impl ZulipClient {
...
+            topic: &'a str,
+            propagate_mode: &'a str,
+            send_notification_to_old_thread: bool,
+            send_notification_to_new_thread: bool,
+        }
+
+        let resp = self
+            .make_request(Method::PATCH, &format!("messages/{message_id}"))
+            .form(&ResolveTopic {
+                topic: &topic.replacen("", "✔ ", 1),
```
Urgau: Forgot that we still need to limit the topic length to 60 characters (since we are adding 2 characters).

*Took inspiration from [our major change](https://github.com/rust-lang/triagebot/blob/32ce419db5a158a00b4f9881723c03c4c2fdaef0/src/handlers/major_change.rs#L445-L455)*

```suggestion
                topic: {
                    let mut topic = topic.replacen("", "✔ ", 1);
                    
                    // Maximum 60, minus the elipsis
                    let mut chars = topic
                        .char_indices()
                        .skip(60 - 1);
                    if let Some((len, _)) = chars.next()
                        && chars.next().is_some()
                    {
                        topic = format!("{}…", &topic[..len]);
                    }
                    
                    &*topic
                }
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2491/32ce419db5a158a00b4f9881723c03c4c2fdaef0..3cbb099e331fda88a280ce91a242a31c11cf2a2c)*

apiraino: This code is not very readable but let's accept it for the sake of having this PR approved. Ideally the correct solution would be to call a Zulip API that also returns maximum topic length

apiraino: also, when you suggest inline changes, could you please also test that your changes compile? :sweat_smile:

Urgau: > This code is not very readable but let's accept it for the sake of having this PR approved.

Agree with you. I looked if std provided something better but it doesn't seems like it's the case. Feel free to send a PR improving it.

> Ideally the correct solution would be to call a Zulip API that also returns maximum topic length

Happy to accept such change.

> also, when you suggest inline changes, could you please also test that your changes compile? 😅

I checked it in isolation [on the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=fff43d35a28f5d4ebbb8c05d80ed32c0), but yeah I didn't checked it on triagebot.

To be honest it's kind of pain to checkout changes of another person with GitHub, I wish where was a simple command I could use (without the `gh` CLI).

Urgau (approved):
Thanks.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2491/32ce419db5a158a00b4f9881723c03c4c2fdaef0..f1c76ab5a750a31df89099a3295c5a3b69f97032)*

PR #2492, opened 2026-08-27
Fix sending of Zulip DMs (take 2)

Kobzol:
Continuation of https://github.com/rust-lang/triagebot/pull/2487. This time I actually tested it locally on Zulip to see that it works :D :see_no_evil: Sorry.

### DIFF
--- a/src/zulip/client.rs
+++ b/src/zulip/client.rs
@@ -80,7 +80,14 @@ impl ZulipClient {
                 },
                 to: match recipient {
                     Recipient::Stream { id, .. } => id.to_string(),
-                    Recipient::DirectMessage { id, .. } => id.to_string(),
+                    // Zulip requires us to send a "list of numbers" when sending DMs
+                    // (see https://zulip.com/api/send-message).
+                    // So we have to send for example `to=[1]`.
+                    // However, the serde-urlencoded crate, which is used by `reqwest`'s form
+                    // method, does not support serializing sequences as struct values.
+                    // So we have to "pre-serialize" it into a string, so that it can be sent, and
+                    // Zulip interprets it as a list of numbers.
+                    Recipient::DirectMessage { id, .. } => format!("[{id}]"),
                 },
                 topic: match recipient {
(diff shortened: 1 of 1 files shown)


nxsaken (approved):

src/zulip/client.rs
```diff
@@ -80,7 +80,14 @@ impl ZulipClient {
                 },
                 to: match recipient {
                     Recipient::Stream { id, .. } => id.to_string(),
-                    Recipient::DirectMessage { id, .. } => id.to_string(),
+                    // Zulip requires us to send a "list of numbers" when sending DMs
+                    // (see https://zulip.com/api/send-message).
```
Urgau, line 84: That makes no sense to me, [Zulip own documentation](https://zulip.com/api/send-message#parameter-to) which clearly states that we can send either a standalone number/string or a list of numbers/strings.

I don't understand where this requirement comes from. Is Zulip own documentation wrong?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2492/0afe6d4537a4e4824c04ec8eff47b169b6371eff..03f4bb3276091c5ba52a9ab42c2ecb6314314ec3)*

Urgau: Oh I see, it's right after:
> For direct messages, this is either a list containing integer user IDs or a list containing string Zulip API email addresses.

This is really not obvious and quite confusing.

Urgau (approved):
LGTM

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2492/0afe6d4537a4e4824c04ec8eff47b169b6371eff..03f4bb3276091c5ba52a9ab42c2ecb6314314ec3)*

Kobzol:
Yeah, it is confusing. They say that for DMs, you have to send a list, but before it clearly worked when we were sending just a single email. It doesn't help that the data is received as a form, which doesn't really have great support for lists in general.

PR #2493, opened 2026-08-27
Add Zulip command to ping goal owners

nxsaken:
At this point we haven't had a need for surgical pings, so I kept it simple.

### DIFF
--- a/src/zulip.rs
+++ b/src/zulip.rs
@@ -263,7 +263,7 @@ async fn handle_command<'a>(
             } => unlock_cmd(&ctx, gh_id, organization, repo, *id).await,
             ChatCommand::Work(cmd) => workqueue_commands(&ctx, gh_id, cmd).await,
             ChatCommand::DocsUpdate => trigger_docs_update(&ctx.zulip, message_data),
-            ChatCommand::GoalsPingOwners => Ok(Some(goals_ping_owners(&ctx).await)),
+            ChatCommand::GoalsPingOwners => Ok(Some(goals_ping_owners(ctx.clone()))),
             ChatCommand::UserInfo {
                 username,
                 organization,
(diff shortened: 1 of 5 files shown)


src/zulip.rs
```diff
@@ -1649,6 +1652,13 @@ async fn post_waiter(
     Ok(None)
 }
 
+async fn goals_ping_owners(ctx: &Context) -> String {
+    match goals::ping_owners(&ctx.github, &ctx.zulip, &ctx.team, false).await {
```
nxsaken: I now see that the below fn may timeout, so I probably should do the same...

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2493/e3d010c0a6074784d0d253568cd61040e4bd7b8d..0e7d453a43c115d6f0ff9edaffe5839abcfa28f4)*

src/zulip.rs
```diff
@@ -1649,6 +1652,16 @@ async fn post_waiter(
     Ok(None)
 }
 
+fn goals_ping_owners(ctx: Arc<Context>) -> String {
+    tokio::task::spawn(async move {
+        match goals::ping_owners(&ctx.github, &ctx.zulip, &ctx.team, false).await {
+            Ok(()) => log::info!("Goal owner pings finished."),
```
Urgau: Just a nit, but in the previous version of command, I added a "finished" message, might be good to also re-add it, otherwise there is no way to know if it finished or if there is a error.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2493/e3d010c0a6074784d0d253568cd61040e4bd7b8d..da7ee3834fdcfc752ecc1d73f289b3a516951c19)*

nxsaken: I was about to just copy it from the fn below and forgot 🤦

Kobzol (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2493/e3d010c0a6074784d0d253568cd61040e4bd7b8d..cfd55675c30f03e87c4689a3948c6e8ede623d5e)*

Kobzol (commented):
Actually, let's only enable it for members of t-goals. Just in case :)

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2493/e3d010c0a6074784d0d253568cd61040e4bd7b8d..cfd55675c30f03e87c4689a3948c6e8ede623d5e)*

nxsaken:
Ahhh, I thought about adding that check there and forgot it didn't really happen 😆

Kobzol (approved):

PR #2494, opened 2026-08-27
Handle re-opening of PRs under community reviews assignment

Urgau:
When PRs under community reviews assignment are closed we remove the label (eg. `S-waiting-on-community-reviews`), which is expected, but we didn't added it back when the PR was being reopened, causing the PR to never get a assignee after reaching the desired minimum community reviews.

This PR therefore adds the plumbing to re-add the label when the PR is being re-opened.

Reported at https://github.com/rust-lang/rust-clippy/pull/17634#issuecomment-5444191077

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -90,6 +90,7 @@ fn get_repo_workqueue(ctx: &Context, repo: &str) -> Arc<RwLock<ReviewerWorkqueue
 #[derive(Debug)]
 pub(super) enum AssignInput {
     Opened { draft: bool },
+    Reopened { draft: bool },
     Closed,
     ReadyForReview,
     UnlabeledMinReviews { draft: bool, open: bool },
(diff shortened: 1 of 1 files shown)


Urgau:
Hum, there is technically the case of the assignee un-assigning themselves from the PR before it's re-opened, in that case there is enough approvals, but I don't know what to do in that case.

The current logic in the PR only adds the label and doesn't not assign someone, same as with normal PRs. I guess we'll figure-out if we ever have this case.

src/handlers/assign.rs
```diff
@@ -191,6 +195,38 @@ pub(super) async fn handle_input(
...
+                }
+                false
+            } else {
+                // We have historically done nothing to do for normal PRs, even if there
+                // are no current assignee
+                false
+            }
+        }
+        AssignInput::Reopened { draft: true } => {
+            // We have historically done nothing for normal PR and community reviews are only handled
```
Kobzol: Should this say draft instead of normal?

Kobzol (approved):
Indeed, one at a time. Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2494/e3d010c0a6074784d0d253568cd61040e4bd7b8d..cdaa6d3e758b5f97a36acbe1adbcc62f6e578c82)*

src/handlers/assign.rs
Urgau: Normal is the word I meant, but I can see how it's confusing, I've clarified that it means "non-community reviewed PRs".

PR #2496, opened 2026-08-29
Handle case where community review threshold was reached before reopen

Urgau:
Follow-up to https://github.com/rust-lang/triagebot/pull/2494 meant to address https://github.com/rust-lang/triagebot/pull/2494#issuecomment-5445627199.

Let's just check if the threshold for minimum community reviews had already been reached before and only if not re-add the "monitoring" label.

### DIFF
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -197,7 +197,7 @@ pub(super) async fn handle_input(
         }
         AssignInput::Reopened { draft: false } => {
             if assign_command.is_none()
-                && !event.issue.assignees.is_empty()
+                && event.issue.assignees.is_empty()
                 && let Some(community_reviews) = &config.community_reviews
             {
                 // There are no assignees for this PR and we are under community reviews.
--- a/src/handlers/assign.rs
+++ b/src/handlers/assign.rs
@@ -196,15 +196,24 @@ pub(super) async fn handle_input(
             assign_command.as_ref().is_some_and(|a| a != GHOST_ACCOUNT)
         }
         AssignInput::Reopened { draft: false } => {
-            if let Some(community_reviews) = &config.community_reviews
-                && assign_command.is_none()
+            if assign_command.is_none()
+                && !event.issue.assignees.is_empty()
+                && let Some(community_reviews) = &config.community_reviews
             {
-                // There are no current assignees, add the community review label
+                // There are no assignees for this PR and we are under community reviews.
                 //
-                // No need to worry about approvals that may have happend between
-                // the time the PR was closed and this re-opening as GitHub doesn't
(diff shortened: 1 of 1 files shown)


src/handlers/assign.rs
Kobzol, line 528: If a user reviews multiple times, we consider their first review, w.r.t. `last_review_user`. Maybe we should instead consider their last?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2496/0d7ad16e4d5ec785d4d8a44eb298f719441c3b00..ec9903568c34551057514bf15a73a19567ca20dc)*

src/handlers/assign.rs
```diff
@@ -196,15 +196,24 @@ pub(super) async fn handle_input(
...
         }
         AssignInput::Reopened { draft: false } => {
-            if let Some(community_reviews) = &config.community_reviews
-                && assign_command.is_none()
+            if assign_command.is_none()
+                && !event.issue.assignees.is_empty()
+                && let Some(community_reviews) = &config.community_reviews
             {
-                // There are no current assignees, add the community review label
+                // There are no assignees for this PR and we are under community reviews.
```
Kobzol, line 203: > There are no assignees

The comment says this, but the code above checks `!event.issue.assignees.is_empty()`, which means "if there are assignees".

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2496/0d7ad16e4d5ec785d4d8a44eb298f719441c3b00..ec9903568c34551057514bf15a73a19567ca20dc)*

src/handlers/assign.rs
Urgau: I don't think it matters, the last reviewer is necessarily one that hasn't approved the PR yet, otherwise we would already have assigned someone.

src/handlers/assign.rs
Urgau: Indeed, my bad. Fixed.

src/handlers/assign.rs
Kobzol: I tried to think of possible edge cases, but probably this isn't an issue. Nevermind.

Kobzol (approved):

Urgau:
Weird, a docker failure. Let's hope it's just spurious.

<details>

```text

#26 ERROR: error writing layer blob: not_found
------
 > exporting to GitHub Actions Cache:
------

 1 warning found (use docker --debug to expand):
 - JSONArgsRecommended: JSON arguments recommended for CMD to prevent unintended behavior related to OS signals (line 57)
ERROR: failed to build: failed to solve: error writing layer blob: not_found

```

</details>

PR #2497, opened 2026-09-02
Implement crate yanking/unyanking on Zulip

Kobzol:
To enable us to move forward with removing individual and team access for `rust-lang` crates.io crates (see https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Crate.20yanking.2Funyanking.20on.20Zulip).

This will need the crates.io yank-only `rust-lang-owner` crates.io token configured in the triagebot environment, in the `CRATES_IO_API_TOKEN` environment variable.

### DIFF
--- a/src/crates_io/mod.rs
+++ b/src/crates_io/mod.rs
@@ -1,0 +1,83 @@
+use reqwest::Client;
+use reqwest::header;
+use reqwest::header::{HeaderMap, HeaderValue};
+use secrecy::{ExposeSecret, SecretString};
+use serde::Serialize;
+use std::env;
+use std::sync::OnceLock;
+
+// OpenAPI spec: https://crates.io/api/openapi.json
+const CRATES_IO_BASE_URL: &str = "https://crates.io/api/v1";
+
+/// Access to the Crates.io API
+pub struct CratesIoApi {
+    client: Client,
+    // The token is loaded lazily, to avoid requiring the API token if crates.io APIs are not
+    // actually accessed.
+    token: OnceLock<SecretString>,
+}
+
+impl CratesIoApi {
+    pub fn new() -> Self {
+        let mut map = HeaderMap::default();
+        map.insert(
+            header::USER_AGENT,
+            HeaderValue::from_static("triagebot@rust-lang.org"),
+        );
+
+        Self {
+            client: reqwest::ClientBuilder::default()
+                .default_headers(map)
+                .build()
+                .unwrap(),
+            token: Default::default(),
+        }
+    }
+
+    fn get_api_token(&self) -> &SecretString {
+        self.token.get_or_init(|| {
+            env::var("CRATES_IO_API_TOKEN")
+                .expect("CRATES_IO_API_TOKEN is missing")
+                .into()
+        })
+    }
+
+    /// Yanks a crate from crates.io.
+    pub(crate) async fn yank_crate(&self, krate: &str, version: &str) -> anyhow::Result<()> {
+        self.req::<()>(
+            reqwest::Method::DELETE,
+            &format!("/crates/{krate}/{version}/yank"),
+        )
+        .await?
+        .error_for_status()?;
+
+        Ok(())
+    }
+
+    /// Unyanks a crate from crates.io.
+    pub(crate) async fn unyank_crate(&self, krate: &str, version: &str) -> anyhow::Result<()> {
+        self.req::<()>(
+            reqwest::Method::PUT,
+            &format!("/crates/{krate}/{version}/unyank"),
+        )
+        .await?
+        .error_for_status()?;
+
+        Ok(())
+    }
+
+    /// Perform a request against the crates.io API
+    async fn req<T: Serialize>(
+        &self,
+        method: reqwest::Method,
+        path: &str,
+    ) -> anyhow::Result<reqwest::Response> {
+        let token = self.get_api_token();
+        let req = self
+            .client
+            .request(method, format!("{CRATES_IO_BASE_URL}{path}"))
+            .bearer_auth(token.expose_secret());
+
+        Ok(req.send().await?)
+    }
+}
--- a/src/crates_io/mod.rs
+++ b/src/crates_io/mod.rs
@@ -5,6 +5,7 @@ use secrecy::{ExposeSecret, SecretString};
 use serde::Serialize;
 use std::env;
 use std::sync::OnceLock;
+use url::Url;
 
 // OpenAPI spec: https://crates.io/api/openapi.json
 const CRATES_IO_BASE_URL: &str = "https://crates.io/api/v1";
@@ -43,10 +44,10 @@ impl CratesIoApi {
     }
 
     /// Yanks a crate from crates.io.
-    pub(crate) async fn yank_crate(&self, krate: &str, version: &str) -> anyhow::Result<()> {
+    pub async fn yank_crate(&self, krate: &str, version: &str) -> anyhow::Result<()> {
         self.req::<()>(
             reqwest::Method::DELETE,
-            &format!("/crates/{krate}/{version}/yank"),
+            self.url(&["crates", krate, version, "yank"]),
         )
         .await?
         .error_for_status()?;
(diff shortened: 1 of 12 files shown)


src/crates_io/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+                .expect("CRATES_IO_API_TOKEN is missing")
+                .into()
+        })
+    }
+
+    /// Yanks a crate from crates.io.
+    pub(crate) async fn yank_crate(&self, krate: &str, version: &str) -> anyhow::Result<()> {
+        self.req::<()>(
+            reqwest::Method::DELETE,
+            &format!("/crates/{krate}/{version}/yank"),
```
marcoieni: we should validate both `krate` and `version`. So people can't pass stuff containing `../` that could delete crates.

found via AI review:

Image

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2497/32ce419db5a158a00b4f9881723c03c4c2fdaef0..c99d98514a0e0e940d9349c14e32a15794ab3467)*

Kobzol: Huh, I had no idea that reqwest normalizes `..`. But I don't think that we should validate parameters inside this function, and in the Zulip code we do validate it. If you don't pass a crate name found in the `team` DB (which anything containing `..` surely won't be), it will error out.

marcoieni: where is the version validated?

Kobzol: Ah, I see, I was talking about crate, but didn't realize that this was talking about the version. Yeah, we should indeed validate that, good find! Probably I should add the `semver` crate and just ensure that it is a valid version. Or we can just check that it doesn't contain `..` or a slash.

marcoieni: I think the semver crate would be nice, as it can catch typos 👍

Kobzol: Ok, I added `semver` and also ensured that the URL injection can't happen anymore. Great catch, I had no idea that the URLs are normalized. We need to be aware of this also in other places.

src/crates_io/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn get_api_token(&self) -> &SecretString {
+        self.token.get_or_init(|| {
+            env::var("CRATES_IO_API_TOKEN")
+                .expect("CRATES_IO_API_TOKEN is missing")
+                .into()
+        })
+    }
+
+    /// Yanks a crate from crates.io.
+    pub async fn yank_crate(&self, krate: &str, version: &str) -> anyhow::Result<()> {
```
Urgau: Given the somewhat sensitive nature of those operations, I wonder if we should at least log them.


```suggestion
    #[instrument(level = "info", skip(self))]
    pub async fn yank_crate(&self, krate: &str, version: &str) -> anyhow::Result<()> {
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2497/32ce419db5a158a00b4f9881723c03c4c2fdaef0..083197c4a9291a1fbb592894248c5e8106f63cf8)*

src/crates_io/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+            self.url(&["crates", krate, version, "yank"]),
+        )
+        .await?
+        .error_for_status()?;
+
+        Ok(())
+    }
+
+    /// Unyanks a crate from crates.io.
+    pub async fn unyank_crate(&self, krate: &str, version: &str) -> anyhow::Result<()> {
```
Urgau:
```suggestion
    #[instrument(level = "info", skip(self))]
    pub async fn unyank_crate(&self, krate: &str, version: &str) -> anyhow::Result<()> {
        tracing::info!("unyanking crate `{krate}@{version}`");
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2497/32ce419db5a158a00b4f9881723c03c4c2fdaef0..083197c4a9291a1fbb592894248c5e8106f63cf8)*

src/zulip.rs
```diff
@@ -1518,10 +1519,155 @@ async fn lookup_zulip_username(ctx: &Context, gh_username: &str) -> anyhow::Resu
...
+    if version.contains("..") || version.contains("/") {
+        return Err("Crate version contains invalid characters".to_string());
+    }
+
+    Ok(())
+}
+
+/// Stream that has to be used for yanking/unyanking of crates.
+/// #t-libs/crates
+const YANK_STREAM_ID: u64 = 351149;
```
Urgau: What about compiler or rustdoc owned crates? It seems a bit weird to do them in the t-libs stream.

I don't think having a dedicated topic/stream per team is too much, but having a single dedicated stream seems fine. Wdyt?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2497/32ce419db5a158a00b4f9881723c03c4c2fdaef0..083197c4a9291a1fbb592894248c5e8106f63cf8)*

src/zulip/commands.rs
```diff
@@ -185,6 +185,24 @@ pub enum StreamCommand {
...
         id: PullRequestNumber,
     },
+    /// Yank a specific crate version from crates.io.
+    /// Can only be performed by members of teams that own the crate.
+    Yank {
+        /// Crate to yank.
+        #[arg(name = "crate")]
+        krate: String,
+        /// Version of the crate to yank
+        version: String,
```
Urgau: Any reason not to directly use `semver::Version` it implements [`FromStr`](https://docs.rs/semver/latest/semver/struct.Version.html#impl-FromStr-for-Version) which I think clap uses by default.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2497/32ce419db5a158a00b4f9881723c03c4c2fdaef0..083197c4a9291a1fbb592894248c5e8106f63cf8)*

src/zulip.rs
```diff
@@ -1518,10 +1519,155 @@ async fn lookup_zulip_username(ctx: &Context, gh_username: &str) -> anyhow::Resu
...
+/// in the team database (and thus on crates.io).
+async fn owns_crate(ctx: &Context, gh_id: u64, krate: &str) -> anyhow::Result<bool> {
+    let crate_map = ctx
+        .team
+        .crate_map()
+        .await
+        .context("Cannot load crate map")?;
+    let Some(krate) = crate_map.crates.get(krate) else {
+        return Err(anyhow::anyhow!(
+            "Crate {krate} is not managed in the team database"
```
Urgau: Someone could have typo-ed it, and I think the current message is too strong as it implies that we found it, just not in the team database, which isn't true.

```suggestion
            "Couldn't find {krate} in the team database"
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2497/32ce419db5a158a00b4f9881723c03c4c2fdaef0..083197c4a9291a1fbb592894248c5e8106f63cf8)*

src/zulip.rs
```diff
@@ -1518,10 +1519,155 @@ async fn lookup_zulip_username(ctx: &Context, gh_username: &str) -> anyhow::Resu
...
+        ));
+    }
+
+    ctx.crates_io
+        .unyank_crate(krate, version)
+        .await
+        .context("Cannot unyank crate")?;
+
+    Ok(Some(format!(
+        "Crate {krate}@{version} was successfully unyanked by {}",
```
Urgau:
```suggestion
        "Crate `{krate}@{version}` was successfully unyanked by {}",
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2497/32ce419db5a158a00b4f9881723c03c4c2fdaef0..083197c4a9291a1fbb592894248c5e8106f63cf8)*

src/crates_io/mod.rs
Kobzol: Good point! Added.

src/zulip.rs
Kobzol: The `crate-maintainers` team is inside libs, and libs kind of implies taking care of crates in general. I'm fine with a dedicated channel, but it might get less visibility. Maybe a topic inside #t-infra? There the yanks will be seen by many people.

src/zulip/commands.rs
Kobzol: Good point, I added semver checking later in the process.

src/zulip.rs
Urgau: Sure, that works too.

Kobzol: Changed it to https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Crate.20yanking.2Funyanking/with/621681127.

Urgau (approved):
Looks good to me.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2497/0624e754f6f8bb7c6be8334d541621844a2f084f..a217355fa14f252da64ad58e9a1c7ee9ff5c678f)*

PR #2499, opened 2026-09-10
rustfmt feature: Sync r-l/r milestones to r-l/rustfmt when merging subtree sync PRs

ytmimi:
This adds a new step to the `milestone_prs::handle` that tries to sync r-l/r milestones to r-l/rustfmt after merging r-l/r subtree sync PRs.

For this to work correctly the r-l/r PR needs to have both the `T-rustfmt` and the `subtree-sync` label. Then on the r-l/rustfmt side rustbot will only set the milestone on closed r-l/rustfmt PRs that were previously labeled with `release-notes`.

It's possible that some r-l/rustfmt PRs get merged into r-l/rustfmt's `main` branch between the time it takes to start the sync process and open the subtree sync PR in r-l/r. To mitigate issues around incorrectly setting a milestone on r-l/rustfmt PRs that didn't make it into the r-l/r sync PR, there's an additional filter that only sets milestones on r-l/rustfmt PRs that were merged at least one hour before the r-l/r sync PR was opened.

cc: @rust-lang/rustfmt @rust-lang/rustfmt-contributors 

----
> [!NOTE]
> I didn't know if there was a good way to test this change. `GithubClient` and `Repository` are both concrete types so I don't think there's a good way to mock out their behavior to directly test different scenarios.


Urgau:
Instead of looking at the closed/merged PRs, I think we should be able to use the merge commits created in r-l/rustfmt to our advantage to have an accurate list of the merged PRs.

We [know](https://github.com/rust-lang/rust/commit/a753af41d503edf1d3d0a568ac8d491f3b36724d) they have a title like this: `Merge pull request #<pr_num> from <user>/<branch>`.

I think we could use that to our advantage, by:
 1. finding all the commits that match the merge title (inside `event.issue.compare(gh)?.commits`)
 2. figure out if the commits that matched are in rust-lang/rustfmt
    - can be done with GitHub API: https://api.github.com/repos/rust-lang/rustfmt/git/commits/a753af41d503edf1d3d0a568ac8d491f3b36724d (we probably have a method for that)
 3. set the milestone for all the PRs we correlated back 

What do you think?

*An optimization to be done is to only to this when files under `src/tools/rustfmt` are being changed. `milestone_submodules` has such detection*

ytmimi:
@Urgau thank you for the feedback! Seems like a much more robust approach than what I was originally thinking. Taking the most recent subtree-sync as an example (https://github.com/rust-lang/rust/pull/161964) this is what I think we could do.

1. Paginate through GET request to https://api.github.com/repos/rust-lang/rust/pulls/161964/commits
2. Pull out all merge commits that have `Merge pull request #<pr_num>`. rustfmt recently started using a merge queue so all new PRs should get merged into the `main` branch this way.
3. Apply the milestone to all PRs that we just found.

Urgau:
> this is what I think we could do

That seems broadly fine with the caveat that we need to check that the merge commits are in r-l/rustfmt, as the subtree sync could have been in a rollup and we would start milestoning the wrong PRs.

Btw, we [already have](https://github.com/rust-lang/triagebot/blob/8e0b53bb30a835a0eb72a4097d16db99af5c4240/src/github/issue.rs#L551) a method to get all the commits of a PR.

ytmimi:
> Btw, we [already have](https://github.com/rust-lang/triagebot/blob/8e0b53bb30a835a0eb72a4097d16db99af5c4240/src/github/issue.rs#L551) a method to get all the commits of a PR.

That's amazing!

> That seems broadly fine with the caveat that we need to check that the merge commits are in r-l/rustfmt, as the subtree sync could have been in a rollup and we would start milestoning the wrong PRs.

I was thinking that we'd just look at the commit messages and use a regex to extract the PR number. In that case would we still need to check that the commits are in rustfmt? Maybe I'm misunderstanding something?

We typically don't include rustfmt subtree syncs in rollups, but I agree that we shouldn't rely on that here. Is the concern that if it's included in a rollup that we might accidentally assign milestones to PR numbers from other repos that just happen to have the same number as rustfmt? Maybe the `T-rustfmt` and `subtree-sync` label checks would help prevent that? Presumably the rollup wouldn't get tagged with `subtree-sync`?

My original thought was that this would only get triggered when the subtree-sync PR was closed and that we'd be able to reference that PR and grab all of its commit messages. At that point we'd know that all the PRs referenced in the `Merge pull request #<pr_number>` were rustfmt PRs. I didn't consider rollups when I first started thinking about this, so maybe there are some edge cases around those that I need to think through. 

I'm also wondering if your original suggestion to use `event.issue.compare(gh)?.commits` would be easier for this? I'm just not familiar with the triagebot repo and what's the easiest way to go about this.

Urgau:
> Is the concern that if it's included in a rollup that we might accidentally assign milestones to PR numbers from other repos that just happen to have the same number as rustfmt?

Yes, that's my concern.

Not checking that the commits are indeed in r-l/rustfmt would effectively prevent doing rollups, which is not ideal, in particular since we don't have a way to prevent someone from accidentally rollup-ing future syncs.

> Maybe the `T-rustfmt` and `subtree-sync` label checks would help prevent that? Presumably the rollup wouldn't get tagged with `subtree-sync`?

I rather avoid gating on labels when we can, they can be forgotten or missed by triagebot (we may miss a webhook); and as you mention it as well the rollup probably won't get tagged with `subtree-sync`.

> I'm also wondering if your original suggestion to use `event.issue.compare(gh)?.commits` would be easier for this?

I think our implementation does not properly gather all the pages (it only does one call), so we would have to fix it first before using it. I will try fixing it this week-end.

But the reason I suggested `compare` is that it's already called in the handler (via `diff`) and that it's output is cached for all the handlers, so one less call.

ytmimi:
I just asked a question in [#t-rustfmt > Merge queue](https://rust-lang.zulipchat.com/#narrow/channel/357797-t-rustfmt/topic/Merge.20queue/with/553552430). I'm wondering if it would simplify things if we could update the merge commit messages from `Merge pull request #{pr_number}` -> `Merge pull request rust-lang/rustfmt#{pr_number}` so that we'd have more confidence that the commit came from rustfmt, and then we'd have an easier time pulling out the PR numbers that we need. Would probably also help us avoid issues with rollups.

-----
Edit: Not sure how feasible this is, but if we could set something like that up then I think we might be able to take a similar approach for other subtrees.

Urgau:
> I'm wondering if it would simplify things if we could update the merge commit messages from `Merge pull request #{pr_number}` -> `Merge pull request rust-lang/rustfmt#{pr_number}`

That would certainly help, unfortunately it's not possible. GitHub doesn't allow setting a custom merge queue commit title (see https://github.com/orgs/community/discussions/15925).

But it's not necessary, we just need to make sure the SHA of the merge commit (the ones with `Merge pull request #{pr_number}`), is in rust-lang/rustfmt. Which can be done pretty easly with GitHub API by trying to fetch the details of the commit in rust-lang/rustfmt https://api.github.com/repos/rust-lang/rustfmt/git/commits/a753af41d503edf1d3d0a568ac8d491f3b36724d.

ytmimi:
@Urgau Are the SHAs from the merge commits guaranteed to be the same? I thought they might be different, but I picked out one of the commits from https://github.com/rust-lang/rust/pull/161964 to double check and it looks like the SHAs are consistent. From this message:

```
Merge pull request #7066 from ytmimi/issue_7062
fix: correct the span used when rewriting `ast::TyKind::FnPtr`
``` 

Which corresponds to this rustfmt PR https://github.com/rust-lang/rustfmt/pull/7066

* rustfmt commit: https://github.com/rust-lang/rustfmt/commit/dbc566c93bd3e6f456a3a65556077897ecd559a6
* rust-lang/rust commit: https://github.com/rust-lang/rust/commit/dbc566c93bd3e6f456a3a65556077897ecd559a6

----

Another example from https://github.com/rust-lang/rust/pull/161964 that's not a merge commit because it was merged before we started using the merge queue, but it looks like the SHAs are consistent there too:

rustfmt PR: https://github.com/rust-lang/rustfmt/pull/6010

* rustfmt commit: https://github.com/rust-lang/rustfmt/commit/0675df612e6dd311a52b88e679dee9bd1e3b108e
* rust-lang/rust commit https://github.com/rust-lang/rust/pull/161964/changes/0675df612e6dd311a52b88e679dee9bd1e3b108e

Urgau:
> Are the SHAs from the merge commits guaranteed to be the same?

Good question. Seems like they are the same with `git subtree`, at least while looking at your subtree sync PRs.

However, looking at a Josh sync, I'm not seeing the same SHA, in rust-lang/rust https://github.com/rust-lang/rust/commit/8173a82b828ff9db17ce013899a753644c62c663, in rust-lang/miri https://github.com/rust-lang/miri/commit/8de2ef99551d48b9d87ca5201c2f7d7933a2946c.

Any hindsight to share @Kobzol?

Kobzol:
With Josh, the commit SHAs in r-l/r and the subtree will be different.

Urgau:
> With Josh, the commit SHAs in r-l/r and the subtree will be different.

That's a good news, for rustfmt at least, less for Josh subtrees.

I guess there isn't an easy way to correlate them without running Josh?

> However, looking at a Josh sync, I'm not seeing the same SHA, in rust-lang/rust [rust-lang/rust@8173a82](https://github.com/rust-lang/rust/commit/8173a82b828ff9db17ce013899a753644c62c663), in rust-lang/miri [rust-lang/miri@8de2ef9](https://github.com/rust-lang/miri/commit/8de2ef99551d48b9d87ca5201c2f7d7933a2946c).

Looking more that those two commits, the one is rust-lang/rust has it's path corrected to `src/tools/miri`, but that's not the same for the rustfmt ones https://github.com/rust-lang/rust/commit/dbc566c93bd3e6f456a3a65556077897ecd559a6, they are not corrected in rust-lang/rust with `git subtree`.

So what I think we can do for Josh subtrees, if we ever get one that wants milestones, is to look at which files the commit modifies.

If the commit only modifies files inside the subtree and is named `Merge pull request #{pr_number}`, then we milestone the corresponding PR.

This is of course hypothetical since rustfmt is not a Josh subtree, and we can just check the SHA with `git subtree`.

jieyouxu:
> This is of course hypothetical since rustfmt is not a Josh subtree, and we can just check the SHA with git subtree.

Right, but I think it's more "not yet", I think we'd like to move to become a Josh subtree to make the syncs easier, once the WIP Josh things are settled down.

src/handlers/milestone_prs.rs
```diff
@@ -190,3 +192,100 @@ async fn milestone_submodule(
...
+            log::error!("failed to fetch repository details for rust-lang/rustfmt");
+            return;
+        };
+
+        let Ok(milestone) = subtree.get_or_create_milestone(&gh, &mv, "open").await else {
+            log::error!("failed to get or create milestone {mv} for rust-lang/rustfmt");
+            return;
+        };
+
+        let merge_re = Regex::new(r"Merge pull request #([0-9]+)").unwrap();
```
Urgau: Let's put the regex in a static `LazyLock` at the top. No reason to recreate it every-time.

Btw, after making it `static` we can have a simple `#[test]` to make sure the `unwrap()` won't crash triagebot. 

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2499/4f2b6bda2ff024925131cbbc3eb56168de75ff00..849f8145c61cae1af72f86223518381e20815716)*

src/handlers/milestone_prs.rs
```diff
@@ -190,3 +192,100 @@ async fn milestone_submodule(
...
+/// milestones to.
+async fn milestone_rustfmt(
+    gh: &GithubClient,
+    event: &IssuesEvent,
+    milestone_version: &str,
+) -> anyhow::Result<()> {
+    if !event.issue.contains_label(&"T-rustfmt".into()) {
+        // This PR doesn't touch rustfmt so we can skip it.
+        return Ok(());
+    }
```
Urgau: I would prefer to not check the presence of the label since it may be missing. Triagebot can fail to add it.

I would suggest instead to look at the files (`event_diff.files`) and see if any of them has `src/tools/rustfmt` as a prefix.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2499/4f2b6bda2ff024925131cbbc3eb56168de75ff00..849f8145c61cae1af72f86223518381e20815716)*

src/handlers/milestone_prs.rs
```diff
@@ -190,3 +192,100 @@ async fn milestone_submodule(
...
+                    && subtree_commit.parents[1].sha == merge_commit.parents[1].sha
+                {
+                    Some(pr_number)
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>()
+            .await;
+
```
Urgau:
```suggestion

        log::info!("milestoning rustfmt PRs: {:?}", &pr_numbers);
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2499/4f2b6bda2ff024925131cbbc3eb56168de75ff00..849f8145c61cae1af72f86223518381e20815716)*

src/handlers/milestone_prs.rs
```diff
@@ -190,3 +192,100 @@ async fn milestone_submodule(
...
+                    return None;
+                };
+
+                if subtree_commit.parents.len() != 2 {
+                    return None;
+                }
+
+                if subtree_commit.sha == merge_commit.sha
+                    && subtree_commit.parents[0].sha == merge_commit.parents[0].sha
+                    && subtree_commit.parents[1].sha == merge_commit.parents[1].sha
```
Urgau: If the SHAs are equal do we even need to check the parents? aren't they always equal since they take part in the SHA calculation.

Feels like to me this whole closure could just be: `subtree.github_commit(&gh, &merge_commit.sha).await.is_ok()`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2499/4f2b6bda2ff024925131cbbc3eb56168de75ff00..849f8145c61cae1af72f86223518381e20815716)*

ytmimi: Was just being overly cautious with the checks. It's probably fine to just go with `subtree.github_commit(&gh, &merge_commit.sha).await.is_ok()`.

src/handlers/milestone_prs.rs
ytmimi: Good call!

src/handlers/milestone_prs.rs
ytmimi: In https://github.com/rust-lang/triagebot/pull/2499#issuecomment-5641230651 you mentioned:
> Looking more that those two commits, the one is rust-lang/rust has it's path corrected to src/tools/miri, but that's not the same for the rustfmt ones https://github.com/rust-lang/rust/commit/dbc566c93bd3e6f456a3a65556077897ecd559a6, they are not corrected in rust-lang/rust with git subtree.

I interpreted that to mean that we wouldn't be able to check for `src/tools/rustfmt` in the file names, but if that's not the case then I'll definitely make this update.

ytmimi:
Pushed a lot of changes. I applied all the feedback and then tried to follow the same structure used by `milestone_submodules` and `milestone_submodule` to generalize things for any other git subtree to use. As mentioned earlier, we'll need to take a different approach for josh subtrees since the SHAs won't match.

src/handlers/milestone_prs.rs
Urgau: Only the individual commits are not prefixed, if [you look](https://api.github.com/repos/rust-lang/rust/compare/17fd5b8a37b6667b6cc137f3cc35f09759768a3b...e3e8a34286c2c9d8766518bee5e7040eabe1c6d1) at the PR compare, you can see in the `files` that all the filenames are correctly prefixed by `src/tools/rustfmt/`. *(Otherwise they wouldn't be merged at the correct place)*.

Sorry for the confusion.

Urgau:
Looking at [#t-rustfmt > Migrating to Josh](https://rust-lang.zulipchat.com/#narrow/channel/357797-t-rustfmt/topic/Migrating.20to.20Josh/with/624242636), seems like the migration to Josh is pretty imminent.

I'm fine merging the PR the meantime, but since it will only be used once (hopefully), it may be better to prepare instead for Josh?

-------

Regarding what to do with Josh subtrees, we talked about it at [#t-infra > saving the original commit SHAs with Josh subtree](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/saving.20the.20original.20commit.20SHAs.20with.20Josh.20subtree/with/624435817) and the consensus is to filter all the "Merge pull request" commits and milestone those who only contain `src/tools/rustfmt` changes.

ytmimi:
> Looking at [#t-rustfmt > Migrating to Josh](https://rust-lang.zulipchat.com/#narrow/channel/357797-t-rustfmt/topic/Migrating.20to.20Josh/with/624242636), seems like the migration to Josh is pretty imminent.
> 
> I'm fine merging the PR the meantime, but since it will only be used once (hopefully), it may be better to prepare instead for Josh?
> 
> Regarding what to do with Josh subtrees, we talked about it at [#t-infra > saving the original commit SHAs with Josh subtree](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/saving.20the.20original.20commit.20SHAs.20with.20Josh.20subtree/with/624435817) and the consensus is to filter all the "Merge pull request" commits and milestone those who only contain `src/tools/rustfmt` changes.

The transition to a Josh Subtree is likely around the corner, but I'd like to test out this current version for rustfmt it you're okay with it. My hope with the current `milestone_git_subtrees` implementation is that it could be easily extended to support other git subtrees. Clippy comes to mind.

That said, once we transition rustfmt to a Josh subtree we can also remove the git subtree implementation if you don't want dead code kicking around in triagebot.

I'm happy to make those changes once the time comes!

Urgau:
> My hope with the current `milestone_git_subtrees` implementation is that it could be easily extended to support other git subtrees. Clippy comes to mind

Clippy is probably also around the corner with regards to using Josh: https://github.com/rust-lang/rust-clippy/pull/17739; but point taken.

ytmimi:
😅 Oh I didn't realize that clippy was planning to transition too. [Well then maybe one of these other subtrees](https://rustc-dev-guide.rust-lang.org/external-repos.html?#external-dependencies-subtrees) :)

Kobzol:
Hopefully all of them will migrate soon, now that we have the new version of Josh working :)

src/handlers/milestone_prs.rs
```diff
@@ -3,10 +3,15 @@ use crate::{
...
 };
 use anyhow::Context as _;
+use futures::stream::{self, StreamExt};
 use regex::Regex;
 use reqwest::StatusCode;
+use std::sync::LazyLock;
 use tracing as log;
 
+static MERGE_QUEUE_COMMIT_TITLE_RE: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"Merge pull request #([0-9]+)").unwrap());
```
Urgau: small nit, let's make sure we match at the start of the commit message and not in the middle of it 
```suggestion
    LazyLock::new(|| Regex::new(r"^Merge pull request #([0-9]+)").unwrap());
```

Urgau (commented):
The logic looks good to me. Only a small nit and we should be ready to go.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2499/12a941ca225969483e9e38a5f7f6d13157925cb1..b1f3ce98223d577039567dd52c858721fe4b0423)*

src/handlers/milestone_prs.rs
ytmimi: Just updated the regular expression

Urgau (approved):
Let's try it.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2499/2ac18aa4503cabbd004cc736710fd5be3990efaa..b6455bb8f0e27c8b1857f288d6ffd6fe6ba9bca8)*

PR #2501, opened 2026-09-13
Fix our LRU cache implementation and add unit-tests

Urgau:
This PR fixes two issues with our LRU cache implementation:
 1. A duplicate entry with a size bigger than the capacity wouldn't be removed from the cache keeping a stale entry. No a case we should ever have.
 2. We assumed that removing only one entry was enough to have sufficient space to not go over the defined capacity, which is not always the case, but was probably fine for us given the uniformity of our inputs.


src/cache.rs
```diff
@@ -83,4 +83,98 @@ impl<K: PartialEq<K>, V: EstimatedSize> LeastRecentlyUsedCache<K, V> {
             false
         }
     }
+
+    /// Current size
+    #[cfg(test)]
+    pub(crate) fn size(&self) -> usize {
```
Kobzol:
```suggestion
    fn size(&self) -> usize {
```
Should be enough for the test, I guess?

Kobzol (approved):
Thanks, left one nit, but you can merge as-is.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2501/02a9e45463e42a7d9f3f17bff0dd26bfaedab051..7568d68027caa9c984e9b194cedacfb1638758cc)*

Kobzol:
Maybe we can bump the capacity, now that we have more RAM.

Urgau:
> Maybe we can bump the capacity, now that we have more RAM.

Will do in any another PR, with the rest of the cache at once.

PR #2502, opened 2026-09-13
Misc validation and escaping

Urgau:


Kobzol (approved):
In the second commit, why would the code crash if the SHAs were invalid?

I'm fine with the PR as-is, but I think that we should really consider using Askama or Tera to get auto-escaping. This kind of string interpolation is not great.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2502/02a9e45463e42a7d9f3f17bff0dd26bfaedab051..0d6b43ae40d095e7c92060cb7e33f0cef2af3033)*

Urgau:
> In the second commit, why would the code crash if the SHAs were invalid?

Indexing would fail if it's not alphanumeric.

https://github.com/rust-lang/triagebot/blob/02a9e45463e42a7d9f3f17bff0dd26bfaedab051/src/gh_range_diff.rs#L952-L953

Urgau:
> I'm fine with the PR as-is, but I think that we should really consider using Askama or Tera to get auto-escaping. This kind of string interpolation is not great.

Agree with you on that. It was fine when things were simple but it's kind of getting out of hand.

Would maybe be a good GSoC project for next year.

PR #2503, opened 2026-09-13
Add warning against skipped workflow runs in PRs

Urgau:
This PR adds warning against skipped workflow runs in PRs.

Per [GitHub docs](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs) there are multiple instructions possible, all of them are checked.

```toml
# Enable skipped workflow runs warning
[skipped-workflow-runs]
```

The handler goes straight to being enabled org-wide (discussed on Zulip).

Requested in [#triagebot > Warning message for skip CI](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Warning.20message.20for.20skip.20CI/with/623886543)

cc @jieyouxu


Urgau:
The warning looks like this:

 - The last commit of this PR contains an instruction to [skip workflow runs](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs) for this PR (i.e., skip running CI).
   This is usually not advisable, as it bypasses automated testing.

src/handlers/check_commits/skipped_workflow_runs.rs
```diff
@@ -0,0 +1,154 @@
...
+    "[skip ci]",
+    "[ci skip]",
+    "[no ci]",
+    "[skip actions]",
+    "[actions skip]",
+];
+
+const INSTRUCTIONS_TRAILER: &[&str] = &["skip-checks:true", "skip-checks: true"];
+
+const WARNING_MESSAGE: &str = "The last commit of this PR contains an instruction to [skip workflow runs](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs) for this PR (i.e., skip running CI).\nThis is usually not advisable, as it bypasses automated testing.";
```
Kobzol:
```suggestion
const WARNING_MESSAGE: &str = "The last commit's message of this PR contains an instruction to [skip workflow runs](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/skip-workflow-runs) for this PR (i.e., skip running CI).\nThis is usually not advisable, as it bypasses automated testing.";
```
Just to make it clear where to search for the instruction, i.e. that it's not in the modified files of the commit.

src/handlers/check_commits/skipped_workflow_runs.rs
```diff
@@ -0,0 +1,154 @@
...
+    let msg = commits.last()?.commit.message.as_str();
+
+    // First check instructions that can happen anywhere
+    if INSTRUCTIONS_ANYWHERE.iter().any(|ins| msg.contains(ins)) {
+        return Some(WARNING_MESSAGE.to_string());
+    }
+
+    // Second let's check the trailer section (instructions at the very end of the commit message)
+
+    let mut lines = msg.lines().rev();
```
Kobzol, line 30: You can have multiple trailers in a single commit, any any of them could match the instruction, right?

Maybe it would be easier to do something like:
```rust
// Collect all trailers
let trailers = input.lines().rev().take_while(|l| !l.is_empty()).collect::<Vec<_>>().into_iter().rev();
if trailers.any(|trailer| INSTRUCTIONS_TRAILER.contains(trailer)) { <error> }
```

Kobzol (commented):
Thanks! Left some comments.

I think it would be maybe a bit easier to understand if it was called "skipped-ci", rather than copying the GitHub name. But you already have this name, and I don't really expect anyone messing with the config, and it is enabled org-wide, so ultimately the name doesn't really matter, I think.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2503/4f2b6bda2ff024925131cbbc3eb56168de75ff00..8c36875cbedccaddb9dba78c0444f6d89b6e01b8)*

src/handlers/check_commits/skipped_workflow_runs.rs
Urgau: > You can have multiple trailers in a single commit, any any of them could match the instruction, right?

Yes there can be multiple trailers, but GitHub docs explicitly states: "If you already have other trailers in your commit message, skip-checks should be last".

Urgau: Modified the comment to make it more explicit.

Urgau:
> I think it would be maybe a bit easier to understand if it was called "skipped-ci"

Yeah, I bounced back and forth multiple times on the naming, ended-up choosing with the GitHub name for consistency but even them are not consistent (they use: ci, actions, checks).

src/handlers/check_commits/skipped_workflow_runs.rs
Kobzol: Oh, I didn't notice that. Then I don't understand the `>= 2` part, shouldn't that be `>= 1`? In any case, it would be nice to have a test for this case.

Urgau: This is another quirk of GitHub: "The trailers section should be included at the end of your commit message *and be preceded by two empty lines*"

Kobzol (approved):

PR #2504, opened 2026-09-13
feat: add a note on llm trailer in commit message

CommanderStorm:
[#triagebot > Scanning for LLM trailers](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Scanning.20for.20LLM.20trailers/with/623906722)

Should be pretty much what we want, or? 🤔
Not sure of the exact wording, that stuff is hard project wide..

## LLM usage

I asked claude for the list of semi-popular `LLM_CO_AUTHORS`, since I only know noreply@antropic.com...
I have no data to back this up, but LGTM 🤷🏻

I also asked for these session things that linux and claude use, since the first is something that people MIGHT use and the second something that they might forget to delete :)
Apparently, that is not really a thing for others, also no data to back this up


src/handlers/check_commits/llm_assisted.rs
```diff
@@ -0,0 +1,158 @@
...
+/// `Co-authored-by` values identifying an LLM agent.
+/// Matched on the email, as "Claude" could possibly be a human name.
+const LLM_CO_AUTHORS: &[&str] = &[
+    "noreply@anthropic.com",
+    "copilot@users.noreply.github.com",
+    "cursoragent@cursor.com",
+    "@aider.chat",
+    "devin-ai-integration[bot]",
+    "google-labs-jules[bot]",
+];
```
Urgau, line 24: This is exactly what I afraid [in Zulip](https://rust-lang.zulipchat.com/#narrow/channel/224082-triagebot/topic/Scanning.20for.20LLM.20trailers/near/623915639). I don't want us to maintain such an ever growing list.

*I would fine if someone else, outside this repository, even outside this org, maintained this list, but not triagebot.*

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2504/4f2b6bda2ff024925131cbbc3eb56168de75ff00..fee091f287aff53f2da3b6f9671bf76c999097da)*

Urgau (changes requested):

src/handlers/check_commits/llm_assisted.rs
CommanderStorm: To clarify what the prefered solution is:
- make it configurable, maybe with this default
- move it to two git gists which I maintain

PR #2505, opened 2026-09-13
ci: Enable clippy and fix lints

CommanderStorm:
trying to run my previous PR through clippy I noticed a few clippy nits.

Thought you might want this (or not, that is fine too)


Kobzol:
I think that last time we talked about this, we decided not to use Clippy. CC @Urgau @apiraino

Urgau:
I'm personally fine having from time to time a PR that fixes some Clippy warnings, but I wouldn't want it enabled by default in CI.

apiraino:
this was discussed multiple times, see #2201 for example

We prefer keeping clippy out of the CI, but thanks for your PR @CommanderStorm

PR #2506, opened 2026-09-15
Handle file renames in our range-diff viewer

Urgau:
This PR fixes our range-diff viewer to handle filename renames, in particular for diff that only consist of a rename.

The implementation is a bit crude, but it's simple works 100% of the time (unlike the 3 other attempts I made).

Discovered at https://github.com/rust-lang/rust/pull/162372#discussion_r4006835985 where we currently so no difference at all!

<details>

image

</details>


Kobzol (approved):
Looks simple enough. Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2506/12a941ca225969483e9e38a5f7f6d13157925cb1..7a1d2c45b2daf02f481fa374b6bc3901a0998d65)*

PR #2509, opened 2026-09-19
Allow negative globs for required_labels in label/unlabel events

clarfonthey:
This is a probably easy solution to the issue encountered here: [#t-libs/backports > #161430: beta-nominated @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/542373-t-libs.2Fbackports/topic/.23161430.3A.20beta-nominated/near/625335373)

Essentially, we notify that backports are added or removed as completely independent messages, but the `beta-nominated` tag is actually removed when merging an accepted backport, causing a bit of confusion.

This would allow us to update the `required_labels` for such messages to `required_labels = ["T-libs", "!beta-accepted"]` to suppress these messages in the case where an accepted backport is "un-nominated" to indicate it's being actually merged into the beta branch.

The suggestion to also send a message in this case would similarly work because we could simply add an extra case where the `beta-accepted` tag *is* required, which mentions that it's being merged.

There don't seem to be any tests for this, and I dunno if you want them, but, I think this should work.


src/handlers/notify_zulip.rs
```diff
@@ -164,8 +164,14 @@ fn parse_open_close_reopen_input(
...
 
-fn has_all_required_labels(issue: &Issue, config: &NotifyZulipLabelConfig) -> bool {
+fn has_valid_label_config(issue: &Issue, config: &NotifyZulipLabelConfig) -> bool {
     for req_label in &config.required_labels {
+        let mut req_label = &**req_label;
+        let mut expected = true;
+        if req_label.starts_with("!") {
+            expected = false;
+            req_label = &req_label[1..];
+        }
```
fmease, line 174:
```suggestion
        let stripped = req_label.strip_prefix("!");
        let expected = stripped.is_none();
        let req_label = stripped.unwrap_or(&req_label);
```

```suggestion
        let mut expected = true;
        let req_label = req_label
            .strip_prefix("!")
            .inspect(|_| expected = false)
            .unwrap_or(&req_label);
```


```suggestion
        let mut req_label = &**req_label;
        let mut expected = true;
        if let Some(label) = req_label.strip_prefix("!") {
            expected = false;
            req_label = label;
        }
```

or something similar


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2509/63312d7d0e87ebb2a3498b2acfc9e7f2979f8a26..14e02f20d92690453c440156f2725e0f8f923cef)*

apiraino:
Trying to summarize the goal of this patch[^1] (would probably good to also save the reasoning in the commit for documentation purposes): this will prevent sending the Zulip message "*beta*-nomination has been removed" when T-release removes the `*-nominated` label, correct?

A sample configuration then will be like (from https://github.com/rust-lang/rust/pull/162994/files):
```toml
[notify-zulip."beta-nominated".libs]
required_labels = ["T-libs", "!beta-accepted"]
zulip_stream = 542373 # #t-libs/backports
topic = "#{number}: beta-nominated"
message_on_add = ["..."], ["..."]
message_on_remove = "PR #{number}'s beta-nomination has been removed."
```

In this example this will prevent (IIUC) any backport nomination label changes to a PR having `T-libs` **and** `beta-accepted` labels, correct? What happens in this scenario?
1. user adds `beta-nominated`
2. team adds `beta-accepted`
3. user wants to add `stable-nominated` :point_left: will it work?

thanks

---
[^1]: I attempted something with negating labels a while ago and didn't bring it to end, forgot exactly why

PR #2510, opened 2026-09-19
Stop trusting Zulip custom profile fields in the `lookup` command

Kobzol:
And also enable using it in streams.

Discussed in https://rust-lang.zulipchat.com/#narrow/channel/546943-all.2Fprivate/topic/zulip.2Fgithub.20username.20mapping/with/625416490.


src/zulip.rs
```diff
@@ -1443,61 +1444,60 @@ async fn lookup_github_username(ctx: &Context, zulip_username: &str) -> anyhow::
...
-            .map(|u| u.user_id))
-    }
-
-    async fn lookup_zulip_id_from_team(
-        ctx: &Context,
-        gh_username: &str,
-    ) -> anyhow::Result<Option<u64>> {
+    // We do not trust GitHub profile fields entered on Zulip, they can be modified arbitrarily
+    // We only trust the team database
+    let zulip_id = async move {
```
Urgau, line 1500: No need for this to be a closure, right?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2510/63312d7d0e87ebb2a3498b2acfc9e7f2979f8a26..8930e5fc76f49ff9d677b9924dd28749cd53d11a)*

Kobzol: It essentially acts like a `try` block.

Urgau (approved):

PR #2513, opened 2026-09-21
Warn about large pull requests

blyxyas:
Pull requests that change over 200 lines (some sources say 250, others 150) are known to be very difficult to review, and extremely easy to contain bugs that the reviewer does not notice.
I personally find it difficult and kind of boring to review an e.g. 700 lines change.

So this PR adds a new `[large_pull_requests]` config, that posts a comment if the pull request is larger than a `threshold`. We can also ignore files, mainly to avoid penalizing adding test cases.
This is my first feature PR to triagebot, so any tips are very welcome!


src/handlers/large_pull_requests.rs
Urgau, line 1: Could you move the module inside the `check_commits/` directory like the other check commits handlers.

src/handlers/large_pull_requests.rs
```diff
@@ -0,0 +1,37 @@
+use anyhow::Context as _;
+
+use crate::config::LargePullRequests;
+use crate::github::{GithubCommit, GithubCompare, IssuesAction};
+use crate::utils::ModifiedPathMatcher;
+use crate::{github::Event, handlers::Context};
+
+pub fn large_pull_request_message(threshold: u64) -> String {
```
Urgau:
```suggestion
fn large_pull_request_message(threshold: u64) -> String {
```

src/handlers/large_pull_requests.rs
```diff
@@ -0,0 +1,37 @@
...
+    format!(
+        "Seems that this commit is larger than expected (threshold: {threshold} lines of code changed) \
+Big pull requests are reviewed much slower than small ones, consider splitting it."
+    )
+}
+
+pub(crate) fn large_pull_requests(
+    compare: &GithubCompare,
```
Urgau:
```suggestion
    files: &[FileDiff],
```

src/handlers/large_pull_requests.rs
```diff
@@ -0,0 +1,37 @@
...
+                None
+            }
+        }).sum();
+
+    if (number_of_changed_lines as u64) >= config.threshold {
+        return Some(large_pull_request_message(config.threshold));
+    }
+
+    None
+}
```
Urgau: Do you think it's would be possible to add some unit-test for this handler?

Nothing complicated, just testing that it works as expected. I'm thinking of two small unit-tests one with too few changed file and one with enough changed files.

Maybe like this?

```rust
#[test]
fn enough_changed_files() {
    let files: [FileDiff; 200] = std::array::from_fn(|_| FileDiff { name: "foo".to_string(), previous_filename: None, patch: "".to_string() });

    assert_eq!(large_pull_requests(&files, &LargePullRequests { /* ... */ }));
}
```

src/config.rs
```diff
@@ -1115,6 +1125,7 @@ mod tests {
                     trigger_files: vec!["posts/".to_string()],
                     exclude_files: vec![],
                 }),
+                large_pull_requests: None,
```
Urgau: You do add the handler to the config above, to make sure we can parse it correctly.

Btw, I think the handler name is going to be `[large-pull-requests]` (`_` vs `-`).

src/config.rs
```diff
@@ -646,6 +647,15 @@ pub(crate) struct MergeConflictConfig {
 #[serde(deny_unknown_fields)]
 pub(crate) struct BotPullRequests {}
 
+#[derive(PartialEq, Eq, Debug, Clone, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub(crate) struct LargePullRequests {
+    #[serde(default)]
+    pub(crate) exclude_files: Vec<String>,
+    pub(crate) threshold: u64,
```
Urgau, line 663: I don't know if it's worth it, but we could make it non-zero. Wdyt?

src/config.rs
```diff
@@ -646,6 +647,15 @@ pub(crate) struct MergeConflictConfig {
 #[serde(deny_unknown_fields)]
 pub(crate) struct BotPullRequests {}
 
+#[derive(PartialEq, Eq, Debug, Clone, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub(crate) struct LargePullRequests {
+    #[serde(default)]
+    pub(crate) exclude_files: Vec<String>,
```
Urgau, line 657: Probably worth adding a comment that it's glob based.

Urgau (commented):
Good start, left some minor things.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2513/81366ca1d7e3418bf59a86e9d07eb58d95a99574..d687b4737ef73c69ddf652d3ceab4922987bdbd5)*

src/config.rs
blyxyas: Hmmm I've made the change, but I think it severely adds much more complexity than the problem it solves :/

src/config.rs
blyxyas: Done!

src/handlers/large_pull_requests.rs
blyxyas: Thanks for this comment, I noticed a bug while writing the test. We were not accounting for whitespace. I've added a `.trim()` now

src/config.rs
```diff
@@ -646,6 +647,17 @@ pub(crate) struct MergeConflictConfig {
...
 pub(crate) struct BotPullRequests {}
 
+#[derive(PartialEq, Eq, Debug, Clone, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+#[serde(deny_unknown_fields)]
+pub(crate) struct LargePullRequests {
+    /// Path globs to exclude; useful for only focusing on logic code,
+    /// permitting as much testing as possible.
+    #[serde(default)]
+    pub(crate) exclude_files: Vec<String>,
```
Urgau, line 657: One last thing. I forgot we had it, but we actually have some validation logic to help users in case the pattern in invalid.

Could you add a similar code? (no tests required)

https://github.com/rust-lang/triagebot/blob/81366ca1d7e3418bf59a86e9d07eb58d95a99574/src/handlers/check_commits/validate_config.rs#L127-L130

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2513/81366ca1d7e3418bf59a86e9d07eb58d95a99574..aa3791064258f1412ad50c6a10baa5a8f7311d5a)*

Urgau:
Btw, I just thought of something, what about subtree syncs? I don't think you would want them to be warned about, and I don't think you can ignore them with just `exclude_files`. 

Maybe worth adding an `exclude_titles` or something like that. (Can be addressed in a future PR.)

blyxyas:
Hmmm that's a good idea, should this also exclude PRs open by bots? I think there are some repos that use Renovate

Urgau:
I don't use renovate and such bots, but I think they only changes lock files and manifest (`Cargo.toml` and such). I think we can just exclude them with `exclude_files`.

src/config.rs
blyxyas: Done in [`f0327ac` (this PR)](https://github.com/rust-lang/triagebot/pull/2513/commits/f0327ac2f62d936fd3e9d6dd751963c23e140f38)

Urgau (approved):
Thanks.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/triagebot/2513/22116df09569c1846a67dd37c2533923f5796dd2..f0327ac2f62d936fd3e9d6dd751963c23e140f38)*

